PR #28, opened 2023-04-10
Add suggested extensions

Hinton:


MGibson1 (commented):
We should probably change these depending on the repo, right?

I think the only ones valid here are 
```
    "MS-vsliveshare.vsliveshare",
    "rust-lang.rust-analyzer",
    "streetsidesoftware.code-spell-checker",
    "tamasfe.even-better-toml",
    "vadimcn.vscode-lldb"
```

MGibson1 (approved):
Good point, thinking too small. We have the language impls in there too

PR #30, opened 2023-04-13
Secret creation and editing from the CLI

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
Added secret create/edit/delete commands to the secrets CLI. PR migrated from the old repository.

- For creating a secret, I'm not super happy with the optional `--note` and `--project-id` appearing in the options section, I feel like it makes them harder to notice and would look best in a separate section, but not sure if that's something we can configure in clap.
- For editing a secret, only one of `--key`, `--value` or `--note` is required, but all can be used simultaneously.
- Deleting a secret is straightforward, just pass the ID and it prints a confirmation message.
- Small change to response serialization. serde_json doesn't append a trailing newline, while yaml does. At the same time, using the color option would leave the text as-is, while having the color disabled would add another newline because of the use of println!. This resulted in `--color yes --format` json having zero trailing newlines while `--color no --format yaml` would have two.

## Code changes

- **Schemas and secrets_request.rs**: Updated to reflect the new `project_ids` in the create secret request
- **main.rs**: Added the new secret create/edit/delete commands
- **render.rs**: Make sure `serialize_response` prints all the responses with exactly one trailing newline character.

Rebased from https://github.com/bitwarden/sdk-old-2023-03-20/pull/116

## Screenshots
Moved from the old PR:

image

image

## Notes
At the moment the server won't accept write operations to secrets from service accounts, I had to modify this file to get it working: [CreateSecretCommand.cs#L37](https://github.com/bitwarden/server/blob/85099307227d0703cef34d70d8567973546cf300/bitwarden_license/src/Commercial.Core/SecretsManager/Commands/Secrets/CreateSecretCommand.cs#L37)
```diff
        var hasAccess = accessClient switch
        {
            AccessClientType.NoAccessCheck => true,
            AccessClientType.User => await _projectRepository.UserHasWriteAccessToProject(project.Id, userId),
+            AccessClientType.ServiceAccount => await _projectRepository.ServiceAccountHasWriteAccessToProject(project.Id, userId),
            _ => false,
        };
```

I also had to set to 1 the Write column of the service account in the AccessPolicy table as the UI doesn't expose that option just yet.


crates/bws/src/main.rs
```diff
@@ -67,6 +68,21 @@ enum Commands {
...
+    #[command(long_about = "Edit a single item")]
+    Edit {
+        #[command(subcommand)]
+        cmd: EditCommand,
+    },
+    #[command(long_about = "Delete one or more items")]
+    Delete {
+        #[command(subcommand)]
+        cmd: DeleteCommand,
+    },
```
Hinton, line 85: Starting to dislike the grouping of commands. Feels like we should group by type and not action, but that's outside the scope of this feature.

crates/bws/src/main.rs
```diff
@@ -89,6 +105,41 @@ enum GetCommand {
     Secret { secret_id: String },
 }
 
+#[derive(Subcommand, Debug)]
+enum CreateCommand {
+    Secret {
+        key: String,
+        value: String,
+
+        /// An optional note to add to the secret
```
Hinton, line 114: Should we move this to be a description of the arg instead of a comment?

Hinton (commented):
Looks good some small comments.

crates/bitwarden-napi/src-ts/bitwarden_client/schemas.ts
```diff
@@ -254,7 +254,11 @@ export interface SecretCreateRequest {
      * Organization where the secret will be created
      */
     organizationId: string;
-    value:          string;
+    /**
+     * IDs of the projects that this secret will belong to
+     */
+    projectIds?: string[] | null;
```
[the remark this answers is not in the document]
coltonhurst: @Hinton thanks for pointing this out. I spoke with @patrick-bitwarden and @dani-garcia. We want to allow the ability to pass in multiple project ids via the bws cli, though it's not supported _yet_ in the backend. Patrick recommended we show a more friendly error instead of one from the server, so I created an internal ticket for it.

crates/bws/src/main.rs
dani-garcia: Done

coltonhurst (approved):
Looks solid and other items brought up are addressed in other tickets / pr's.

Thanks @dani-garcia!

PR #32, opened 2023-04-17
PM-1727 Add fingerprint

Hinton:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

PM-1727

We needed a feature that is exposed in all applications to experiment with integrating the SDK in multiple environments. The user fingerprint is a isolated standalone functionality and seemed a good candidate for this.

## Before you submit


crates/bitwarden/src/crypto.rs
```diff
@@ -278,13 +282,51 @@ pub(crate) fn stretch_key(secret: [u8; 16], name: &str, info: Option<&str>) -> S
...
+    let minimum_entropy = 64;
+
+    let entropy_per_word = (EFF_LONG_WORD_LIST.len() as f64).log2();
+    let num_words = ((minimum_entropy as f64 / entropy_per_word).ceil() as f64).to_owned() as i64;
+
+    let hash_arr: Vec<u8> = hash.to_vec();
+    let entropy_available = hash_arr.len() * 4;
+    if num_words as f64 * entropy_per_word > entropy_available as f64 {
+        return Err(Error::Internal("Output entropy of hash function is too small"));
+    }
```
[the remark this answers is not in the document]
dani-garcia: Numbers on JS are always 64bit floating point so I'd keep it the same, but I don't expect the precision loss of using f32 would have any effect on this.

We might even get away with using integer numbers, but not sure how the rounding might go here.
```rust
let entropy_per_word = EFF_LONG_WORD_LIST.len().ilog2();
let num_words = minimum_entropy / entropy_per_word;

let hash_arr: Vec<u8> = hash.to_vec();
let entropy_available = hash_arr.len() * 4;
if num_words * entropy_per_word > entropy_available as u32 {
...
```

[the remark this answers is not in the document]
Hinton: `entropy_per_word` is roughly 12.9 in the test, I think we would have to round up to be certain we have enough entropy.

dani-garcia (dismissed):
Looks good to me, also tested it with a couple of users and the fingerprints match with the web vault

crates/bitwarden/src/sdk/request/fingerprint_request.rs
```diff
@@ -0,0 +1,11 @@
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Debug, JsonSchema)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub struct FingerprintRequest {
+    /// The user's user id, used in the fingerprint generation process.
+    pub user_id: String,
```
MGibson1, line 8: Might want to make this more generic `fingerprint_material` or something. This is used to generate fingerprint from a users email, instead of ID when authenticating with another device.

languages/csharp/schemas.cs
```diff
@@ -97,6 +97,10 @@ public partial class ClientSettings
     /// Returns:
     /// [UserApiKeyResponse](crate::sdk::response::user_api_key_response::UserApiKeyResponse)
     ///
+    /// Get the users passphrase
```
MGibson1, line 100:
```suggestion
    /// Get the user's passphrase
```

MGibson1, line 100: man, this is some seriously unfortunate commenting, huh?

MGibson1 (approved):

PR #40, opened 2023-04-24
Build aarch64 for linux

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

We should switch to using cross for cross-compiling the CLI.

## Code changes

## Before you submit


Hinton:
Seems we need to switch over to using cross or something for this to work properly.

PR #44, opened 2023-04-27
Rework the WASM code

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

The objective of this PR is to:

- Make WASM build produce an expected package.
- Investigate if we can share BitwardenClient between WASM and NAPI.

## Before you submit


Cargo.toml
```diff
@@ -21,3 +21,8 @@ codegen-units = 1
 # Stripping the binary reduces the size by ~30%, but the stacktraces won't be usable anymore. 
 # This is fine as long as we don't have any unhandled panics, but let's keep it disabled for now
 # strip = true
+
```
dani-garcia, line 24: This profile doesn't seem used anywhere? We can replace `--release` for `--profile release-wasm` in the README instructions if it's useful

README.md
```diff
@@ -91,3 +91,35 @@ generation. In that case we excluded the file from being generated and manually
...
+Install `wasm32-unknown-unknown` rust target.
+
+```bash
+rustup target add wasm32-unknown-unknown
+```
+
+Install wasm-bindgen-cli
+
+```bash
+cargo install -f wasm-bindgen-cli
```
dani-garcia, line 106: I had to separately install `wasm-opt`, used `brew install binaryen` on Mac, or from the releases page: https://github.com/WebAssembly/binaryen/releases. Probably a good idea to mention it here as well

crates/bitwarden/Cargo.toml
```diff
@@ -19,40 +19,46 @@ secrets = []  # Secrets manager API
 internal = [] # Internal testing methods
 mobile = []   # Mobile-specific features
 
+stdweb = ["instant/stdweb"]
```
dani-garcia, line 22: This seems unused as well.

dani-garcia (commented):
Some small comments, we're also now using `chrono` to handle time serialization, I think it would be a good idea to also use it where we are using `std::time::Instant`. It should be WASM compatible as well and we won't need a separate `instant` library just for this use case. I can submit a PR with just that change you want.

vgrassia (approved):

crates/bitwarden/Cargo.toml
```diff
@@ -37,6 +39,7 @@ chrono = { version = ">=0.4.26, <0.5", features = [
 getrandom = { version = ">=0.2.9", features = ["js"] }
 hkdf = ">=0.12.3, <0.13"
 hmac = ">=0.12.1, <0.13"
+instant = ">=0.1.12, <0.2"
```
dani-garcia, line 42: Now that we changed the token renewal instants to timestamps this dependency isn't needed anymore, but we probably need to add the `wasmbind` feature to `chrono`

README.md
```diff
@@ -97,3 +97,39 @@ generation. In that case we excluded the file from being generated and manually
...
+```bash
+# dev
+cargo build -p bitwarden-wasm -p bitwarden --target wasm32-unknown-unknown --features wasm-bindgen
+wasm-bindgen --target bundler --out-dir languages/js/wasm ./target/wasm32-unknown-unknown/debug/bitwarden_wasm.wasm
+wasm-bindgen --target nodejs --out-dir languages/js/wasm/node ./target/wasm32-unknown-unknown/debug/bitwarden_wasm.wasm
+
+# release
+cargo build -p bitwarden -p bitwarden-wasm --target wasm32-unknown-unknown --features wasm-bindgen --release
+wasm-bindgen --target bundler --out-dir languages/js/wasm ./target/wasm32-unknown-unknown/release/bitwarden_wasm.wasm
+wasm-bindgen --target nodejs --out-dir languages/js/wasm/node ./target/wasm32-unknown-unknown/release/bitwarden_wasm.wasm
```
dani-garcia, line 126: Are we going to support using the wasm builds in node while also offering the node-api support?

Hinton: I would like to offer it at least while we explore if we can transition off the napi bindings internally.

dani-garcia (dismissed):

dani-garcia (approved):

vgrassia (approved):

PR #49, opened 2023-05-18
Update swagger bindings to support uuids

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Update swagger bindings to support uuids, separated this from the state management PR to make it easier to review
 while the state is being reworked a bit 

## Code changes
- **crates/bitwarden_api_{api,identity}:** Code generated from the swagger specification
- **crates/{bitwarden,bws}:** Updated our code manually to use UUIDs and fix compile errors from the updated bindings
- **openapitools.json:** Updated the swagger code generator version, the only differences are enum variant names are `Variant0` instead of `_0` and identifiers with reserved keywords now use raw identifiers (`r#type` instead of `_type`)
- **support/openapi-template/*:** Updated the templates to add the uuid dependency and types

## Before you submit


crates/bitwarden/src/commands/login.rs
```diff
@@ -234,7 +238,7 @@ pub(crate) async fn renew_token(client: &mut Client) -> Result<()> {
                 client_secret,
                 ..
             } => {
-                AccessTokenRequest::new(&service_account_id, &client_secret)
+                AccessTokenRequest::new(*service_account_id, &client_secret)
```
Hinton, line 241: Haven't seen `*` used before, how does that work?

crates/bitwarden-api-api/src/models/update_domains_request_model.rs
```diff
@@ -16,7 +16,7 @@ pub struct UpdateDomainsRequestModel {
         rename = "excludedGlobalEquivalentDomains",
         skip_serializing_if = "Option::is_none"
     )]
-    pub excluded_global_equivalent_domains: Option<Vec<i32>>,
+    pub excluded_global_equivalent_domains: Option<Vec<crate::models::GlobalEquivalentDomainsType>>,
```
Hinton, line 19: We should decouple the server response/requests from this enum in the future. No need to be aware of the 70 equivalent domains.

support/openapi-template/hyper/api.mustache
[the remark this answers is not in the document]
dani-garcia: Sounds good, I will try to pull the new upstream changes and see what can be upstreamed after.

crates/bitwarden/src/commands/login.rs
dani-garcia: It's a dereference operator, the reverse of &.

We have a &Uuid and want to make it an Uuid, because Uuid is Copy, we can just use * which will copy the value. If it wasn't copy we'd be using clone() or to_owned() instead

PR #51, opened 2023-05-19
[SM-777] Rename the "password_login" command to match the verb phrasing

gbubemismith:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [X] Tech debt (refactoring, code cleanup, dependency upgrades, etc)
- [ ] Other
```

## Objective

Rename the `password_login` command to `login_with_password` to match the verb phrasing of the other commands.
## Code changes

- **crates/bitwarden/src/commands/login.rs:** Renamed function name to `login_with_password`.
- **crates/bitwarden/src/client/client.rs:** Updated file with renamed function name.

## Before you submit

### DIFF
--- a/crates/bitwarden-json/src/client.rs
+++ b/crates/bitwarden-json/src/client.rs
@@ -39,7 +39,7 @@ impl Client {
         };
 
         match cmd {
-            Command::PasswordLogin(req) => self.0.password_login(&req).await.into_string(),
+            Command::LoginWithPassword(req) => self.0.login_with_password(&req).await.into_string(),
             Command::AccessTokenLogin(req) => self.0.access_token_login(&req).await.into_string(),
             Command::GetUserApiKey(req) => self.0.get_user_api_key(&req).await.into_string(),
             Command::ApiKeyLogin(req) => self.0.api_key_login(&req).await.into_string(),
--- a/crates/bitwarden/src/client/client.rs
+++ b/crates/bitwarden/src/client/client.rs
@@ -11,8 +11,8 @@ use crate::{
         encryption_settings::{EncryptionSettings, SymmetricCryptoKey},
     },
     commands::{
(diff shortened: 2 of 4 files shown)


Hinton (changes requested):
I don't think we need the preposition `with`. Typically prepositions are omitted in function names unless they add significant meaning to the term. In this case `LoginPassword` and `LoginWithPassword` infers the same meaning.

If we want to do this change then we need to ensure our function names are consistent. Access tokens still use `AccessTokenLogin` which should now be `LoginAccessToken`.

Since this would be a breaking change in the API definition, we should take the opportunity and extract the login commands into an `Auth` group which would result in something like `Auth::LoginPassword` and `Auth::LoginAccessToken`.

Hinton:
@dasien The methods aren't working and should to my knowledge not be in use. I've opened a PR which hides them from the generated rust documentation https://github.com/bitwarden/sdk/pull/54, and we will look into putting all unused methods behind a compile time feature flag https://doc.rust-lang.org/cargo/reference/features.html.

With that said, I don't believe it's currently worth renaming any of the existing methods before we figure out the exact interface we want to use. Breaking changes are acceptable in pre 0. crates though, however a minor bump is usually good manner.

dasien:
@Hinton Sounds like we don't have an issue since the change won't be breaking anyone. 

With regard to calling style and interface definition, i am in favor of discrete methods rather than a CLI style single entry point for functions.  "LoginWithPassword", "LoginWithAccessToken", etc. with parameters relevant to that call, rather than "Login" which takes a structure of parameters which the method uses to determine which actual login method to call internally.  

Where we can reuse logic by having private methods which do the actual work, i think we should do that, but the interface/facade presented to the user should be clear and discrete methods which don't require the developer to look up how to create a payload of parameters.

Open to other thoughts though.

Hinton:
Closing this for now. We've gone through a few refactors and will look into auth methods in a bit.

PR #53, opened 2023-05-24
SM-417: Add project create, update, and delete commands

coltonhurst:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Adds the ability to create, update, and delete projects via the Bitwarden SDK.

## Code changes

## Before you submit

### DIFF
--- a/crates/bitwarden/src/commands/projects.rs
+++ b/crates/bitwarden/src/commands/projects.rs
@@ -1,9 +1,21 @@
+use bitwarden_api_api::models::{ProjectCreateRequestModel, ProjectUpdateRequestModel};
+
 use crate::{
     client::Client,
     error::{Error, Result},
     sdk::{
-        request::projects_request::{ProjectGetRequest, ProjectsListRequest},
-        response::projects_response::{ProjectResponse, ProjectsResponse},
+        request::projects_request::{
+            ProjectCreateRequest,
+            ProjectGetRequest,
+            ProjectsListRequest,
+            ProjectPutRequest,
+            ProjectsDeleteRequest
+        },
+        response::projects_response::{
+            ProjectResponse,
+            ProjectsResponse,
+            ProjectsDeleteResponse
+        },
     },
 };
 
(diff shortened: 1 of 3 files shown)


dani-garcia:
Looks like a great start, thanks @coltonhurst!

A small thing missing, we have to add the new create, update, delete commands to the big command struct here: https://github.com/bitwarden/sdk/blob/9e2e04c1b3a81505c1daeaab85486df1359bad9d/crates/bitwarden/src/sdk/request/command.rs#L120-L127

And then use it here: 
https://github.com/bitwarden/sdk/blob/9e2e04c1b3a81505c1daeaab85486df1359bad9d/crates/bitwarden-json/src/client.rs#L57

That way the new commands can be used both directly with the rust crate but also from the JSON wrapper and all the language bindings.

dani-garcia (approved):

crates/bitwarden/src/sdk/request/projects_request.rs
```diff
@@ -9,9 +9,36 @@ pub struct ProjectGetRequest {
...
+pub struct ProjectCreateRequest {
+    /// Organization where the project will be created
+    pub organization_id: Uuid,
+
+    pub name: String,
+}
+
+#[derive(Serialize, Deserialize, Debug, JsonSchema)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub struct ProjectPutRequest {
```
Hinton, line 23: Should this be called updateRequest? `Put` feels like an implementation detail.

crates/bitwarden/src/commands/projects.rs
```diff
@@ -23,6 +28,37 @@ pub(crate) async fn get_project(
...
+        &config.api,
+        input.organization_id,
+        project,
+    )
+    .await?;
+
+    let enc = client
+        .get_encryption_settings()
+        .as_ref()
+        .ok_or(Error::VaultLocked)?;
```
Hinton, line 57: Isn't `enc` already defined above? I think we can reuse it.

crates/bitwarden/src/commands/projects.rs
```diff
@@ -41,3 +77,43 @@ pub(crate) async fn list_projects(
...
+
+    let config = client.get_api_configurations().await;
+    let res =
+        bitwarden_api_api::apis::projects_api::projects_id_put(&config.api, input.id, project)
+            .await?;
+
+    let enc = client
+        .get_encryption_settings()
+        .as_ref()
+        .ok_or(Error::VaultLocked)?;
```
Hinton, line 104: Same

Hinton (commented):
Two minor comments, looks good otherwise!

crates/bitwarden/src/sdk/request/projects_request.rs
coltonhurst: This was to follow the convention in [secrets_request.rs](https://github.com/bitwarden/sdk/blob/master/crates/bitwarden/src/sdk/request/secrets_request.rs) which uses `SecretPutRequest`. We definitely can refactor out `Put` in multiple locations and use `Update` though. This could be done in a different ticket, this one, or we take a different approach?

crates/bitwarden/src/commands/projects.rs
coltonhurst: Hmm, any recommendations on a better way to do this? By removing it we break one of the borrowing rules. With it removed we have the following error, the compiler delineates the following:
- immutable borrow on line 35/36
- mutable borrow on line 46
- immutable borrow on line 54

`cannot borrow *client as mutable because it is also borrowed as immutable`

crates/bitwarden/src/commands/projects.rs
coltonhurst: Same as [here](https://github.com/bitwarden/sdk/pull/53#discussion_r1227286814)

crates/bitwarden/src/commands/projects.rs
coltonhurst: Discussed with @dani-garcia, addressing this in this PR: https://github.com/bitwarden/sdk/pull/70

crates/bitwarden/src/sdk/request/projects_request.rs
coltonhurst: Discussed this, we will address this in a separate PR

PR #54, opened 2023-05-25
Hide internal methods from docs

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Hide internal methods from the public docs. We should put these behind a features flag, but that's slightly more invasive.

## Before you submit


Hinton:
https://github.com/bitwarden/sdk/pull/56

PR #56, opened 2023-05-25
Add feature flag for internal commands

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Move all experimental portions behind a internal feature flag to prevent accidental usage.

## Before you submit

### DIFF
--- a/crates/bitwarden-json/Cargo.toml
+++ b/crates/bitwarden-json/Cargo.toml
@@ -13,6 +13,9 @@ categories = ["api-bindings"]
 edition = "2021"
 rust-version = "1.57"
 
+[features]
+internal = [] # Internal testing methods
+
 [dependencies]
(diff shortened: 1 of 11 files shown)


crates/bitwarden-json/Cargo.toml
```diff
@@ -13,6 +13,9 @@ categories = ["api-bindings"]
 edition = "2021"
 rust-version = "1.57"
 
+[features]
+internal = [] # Internal testing methods
```
dani-garcia, line 17:
```suggestion
internal = ["bitwarden/internal"] # Internal testing methods
```

We need to pass through the internal feature otherwise trying to build `bitwarden-json` with `internal` by itself will fail

PR #58, opened 2023-05-30
Update the SDK Example

coltonhurst:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Update the SDK example in the README now that we are using the `uuid` crate. Specifically, the `SecretIdentifiersRequest.organization_id`'s type is now a `Uuid`.

## Code changes

- **README.md:** Update the readme example
- **lib.rs:** Add a check to test the readme docs

## Before you submit

### DIFF
--- a/crates/bitwarden/README.md
+++ b/crates/bitwarden/README.md
@@ -28,6 +28,7 @@ use bitwarden::{
         },
     },
 };
+use uuid::Uuid;
 
 async fn test() -> Result<()> {
     // Use the default values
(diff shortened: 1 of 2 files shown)


crates/bitwarden/README.md
```diff
@@ -46,7 +47,8 @@ async fn test() -> Result<()> {
     let token = AccessTokenLoginRequest { access_token: String::from("") };
     client.access_token_login(&token).await.unwrap();
 
-    let org_id = SecretIdentifiersRequest { organization_id: String::from("") };
+    // The "v4" feature must be enabled to use `new_v4()` in the uuid crate
+    let org_id = SecretIdentifiersRequest { organization_id: Uuid::new_v4() };
```
dani-garcia, line 51: I think we should probably show an example parsing the UUID instead of generating a random one, like in `lib.rs`. I assume that's going to be the most common use case:

https://github.com/bitwarden/sdk/blob/729266454f48f9773802f714404892a614765269/crates/bitwarden/src/lib.rs#L49

To avoid having broken code in the Readme in the future, we can add something like this in the readme:
```rust
#[doc = include_str!("../README.md")]
mod readme {}

```

coltonhurst: Sounds good, and great idea. Updated 👍

coltonhurst:
For posterity, the ticket for this was created post-merge: SM-800

PR #64, opened 2023-06-08
[PM1724] State Management v2

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
Version two of the state service implementation, now supporting more granular changes (instead of just overwriting the whole profile) and using file locks as well.

Also implemented state services with different views over the global state to isolate the changes a bit better, this is put to the test on a separate PR (https://github.com/bitwarden/sdk/pull/65) handling folder creation/update/deletion. Example:
```rust
// Create a new folder
client
    .get_state_service(FOLDERS_SERVICE)
    // This applies the change in the callback to the state in a synchronised way to avoid overwriting other changes
    .modify(move |folders| { // `folders` here is a mutable reference to the Folders HashMap
        folders.insert(id, Folder { id, name, revision_date });
        Ok(())
    })
    .await?;
```

## Code changes
The state code is stored inside `crates/bitwarden/src/state/`, inside there you can find the following:
- `state.rs`: Contains all the state handling code, and is mostly meant to be generic over state contents and storage medium, to implement another storage type, like secure storage, biometric storage, etc we'd just need to implement the trait `StateStorageMedium`. Currently implemented are file based storage for non-WASM and LocalStorage for WASM.
- `state_service.rs`: Contains the implementation for the different state "views" that will allow the services to access only the part of the state that they require. Check the separate Folders PR (https://github.com/bitwarden/sdk/pull/65) for an example on how it's used for the Folders service, there's also `src/client/[auth,keys,profile].rs`, which handle some parts of the state that are used internally.
- The `sync` function has the response type removed, the response is processed internally by the state handling code, and clients should access the data through there, instead of through the raw sync response.
- `Client` has been modified to move most of it's internal state into the new state management solution. This has caused some functions to need to become async, to handle the synchronised access to the state. 
- The client now also has methods to access the state, plus a `session_login` method to restore the client from a session persisted to disk. At the moment, `session_login` also forces a vault unlock, though in the future we probably want to create separate `lock` and `unlock` functionality.


crates/bitwarden-json/src/command.rs
```diff
@@ -32,7 +35,7 @@ pub enum Command {
     ///
     /// This command is not capable of handling authentication requiring 2fa or captcha.
     ///
-    /// Returns: [PasswordLoginResponse](crate::sdk::auth::response::PasswordLoginResponse)
+    /// Returns: [PasswordLoginResponse](bitwarden::auth::response::PasswordLoginResponse)
```
Hinton, line 38: Can you please move all of these changes into a separate PR?

crates/bitwarden/src/client/auth_settings.rs
```diff
@@ -2,46 +2,83 @@ use std::num::NonZeroU32;
...
+
+                memory: response
+                    .kdf_memory
+                    .and_then(|e| NonZeroU32::new(e as u32))
+                    .unwrap_or_else(default_argon2_memory),
+                parallelism: response
+                    .kdf_parallelism
+                    .and_then(|e| NonZeroU32::new(e as u32))
+                    .unwrap_or_else(default_argon2_parallelism),
+            },
```
Hinton, line 58: Can we move the argon2 work to a separate PR?

crates/bws/src/main.rs
```diff
@@ -461,7 +461,7 @@ async fn process_commands() -> Result<()> {
         } => {
             let secret = client
                 .secrets()
-                .create(&SecretCreateRequest {
+                .create(SecretCreateRequest {
```
Hinton, line 464: Having a somewhat hard time following, why did this change from a reference? We seem to be changing it shortly afterwards. Maybe something for a separate PR?

crates/bitwarden-json/src/command.rs
dani-garcia: Moved into https://github.com/bitwarden/sdk/pull/80

crates/bws/src/main.rs
dani-garcia: Not sure why those were there, they aren't needed, I've removed them now

crates/bitwarden-json/src/response.rs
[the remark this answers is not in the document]
dani-garcia: Unrelated minor thing indeed, I've separated it into https://github.com/bitwarden/sdk/pull/81

crates/bitwarden-json/src/command.rs
```diff
@@ -131,39 +139,70 @@ pub enum ProjectsCommand {
...
+    ///
+    Create(FolderCreateRequest),
+
+    /// > Requires Authentication
+    /// > Requires an unlocked vault and calling Sync at least once
+    /// Lists all folders in the vault
+    ///
+    /// Returns: [FoldersResponse](bitwarden::platform::folders::FoldersResponse)
+    ///
+    List(EmptyRequest),
```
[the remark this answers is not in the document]
dani-garcia: If the enums are empty, they are handled in a different way for the schema generation, from example in the TS side:
```diff
    projects?:         ProjectsCommand;
-   folders?:          FoldersCommand;
+   folders?:          FoldersCommandClass | FoldersCommandEnum;
}

....

+export enum FoldersCommandEnum {
+   List = "list",
+}
```

FoldersCommand gets renamed to FoldersCommandClass and mostly stays the same, but the List action gets added to a separate enum. I feel like this would overcomplicate things for the few cases that don't require inputs

crates/bitwarden/src/client/auth_settings.rs
dani-garcia: Removed from here, PR in progress now

crates/bitwarden/src/state/domain.rs
Hinton, line 1: Not sure if I like the filename of this. Ideally different models should live in different files.

crates/bitwarden/src/state/domain.rs
```diff
@@ -0,0 +1,188 @@
...
+}
+
+#[derive(Default, Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct AccountData {
+    pub keys: Option<Keys>,
+    pub profile: Option<Profile>,
+
+    pub ciphers: HashMap<Uuid, Cipher>,
+    pub folders: HashMap<Uuid, Folder>,
```
Hinton, line 41: I would like to have a generic way to assign data to an account, so that we don't necessarily fill this with 20 hashmaps, and everyone has to constantly modify it.

That would also allow the consumers to store data without needing to maintain a separate system for it.

crates/bitwarden/src/state/domain.rs
```diff
@@ -0,0 +1,188 @@
...
+    pub id: Uuid,
+    pub organization_id: Option<Uuid>,
+    pub folder_id: Option<Uuid>,
+
+    pub name: CipherString,
+
+    pub creation_date: DateTime<Utc>,
+    pub deleted_date: Option<DateTime<Utc>>,
+    pub revision_date: DateTime<Utc>,
+}
```
Hinton, line 93: Can we remove cipher from this PR?

crates/bitwarden/src/state/domain.rs
Hinton, line 1: I would like us to try and trim down the number of models we have initially, and just focus on getting things right.

crates/bitwarden/src/state/domain.rs
dani-garcia: This struct wasn't in use in the last revision of the PR, it remained there from when the state was handled as a single block. I've deleted it now.

I think what you want to do can be done with `client .get_state_service(FOLDERS_SERVICE)`, by defining any service, it's just a type and a namespace. It's not exposed outside of the SDK but it can be done, we'd just need to avoid conflicting namespaces.

crates/bitwarden/src/state/domain.rs
dani-garcia: Removed it

crates/bitwarden/src/state/domain.rs
dani-garcia: I've trimmed it to the minimum, and removed all the folder related code now, we can merge that later with some of the typestate changes Matt has on his PR

dani-garcia: Yeah, I don't think it's a great place for them, I've renamed the file to models which I think is somewhat better, and I'll see where else we can move those models to, the only remaining ones are the Encryption Keys, Profile and Auth tokens, so I guess the best place to put them would be `auth` and `platform`? Or maybe `client`, next to `AuthSettings` and `EncryptionSettings`?

crates/bitwarden-json/Cargo.toml
```diff
@@ -17,7 +17,7 @@ rust-version = "1.57"
 internal = ["bitwarden/internal"] # Internal testing methods
 
 [dependencies]
-schemars = ">=0.8.12, <0.9"
+schemars = { version = ">=0.8.12, <0.9", features = ["uuid1", "chrono"] }
```
Hinton, line 20: Do we need `uuid` here?

Hinton:
We have 36 warnings building this. I think we need to put more things behind flags.

crates/bitwarden-json/Cargo.toml
dani-garcia: I've added it because the `SessionLogin` command contains an UUID internally, but that's defined in the `bitwarden` crate, so it's not needed here indeed.

Technically we're inheriting both features from the `bitwarden` crate, so we could remove them both to avoid duplication anyway.

dani-garcia:
A few of those warnings are on the master branch, I'll make a new PR to fix them

crates/bitwarden-json/Cargo.toml
Hinton: Hmm, if the json crate itself needs uuid1 we should keep it.

crates/bitwarden/src/state/state.rs
```diff
@@ -0,0 +1,293 @@
...
+                Ok(())
+            }))
+            .await?;
+
+        self.cache = result.get(&self.account_id).cloned().unwrap_or_default();
+        Ok(())
+    }
+}
+
+fn load_medium(path: &Option<String>) -> Box<dyn StateStorageMedium> {
```
Hinton, line 130: Is medium some terminology?

crates/bitwarden/src/state/state.rs
```diff
@@ -0,0 +1,293 @@
...
+        &mut self,
+        f: Box<dyn for<'a> FnOnce(&'a mut StateMap) -> Result<()> + Send + 'b>,
+    ) -> Result<StateMap>;
+}
+
+#[cfg(not(target_arch = "wasm32"))]
+#[derive(Debug)]
+struct FileStateStorageMedium {
+    path: PathBuf,
+}
```
Hinton, line 160: We should probably move out wasm code to it's own file to easier conditionally compile it.

crates/bitwarden/src/state/state.rs
```diff
@@ -0,0 +1,293 @@
...
+        #[cfg(target_arch = "wasm32")]
+        {
+            Box::new(WasmLocalStorageMedium::new(path.to_owned()))
+        }
+        #[cfg(not(target_arch = "wasm32"))]
+        {
+            Box::new(FileStateStorageMedium::new(path.into()))
+        }
+    } else {
+        Box::new(InMemoryStateStorageMedium::new())
```
Hinton, line 141: Why do we need a in memory state?

Hinton (commented):
Did an initial review.

There are two architectural changes in this PR.
1. Introducing `state_services` which are loosely stored on the client object.
2. State persistent mechanism.

I will need to do some deeper digging into this to properly understand everything.

crates/bitwarden/src/state/state.rs
```diff
@@ -0,0 +1,293 @@
+use std::{collections::HashMap, fmt::Debug, path::PathBuf};
+
+use async_lock::Mutex;
```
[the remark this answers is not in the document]
dani-garcia: We don't actually use `tokio` on the `bitwarden` library crate, for compatibility reasons with WASM

crates/bitwarden/src/auth/commands/login.rs
```diff
@@ -218,17 +263,19 @@ async fn request_access_token(
 }
 
 pub(crate) async fn renew_token(client: &mut Client) -> Result<()> {
-    const TOKEN_RENEW_MARGIN: Duration = Duration::from_secs(5 * 60);
+    let token_renew_margin = chrono::Duration::seconds(5 * 60);
```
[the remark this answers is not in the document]
dani-garcia: Previously it was const, so it was only computed once, `chrono`'s `Duration::seconds` is not const, so we can't do the same here.

Moving it outside the function would mean making it static by using something like `Lazy`, but I don't think it's worth the effort, `Duration::seconds` is only checking if the number is out of bounds so it should be pretty fast.

crates/bitwarden/src/client/client.rs
```diff
@@ -166,15 +168,15 @@ impl Client {
     }
 
     #[cfg(feature = "internal")]
-    pub(crate) fn get_auth_settings(&self) -> &Option<AuthSettings> {
-        &self.auth_settings
+    pub(crate) async fn get_auth_settings(&self) -> Option<AuthSettings> {
+        Auth::get(self).await.kdf
     }
```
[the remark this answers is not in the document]
dani-garcia: Yeah, I feel like it should be `kdf_settings` and the struct renamed to `KdfSettings`. It only holds the email and KDF parameters, it's only tangentially related to authentication. What do you think?

crates/bitwarden/src/client/auth.rs
```diff
@@ -0,0 +1,60 @@
...
+
+#[derive(Default, Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Auth {
+    pub access_token: String,
+    pub refresh_token: Option<String>,
+    pub token_expiration: Option<chrono::DateTime<Utc>>,
+    pub login_method: Option<LoginMethod>,
+
+    pub kdf: Option<AuthSettings>,
```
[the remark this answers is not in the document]
dani-garcia: Even though these and the struct itself are `pub`, the module is declared as `pub(crate)` in the `mod.rs` file, so it's not really public, but I'll change it to keep it consistent.

crates/bitwarden/src/crypto.rs
```diff
@@ -69,6 +70,15 @@ impl std::fmt::Debug for CipherString {
...
 }
 
+impl schemars::JsonSchema for CipherString {
+    fn schema_name() -> String {
+        "CipherString".to_owned()
+    }
+    fn json_schema(gen: &mut schemars::gen::SchemaGenerator) -> schemars::schema::Schema {
+        gen.subschema_for::<String>()
+    }
+}
```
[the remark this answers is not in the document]
dani-garcia: I think the idea was being able to generate a schema of the internal state, back when I just defined it as a single big struct, now it probably doesn't make much sense.

crates/bitwarden/src/client/client.rs
```diff
@@ -86,6 +84,8 @@ impl Client {
             .build()
             .unwrap();
 
+        let state = State::load_state(&settings);
```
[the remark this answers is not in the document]
dani-garcia: This doesn't actually load and parse the file yet, it's just initializing the state with the correct path and everything. Probably worth renaming to `initialize_state`.

The file would be loaded in the current `session_login`, by the line `client.state.load_account(input.user_id).await?;`.

I'll do some testing, but I expect it to be plenty fast.

crates/bitwarden/src/state/state.rs
dani-garcia: I've used storage medium to refer to the different storage backends, in a similar way than CD's or DVD's would be different storage media.

It doesn't really add anything to the name though, we can just rename the `StateStorageMedium` trait to `Storage` plainly, and rename `load_medium` to `get_storage` (Because it's not actually loading anything yet).

crates/bitwarden/src/auth/commands/login.rs
```diff
@@ -146,23 +152,62 @@ pub(crate) async fn access_token_login(
...
+    use crate::client::{keys::Keys, profile::Profile};
+
+    client.state.load_account(input.user_id).await?;
+
+    let Some(profile) = Profile::get(client).await else {return Err(Error::NotAuthenticated)};
+    let auth = Auth::get(client).await;
+
+    let Some(expires) = auth.token_expiration else {return Err(Error::VaultLocked)};
+    let Some(login_method) = auth.login_method else {return Err(Error::VaultLocked)};
+    let expires_seconds = (expires.timestamp() - chrono::Utc::now().timestamp()) as u64;
```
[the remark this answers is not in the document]
dani-garcia: Just realised that calling `set_tokens` here didn't make sense. `set_tokens` both sets the access token for the API clients, and also stores the access and refresh tokens and their expiration in the state.

In this case we are only interested in the first part, it doesn't make sense here to save it in the state when we just read it from state, so I've separated that part to a new `set_access_token`, which means we don't have to use the expiration here at all.

crates/bitwarden/src/auth/commands/login.rs
```diff
@@ -146,23 +152,62 @@ pub(crate) async fn access_token_login(
...
+pub(crate) async fn session_login(client: &mut Client, input: &SessionLoginRequest) -> Result<()> {
+    use crate::client::{keys::Keys, profile::Profile};
+
+    client.state.load_account(input.user_id).await?;
+
+    let Some(profile) = Profile::get(client).await else {return Err(Error::NotAuthenticated)};
+    let auth = Auth::get(client).await;
+
+    let Some(expires) = auth.token_expiration else {return Err(Error::VaultLocked)};
+    let Some(login_method) = auth.login_method else {return Err(Error::VaultLocked)};
```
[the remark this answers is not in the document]
dani-garcia: This is removed with the `set_access_token` change mentioned below.

crates/bitwarden-json/src/client.rs
```diff
@@ -47,6 +47,8 @@ impl Client {
             Command::PasswordLogin(req) => self.0.password_login(&req).await.into_string(),
             Command::AccessTokenLogin(req) => self.0.access_token_login(&req).await.into_string(),
             #[cfg(feature = "internal")]
+            Command::SessionLogin(req) => self.0.session_login(&req).await.into_string(),
```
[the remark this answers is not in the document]
dani-garcia: Done in [1d43f45](https://github.com/bitwarden/sdk/pull/64/commits/1d43f45772e14c0ddb78448cf14a66eb9bfb9c48)

crates/bitwarden/src/crypto.rs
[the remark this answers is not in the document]
dani-garcia: Removed in [22c4a74](https://github.com/bitwarden/sdk/pull/64/commits/22c4a742725ea0076dc40e42894cad5c69cfecb0)

PR #65, opened 2023-06-09
Typestate

MGibson1:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Still a WIP, especially regarding gettings schemas properly generated, but I wanted to get you some actual code tonight, @dani-garcia. Let me know what you think about this direction.## Before you submit


dani-garcia:
I've rebased this against the latest state PR and implemented a few things that were missing, but with this we have an implementation of the folders service.

PR #66, opened 2023-06-09
PW Auth

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Code changes

## Before you submit


crates/bw/src/config.rs
Hinton, line 1: Would be nice to share this code between the bws and bw packages.

dani-garcia (approved):

PR #67, opened 2023-06-09
SM-784: Enable Project Create/Edit/Delete CLI Commands

coltonhurst:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Based on SM-417 and [this PR](https://github.com/bitwarden/sdk/pull/30).

This PR adds the create, edit, and delete commands for projects to the `bws` cli.

## Code changes

- **main.rs:** Add CLI functionality for creating, editing, and deleting projects

## Before you submit


crates/bws/src/main.rs
```diff
@@ -90,6 +108,26 @@ enum GetCommand {
...
+#[derive(Subcommand, Debug)]
+enum CreateCommand {
+    Project { name: String },
+}
+
+#[derive(Subcommand, Debug)]
+enum EditCommand {
+    #[clap(group = ArgGroup::new("edit_field").required(true).multiple(true))]
+    Project {
+        project_id: String,
```
dani-garcia, line 120: If you change this type to be Uuid here directly, you won't need to manually do Uuid::parse_str afterwards

crates/bws/src/main.rs
```diff
@@ -90,6 +108,26 @@ enum GetCommand {
...
+        #[arg(long, group = "edit_field")]
+        name: String,
+    },
+}
+
+#[derive(Subcommand, Debug)]
+enum DeleteCommand {
+    Project { project_ids: Vec<String> },
```
dani-garcia, line 128: Same for this one

dani-garcia (commented):
Looks good, just two small changes

crates/bws/src/main.rs
coltonhurst: Great point, updated!

crates/bws/src/main.rs
coltonhurst: Great point, updated!

dani-garcia (approved):
Looks good!

PR #68, opened 2023-06-10
Restructure folder structure

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

The current folder structure isn't scalable and we should rework it.

This PR proposes that we create feature groups in the root of the `bitwarden` package. One such group is auth.

```
auth
  comands
  request
  response
  api
    request
    response
```

If needed we can group multiple features under a broader name, potentially the team name which allows for easy code ownership assignment.

## Before you submit


crates/bitwarden/src/auth/mod.rs
```diff
@@ -1,2 +1,4 @@
+pub mod api;
```
Hinton, line 1: This should probably not be public, however we use api response in captcha response. Might be worth figuring that out.

dani-garcia: Would using `pub(super) mod api;` help here? That way it'll only be accesible to the auth module.

dani-garcia:
Looks good overall to me, is the separation between `auth/[request,response]` and `auth/api/[request,response]` simply based on "this is the part that deals with client<->sdk" and "this is the part that deals with sdk<->server"?

Should we do something with the sdk module, that also has a lot of unrelated stuff inside?

Hinton:
Pretty much. Sdk is now gone.

MGibson1 (approved):

MGibson1 (approved):

PR #71, opened 2023-06-16
Enable some optimization flags

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
I've done some performance analysis on the SDK, by using a test crate that creates a `bitwarden_json` client to login to an account with 7k items, doing a sync and decrypting all the ciphers. This is based on my state management PR so note that not all the sync data is decrypted which means these numbers aren't representative of what the final result is going to be once the SDK is complete.

| Operation           | Before                | After |
|-----------------|-------|-------|
| Debug - /sync    | 21s   | **1.5s** |
| Debug - decrypt | 500ms | **52ms**|
| Debug - binary size | 24 MB | **8.8 MB** |
| Debug - clean build | **25s** | 38s |
| Debug - incremental build | 0,8s | 0,8s |

| Operation           | Before                | After |
|-----------------|-------|-------|
| Release - /sync    | 1.1s   | 1.0s|
| Release - decrypt | 43ms | 41ms|
| Release - binary size | 7.4 MB | **4.8 MB - 3.4 MB (stripped)** |
| Release - clean build | **31s** | 59s |

In debug, clean build times are about 1.5x slower, but incremental builds are the same, while the resulting binary is an order of magnitude faster.

In release, clean build times are about 2x slower, while runtime is the same. The resulting binary is 40% smaller, and could be improved further by stripping the binary symbols.

We can look into creating a separate CI profile with zero optimizations if the increased compilation time turns out to be too big for CI as well.

I've also tried to analyze memory usage of the binary while both in debug and release, and also before and after this change, and that seems pretty much constant, and just under 60MB maximum.
I've used the command `/usr/bin/time -l ../../target/release/test_json`, not sure how precise that is.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,3 +1,22 @@
 [workspace]
 
 members = ["crates/*"]
+
+# Compile all dependencies with some optimizations when building this crate on debug
+# This slows down clean builds by about 50%, but the resulting binaries can be orders of magnitude faster
+# As clean builds won't occur very often, this won't slow down the development process
+[profile.dev.package."*"]
+opt-level = 2
+
+# Turn on a small amount of optimisation in development mode. This might interfere when trying to use a debugger
+# if the compiler decides to optimize some code away, if that's the case, it can be set to 0 or commented out
+[profile.dev]
+opt-level = 1
+
+# Turn on LTO on release mode
+[profile.release]
+lto = "thin"
+codegen-units = 1
+# Stripping the binary reduces the size by ~30%, but the stacktraces won't be usable anymore. 
+# This is fine as long as we don't have any unhandled panics, but let's keep it disabled for now
+# strip = true


Cargo.toml
```diff
@@ -1,3 +1,22 @@
...
+# Compile all dependencies with some optimizations when building this crate on debug
+# This slows down clean builds by about 50%, but the resulting binaries can be orders of magnitude faster
+# As clean builds won't occur very often, this won't slow down the development process
+[profile.dev.package."*"]
+opt-level = 2
+
+# Turn on a small amount of optimisation in development mode. This might interfere when trying to use a debugger
+# if the compiler decides to optimize some code away, if that's the case, it can be set to 0 or commented out
+[profile.dev]
+opt-level = 1
```
MGibson1, line 14: what was the impact of this optimization? I'd prefer not to optimize our code, since that can be a difficult-to-understand problem as new contributors are onboarded

dani-garcia: Running the following operations with a vault of 7k items, using PBKDF2 with 600k rounds:
- Login with password:
  - All the optimizations: 1s
  - Only external dependencies: 6s
  - Previously: 12s

- Sync without decryption:
  - All the optimizations: 350ms
  - Only external dependencies: 670ms
  - Previously: 850ms

- Decryption of the items:
  - All the optimizations: 52ms
  - Only external dependencies: 276ms
  - Previously: 511ms

The biggest slow down when removing that optimization is during the login, entirely because of PBKDF2. This seems to be because those functions are defined as [inline] in the crate so the compiler doesn't run the optimizations that we have defined for external crates.

Hinton: I guess we could move the auth logic into a separate crate and optimize just that. I assume there's no way to tell rust to not inline the crate?

MGibson1: I don't think we can move to a separate crate, we need kdf derivations in a few places (password reprompt, change password) so that can't be a long-term solution.

Let's just try and make this a known possibility like you've done here.

MGibson1 (approved):

PR #72, opened 2023-06-16
Remove schemas from repo

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Having the schemas commited to the repo makes PR reviews cumbersome, when simply generating them when needed is very easy. 

This PR removes the schemas from the repo and adds them to `.gitignore`, while also updating CI so they are generated before running the NAPI builds.

Also enabled the preserve_order flag of schemars to make sure schema fields are generated in the same order as the rust structs


.github/workflows/build-napi.yml
```diff
@@ -84,6 +84,10 @@ jobs:
         with:
           key: ${{ matrix.settings.target }}-cargo-${{ matrix.settings.os }}
 
+      - name: Generate schemas
+        run: npm ci && npm run schemas
+        working-directory: .
+
```
Hinton, line 90: We re-run `npm ci` twice. We can put it below the next step instead.

dani-garcia: They are being run at different directories though:
The `Generate schemas` step is being run at the root of the repository, where `quicktype` code generation dependency is.
The `Install dependencies` step is being run inside the `crates/bitwarden-napi` folder, that's where the NAPI dependencies are.

MGibson1: Shouldn't schema generation be a separate workflow with a published artifact? We're going to want these schemas in more than just the napi builds

dani-garcia: Separated into a new workflow that publishes the schema artifacts, let me know what you think.

.github/workflows/release-napi.yml
```diff
@@ -131,6 +134,12 @@ jobs:
           cache: "npm"
           cache-dependency-path: crates/bitwarden-napi/package-lock.json
 
+      - name: Retrieve schemas
+        uses: actions/download-artifact@9bc31d5ccc31df68ecc42ccf4149144866c47d8a # v3.0.2
+        with:
+          name: schemas.ts
+          path: ${{ github.workspace }}/crates/bitwarden-napi/src-ts/bitwarden_client/
```
Hinton, line 141: I think we would already have these generated at this point? We download artifacts from the `build-napi` workflow.

.github/workflows/generate_schemas.yml
```diff
@@ -0,0 +1,64 @@
...
+        with:
+          cache: "npm"
+          cache-dependency-path: "package-lock.json"
+          node-version: "16"
+
+      - name: NPM setup
+        run: npm ci
+
+      - name: NPM Schemas
+        run: npm run schemas
```
Hinton, line 36: We should ensure we cache rust before running this.

Hinton (changes requested):
@bitwarden/dept-devops are there any better ways to only run the schema generation once?

.github/workflows/release-napi.yml
dani-garcia: Hmm that's likely the case yeah, though I'm not sure how to test the release workflows. I removed it for now

.github/workflows/generate_schemas.yml
dani-garcia: Done

.github/workflows/release-napi.yml
Hinton: I think this needs to be solved before this can be merged. I believe a good starting point is just to try and download them from the generate_schema workflow but not trigger it and have devops verify it looks good.

Hinton: I believe we have to update the code for the release workflow so that it downloads the already built schema.ts. Since it's a required part of the build process.

Hinton: Yes. I think that should work.

dani-garcia: Done in 28ab623e5f6bfadc3697bd550ac316cc5eab429c

Hinton: I think you need to specify some more fields to pick a workflow etc.

dani-garcia:
Seems like moving from `download-artifact` to `bitwarden/gh-actions/download-artifacts` made it work: https://github.com/bitwarden/sdk/actions/runs/5511628118/jobs/10047435090
```
Run bitwarden/gh-actions/download-artifacts@72594be690a4e7bfa87b1402b2aedc75acdbff12
==> Workflow: build-napi.yml
==> Repo: bitwarden/sdk
==> Conclusion: success
==> Branch: remove_schemas
==> RunID: 5511530350
==> Artifact: 795802[14](https://github.com/bitwarden/sdk/actions/runs/5511628118/jobs/10047435090#step:5:15)8
==> Downloading: schemas.ts.zip (35.77 kB)
  inflating: /home/runner/work/sdk/sdk/crates/bitwarden-napi/src-ts/bitwarden_client/schemas.ts
```

That said there's now an error in the `retrieve secrets` step:
```
Error: The user, group or application 'appid=***;oid=911d2394-b758-4484-a62e-1f0cb6cae22c;iss=https://sts.windows.net/***/' does not have secrets get permission on key vault 'bitwarden-prod-kv;location=eastus'. For help resolving this issue, please see https://go.microsoft.com/fwlink/?linkid=2125287
Error: Could not download the secret npm-api-key
```

I assume I just don't have enough permissions? Not sure if this step is even needed when using `Dry run`, but that's outside this PR.

.github/workflows/release-napi.yml
```diff
@@ -131,6 +131,26 @@ jobs:
           cache: "npm"
           cache-dependency-path: crates/bitwarden-napi/package-lock.json
 
+      - name: Download schemas
+        if: ${{ github.event.inputs.release_type != 'Dry Run' }}
+        uses: bitwarden/gh-actions/download-artifacts@72594be690a4e7bfa87b1402b2aedc75acdbff12
```
MGibson1, line 136: Add version comment

.github/workflows/release-napi.yml
```diff
@@ -131,6 +131,26 @@ jobs:
...
+        with:
+          workflow: build-napi.yml
+          artifacts: schemas.ts
+          path: ${{ github.workspace }}/crates/bitwarden-napi/src-ts/bitwarden_client/
+          workflow_conclusion: success
+          branch: ${{ github.ref_name }}
+
+      - name: Dry Run - Download schemas
+        if: ${{ github.event.inputs.release_type == 'Dry Run' }}
+        uses: bitwarden/gh-actions/download-artifacts@72594be690a4e7bfa87b1402b2aedc75acdbff12
```
MGibson1, line 146: Add version comment

MGibson1 (changes requested):
~~Just two qc things, but this looks like it's working to me.~~
nvm, seems we have some baggage around our own actions

.github/workflows/release-napi.yml
MGibson1: actually, since these are using the bitwarden actions repo, we should just remove the versioning to avoid the renovate PRs. we implicitly trust our own actions...

dani-garcia:
Another dry-run and it seems to be working now: 
https://github.com/bitwarden/sdk/actions/runs/5537239266

Hinton (approved):

PR #76, opened 2023-06-21
SM-815: Switch Command Order

coltonhurst:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Switch the command order for the bws cli. Instead of `bws create secret`, move to a more logical `bws secret create` structure.

We have left the existing commands but have removed them from the help menu and marked them as deprecated.

## Code changes

- **main.rs:** Update command order

## Before you submit

### DIFF
--- a/crates/bws/src/main.rs
+++ b/crates/bws/src/main.rs
@@ -62,41 +62,6 @@ struct Cli {
 
 #[derive(Subcommand, Debug)]
 enum Commands {
-    #[command(long_about = "Commands available on Secrets")]
-    Secret {
-        #[command(subcommand)]
-        cmd: SecretCommand,
-    },
-    #[command(long_about = "Commands available on Projects")]
-    Project {
-        #[command(subcommand)]
-        cmd: ProjectCommand,
-    },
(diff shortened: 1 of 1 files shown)


crates/bws/src/main.rs
```diff
@@ -97,6 +107,57 @@ enum Commands {
...
+
+        #[arg(long, help = "An optional note to add to the secret")]
+        note: Option<String>,
+
+        #[arg(long, help = "The ID of the project this secret will be added to")]
+        project_id: Option<Uuid>,
+    },
+    List {
+        project_id: Option<Uuid>,
+    },
```
Hinton, line 127: I think we should move List to the top, or bottom. The other commands works on a single entity so in my opinion it makes sense to group them. Or group everything alphabetically.

coltonhurst: In case we add other commands in the future that may be unrelated and harder to sort... I have sorted the enum variants in `bws/src/main.rs` alphabetically.

Hinton (approved):
Lgtm, thanks.

PR #77, opened 2023-06-26
SM-807: BWS CLI & SDK Updates

coltonhurst:
## Type of change

```
- [x] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

This PR covers the following items:
- `project_id` is now a required argument when creating a secret.
- Editing a secret was broken due to not handling the list of project id’s correctly. This has been fixed.
- A user is now able to update the project a secret is in if they choose to.
- The SecretCommand edit now allows multiple options to be set at the same time. This was forgotten on the initial command order change ([link](https://github.com/bitwarden/sdk/pull/76)).
- Proper error messaging has been added in the case that none, some, or all secret / project deletions fail

Error messages when deleting projects or secrets now follow this template:

```
X project(s) deleted successfully.
X project(s) had an error:
{id of project}: {server error message}
```

Line 1 should always print. Lines 2 and 3 should only print if there were any errors.

## Before you submit


crates/bws/src/main.rs
```diff
@@ -108,16 +108,15 @@ enum SecretCommand {
     Create {
         key: String,
         value: String,
+        project_id: Uuid,
 
         #[arg(long, help = "An optional note to add to the secret")]
         note: Option<String>,
-
-        #[arg(long, help = "The ID of the project this secret will be added to")]
```
Hinton, line 115: Did we intentionally drop the help text?

crates/bws/src/main.rs
```diff
@@ -516,12 +548,38 @@ async fn process_commands() -> Result<()> {
...
+            for secret in result.data.iter() {
+                match &secret.error {
+                    Some(error_message) => {
+                        secrets_failed.push((secret.id, error_message.to_owned()))
+                    }
+                    None => secrets_success += 1,
+                }
+            }
+
+            if secrets_success == 0 || secrets_success > 1 {
```
Hinton, line 568:
```suggestion
            if secrets_success >= 0 {
```

crates/bws/src/main.rs
```diff
@@ -516,12 +548,38 @@ async fn process_commands() -> Result<()> {
...
-            println!("Secret deleted correctly");
+            let mut secrets_success = 0;
+            let mut secrets_failed: Vec<(Uuid, String)> = Vec::new();
+
+            for secret in result.data.iter() {
+                match &secret.error {
+                    Some(error_message) => {
+                        secrets_failed.push((secret.id, error_message.to_owned()))
+                    }
+                    None => secrets_success += 1,
```
Hinton, line 564: Do we need to count this? We could just get `input - failed`. Which probably lets you rewrite this section into a filter & map.

crates/bws/src/main.rs
coltonhurst: In this conditional I am purposefully excluding 1 for the grammar of the output. If `secrets_success == 1`, we print this message: `"{} secret deleted successfully."` (instead of `secrets` with the `s`)

crates/bws/src/main.rs
coltonhurst: I intentionally removed this line so that `--project-id` would no longer be passed via a command line flag but used as a parameter, like we do in other commands (ex: `secret edit`), especially as it is required. Is there a way to retain the help text?

Help menu output before and after this change below.

Before:
```
Usage: bws secret create [OPTIONS] --project-id <PROJECT_ID> <KEY> <VALUE>

Arguments:
  <KEY>    
  <VALUE>  

Options:
      --project-id <PROJECT_ID>
          The ID of the project this secret will be added to
      --note <NOTE>
          An optional note to add to the secret
...
```

After:
```
Usage: bws secret create [OPTIONS] <KEY> <VALUE> <PROJECT_ID>

Arguments:
  <KEY>         
  <VALUE>       
  <PROJECT_ID>  

Options:
      --note <NOTE>
          An optional note to add to the secret
...
```

dani-garcia: I think just removing the `long,` would make it into a positional argument, but I haven't tested it

coltonhurst: Ah, that seems to work! Thanks, fixed in [032f4e0](https://github.com/bitwarden/sdk/pull/77/commits/032f4e074656b5cce92ef562ecb77516e7c41701). This is the output now:

```
Usage: bws secret create [OPTIONS] <KEY> <VALUE> <PROJECT_ID>

Arguments:
  <KEY>         
  <VALUE>       
  <PROJECT_ID>  The ID of the project this secret will be added to

Options:
      --note <NOTE>
          An optional note to add to the secret
...
```

crates/bws/src/main.rs
Hinton: Is there a reason to print "0 secrets deleted successfully?"

coltonhurst: I wanted to make it clear to the user than none of the secrets they tried to delete were deleted. (Instead of just printing X secrets were deleted.) No other reason.

crates/bws/src/main.rs
coltonhurst: Good point. Switched to using `filter_map` in [b8074ff](https://github.com/bitwarden/sdk/pull/77/commits/b8074ff181501627822ef320e449c7cb419b29e4)

crates/bws/src/main.rs
```diff
@@ -388,26 +388,23 @@ async fn process_commands() -> Result<()> {
         | Commands::Delete {
             cmd: DeleteCommand::Project { project_ids },
         } => {
+            let mut projects_success = project_ids.len();
```
Hinton, line 391: We should ideally try and avoid mutable variables. In this case we can just assign the variable on line 403 directly?

coltonhurst: Because `Vec` doesn't implement the Copy trait, the value is moved on line 395 when `.delete(ProjectsDeleteRequest { ids: project_ids })` is called, so I need to capture the value before it's moved. I see three options, please let me know what you think!
- We could change the `delete` params to accept a reference to `ProjectsDeleteRequest`, but I figure that would be outside the scope of this PR.
- I can just use variable shadowing and re-declare `projects_success`
- Something else / leave as is

coltonhurst: So you want something like the following:

`let count = project_ids.len();`

Then later on line 562: `let deleted_projects = count - projects_failed.len();`

Hinton: Yes. That gets rid of the mutable variables which are bad.

coltonhurst: Refactored in [daf7921](https://github.com/bitwarden/sdk/pull/77/commits/daf7921b6c4bba6d97328e0bd113c7fa077319af)

dani-garcia (dismissed):

Hinton (dismissed):

dani-garcia (approved):

PR #78, opened 2023-06-27
[DEVOPS-1433] Change renovate config to follow template

michalchecinski:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [x] Other
```

## Objective

Change renovate config to follow template: https://github.com/bitwarden/template/blob/main/.github/renovate.json

## Code changes

- **.github/renovate.json:** 

## Before you submit

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,13 +1,23 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "extends": ["config:base", "schedule:monthly", ":maintainLockFilesWeekly"],
+  "extends": [
+    "config:base",
+    ":combinePatchMinorReleases",
+    ":dependencyDashboard",
+    ":maintainLockFilesWeekly",
+    ":pinAllExceptPeerDependencies",
+    ":prConcurrentLimit10",
+    ":rebaseStalePrs",
+    "schedule:weekends",
+    ":separateMajorReleases"
+  ],
   "separateMajorMinor": true,
+  "enabledManagers": ["cargo", "github-actions", "npm", "nuget"],
   "packageRules": [
     {
+      "groupName": "npm minor",
       "matchManagers": ["npm"],
-      "matchPackagePatterns": ["*"],
-      "matchUpdateTypes": ["minor", "patch"],
-      "groupName": "node non-major"
+      "matchUpdateTypes": ["minor", "patch"]
     },
     {
       "matchManagers": ["cargo"],
@@ -22,15 +32,14 @@
       "groupName": "pyo3 non-major"
     },
     {
+      "groupName": "nuget minor",
       "matchManagers": ["nuget"],
-      "matchPackagePatterns": ["*"],
-      "matchUpdateTypes": ["minor", "patch"],
-      "groupName": "dotnet non-major"
-    },
-    {
-      "matchManagers": ["github-actions"],
-      "matchPackagePatterns": ["*"],
-      "groupName": "gh actions all"
+      "matchUpdateTypes": ["minor", "patch"]
     }
+    {
+      "groupName": "gh minor",
+      "matchManagers": ["github-actions"],
+      "matchUpdateTypes": ["minor", "patch"]
+    },
   ]
 }
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -5,13 +5,11 @@
     ":combinePatchMinorReleases",
     ":dependencyDashboard",
     ":maintainLockFilesWeekly",
-    ":pinAllExceptPeerDependencies",
     ":prConcurrentLimit10",
     ":rebaseStalePrs",
     "schedule:weekends",
     ":separateMajorReleases"
   ],
-  "separateMajorMinor": true,
   "enabledManagers": ["cargo", "github-actions", "npm", "nuget"],
   "packageRules": [
     {


Hinton:
The SDK should not pin it's dependencies since we expect rust crates to consume it. We should strive to use as loose version requirements as possible of the core crates.

.github/renovate.json
```diff
@@ -1,13 +1,23 @@
...
+    ":combinePatchMinorReleases",
+    ":dependencyDashboard",
+    ":maintainLockFilesWeekly",
+    ":pinAllExceptPeerDependencies",
+    ":prConcurrentLimit10",
+    ":rebaseStalePrs",
+    "schedule:weekends",
+    ":separateMajorReleases"
+  ],
   "separateMajorMinor": true,
```
withinfocus, line 14: ⚠️ This is covered by the above (`:separateMajorReleases`) and can be removed.

.github/renovate.json
```diff
@@ -1,13 +1,23 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "extends": ["config:base", "schedule:monthly", ":maintainLockFilesWeekly"],
+  "extends": [
+    "config:base",
+    ":combinePatchMinorReleases",
+    ":dependencyDashboard",
+    ":maintainLockFilesWeekly",
+    ":pinAllExceptPeerDependencies",
```
withinfocus, line 8: ⚠️ Like @Hinton requested.

.github/renovate.json
```diff
@@ -22,15 +32,14 @@
       "groupName": "pyo3 non-major"
```
withinfocus, line 32: @Hinton is this patching itself into some existing group? Otherwise I think we'd replace with the template.

withinfocus (changes requested):

.github/renovate.json
michalchecinski: Done

.github/renovate.json
```diff
@@ -1,13 +1,21 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "extends": ["config:base", "schedule:monthly", ":maintainLockFilesWeekly"],
-  "separateMajorMinor": true,
+  "extends": [
+    "config:base",
+    ":combinePatchMinorReleases",
```
Hinton, line 5: Any 0. release considers minors to be breaking changes. And since we don't really want to bump patch we should omit those.

.github/renovate.json
```diff
@@ -1,13 +1,21 @@
...
+    ":dependencyDashboard",
+    ":maintainLockFilesWeekly",
+    ":prConcurrentLimit10",
```
Hinton, line 8: Change the limit to 3 since we still don't have mac runners, and this would consume the whole orgs runners.

Hinton (changes requested):

.github/renovate.json
withinfocus: Should we be using <v1 releases in a production-quality product? How long would that be around for to be worth the customization. Nonetheless this shortcut means `"separateMinorPatch": false` and the opposite could be placed in the cargo-specific package rules lower down.

.github/renovate.json
withinfocus: My take on us using this as the default is that it will be the norm for smaller repos, and I consider the SDK one of those. The large repos e.g. `clients` or `server` need more restriction. Remember that this will only run on the weekends too; we can't stop rebase events creating more runs at any time however.

.github/renovate.json
```diff
@@ -1,13 +1,21 @@
...
+    ":rebaseStalePrs",
+    "schedule:weekends",
+    ":separateMajorReleases"
```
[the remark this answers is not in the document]
withinfocus: Per the other comments, or your desire for this repo specifically?

.github/renovate.json
Hinton: We can look into it after we have enterprise, until then we can't have 10 renovate PRs since that would consume all mac runners indefinitely.

.github/renovate.json
withinfocus: You'd need to specify the package names and update types to group on here though, which aren't present. Is this working as expected?

.github/renovate.json
Hinton: That's how the rust ecosystem works. 0. packages are essentially reasonable stable, they just don't promise api stability. Avoiding all 0. packages means we have to write all the crypto ourselves.

.github/renovate.json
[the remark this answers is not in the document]
Hinton: This repo.

.github/renovate.json
withinfocus: There's over a dozen configs with this already deployed ... any idea how many dependencies need updating here? As I've deployed this around the organization I usually only see it creating a handful and being done.

.github/renovate.json
MGibson1: Minor releases in rust are breaking changes, we need to bundle only patch

.github/renovate.json
Hinton: Kind of, we just have to hold on updating until they all match. `"matchPackagePatterns": ["pyo3*"],`

.github/renovate.json
Hinton: No idea. We should strive towards staying on lower versions because we want to encourage adoption of the sdk. Which means we should have broad versions. I kind of don't really care about staying up to date assuming the lockfile refreshes us to the latest version.

michalchecinski:
Hey @Hinton @withinfocus 👋 What are the next steps for this PR? Should we keep the renovate config as it is currently on the master? Since there are specific setup for this repo needed, maybe we can do an exception and not follow the template?

withinfocus:
@Hinton covered this actually with #90 and I forgot this was still here. Can close this and consider it done.

michalchecinski:
Thanks @withinfocus ! Closing now

PR #79, opened 2023-06-27
Move all tests to test workflow

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

- Update the CI workflow to move all the tests into the test workflow.
- Change the test workflow to remove parallel package execution.
	- We have to re-build each dependency for each package which is slow
	- Build > test time currently.

## Before you submit


MGibson1 (approved):
< 1/2 the time? no brainer :shipit:

PR #84, opened 2023-06-30
Argon2 support

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
Support for Argon2id key derivation. 

## Code changes
- **auth_settings.rs:** Replaced the current KdfType and kdf_iterations for a nicer enum, extracted the KDF functions and moved them to `crypto.rs`, to remove duplication.
- **crypto.rs:**: Create new `hash_kdf` function with support for argon2, used both in this file and `auth_settings.rs`
- **utils.rs:**: Added default values

### DIFF
--- a/crates/bitwarden/src/client/auth_settings.rs
+++ b/crates/bitwarden/src/client/auth_settings.rs
@@ -5,46 +5,65 @@ use bitwarden_api_identity::models::{KdfType, PreloginResponseModel};
 
 use crate::{
     crypto::{PbkdfSha256Hmac, PBKDF_SHA256_HMAC_OUT_SIZE},
-    util::{default_kdf_iterations, BASE64_ENGINE},
+    error::Result,
+    util::{
+        default_argon2_iterations, default_argon2_memory, default_argon2_parallelism,
+        default_pbkdf2_iterations, BASE64_ENGINE,
+    },
 };
 
 #[derive(Debug)]
 pub(crate) struct AuthSettings {
     pub email: String,
-    kdf_type: KdfType,
-    pub(crate) kdf_iterations: NonZeroU32,
+    pub(crate) kdf: Kdf,
+}
+
+#[derive(Debug)]
+pub enum Kdf {
+    PBKDF2 {
+        iterations: NonZeroU32,
+    },
+    Argon2id {
+        iterations: NonZeroU32,
+        memory: NonZeroU32,
+        parallelism: NonZeroU32,
+    },
 }
 
 impl AuthSettings {
     pub fn new(response: PreloginResponseModel, email: String) -> Self {
-        Self {
-            email,
-            kdf_type: response.kdf.unwrap_or_default(),
-            kdf_iterations: response
-                .kdf_iterations
-                .and_then(|e| NonZeroU32::new(e as u32))
-                .unwrap_or_else(default_kdf_iterations),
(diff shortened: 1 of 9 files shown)


Hinton (commented):
Can we add some unit tests for this? Would like to trigger all the `expect` and new methods.

crates/bitwarden/src/crypto.rs
```diff
@@ -449,6 +493,32 @@ mod tests {
                 166, 209, 247, 194, 87, 232, 26, 48, 85, 193, 249, 179, 155
             ]
         );
+
+        let (key, mac) = stretch_key_password(
```
Hinton, line 497: I think we should add a new test scenario for these? Should assist with tracking down the failing test.

crates/bitwarden/src/client/auth_settings.rs
```diff
@@ -75,12 +95,33 @@ mod tests {
...
+                .unwrap(),
             "ZF6HjxUTSyBHsC+HXSOhZoXN+UuMnygV5YkWXCY4VmM="
         );
         assert_eq!(
-            settings.make_user_password_hash("asdfasdf"),
+            settings.make_user_password_hash("asdfasdf").unwrap(),
             "wmyadRMyBZOH7P/a/ucTCbSghKgdzDpPqUnu/DAVtSw="
         );
+
+        let res = PreloginResponseModel {
```
Hinton, line 108: Move to a separate test?

dani-garcia:
I've added some tests that include argon2. 

The `expect` would never be triggered because they are all about invalid input array sizes, and we're using fixed size arrays.

For example in `stretch_key_password`: 
https://github.com/bitwarden/sdk/pull/84/files#diff-be83f34759339801d2c3f51076b7cf51899ecc7bb310cb72761433527b1abdf3R286-R297

`Hkdf::from_prk` expects `master_key` to be at least 32 bytes, but that's enforced by  the `hash_kdf` call above.
`hkdf::expand` expects `key` and `mac_key` to not exceed a maximum size (32 * 255), but they are defined as 32 byte fixed size arrays.

crates/bitwarden/src/crypto.rs
dani-garcia: During the test you get the stack trace indicating the failing line, which should help already, so I consider the separation more of a style choice, but if you prefer I can move it to a separate test function.

Hinton: You do, but you have to look at the source code to understand what failed, separate methods lets you immediately know the method name which to me adds some value.

crates/bitwarden/src/client/encryption_settings.rs
```diff
@@ -104,9 +104,8 @@ impl EncryptionSettings {
         let (key, mac_key) = crate::crypto::stretch_key_password(
             password.as_bytes(),
             auth.email.as_bytes(),
-            auth.kdf_iterations,
-        )
-        .map_err(|_| CryptoError::KeyStretch)?;
```
Hinton, line 109: Is this error unused now?

dani-garcia: Yeah, seems like we have no other uses of that variant anymore, I'll remove it.

PR #91, opened 2023-07-11
Update workflow credentials and key vault

mimartin12:
## Type of change

```
- [X] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
There were security changes around how Azure Key Vaults were accessed with GH secrets. This updates the napi release workflow to use the newest credentials and Azure Key Vault.

## Code changes

- **.github/workflows/release-napi.yml:** Update the napi release workflow with the newest credentials.

## Before you submit


dani-garcia:
Some of the other workflows seem to be using the same old credentials and key vault as well, should those be updated too?
- publish-rust-crates.yml
- release-cli.yml
- version-bump.yml

dani-garcia (approved):

PR #92, opened 2023-07-12
SM-837: Add Secret and Project Length Restrictions

coltonhurst:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Add Secret and Project length restrictions for each field:
- `project_name` <= 500
- `secret_key` <= 500
- `secret_value` <= 3000
- `secret_note` <= 7000

## Code changes

- **error.rs:** Add the FieldLengthExceeded error
- **project/create.rs:** Add and call the validate function
- **project/mod.rs:** Add the project_name max const
- **project/update.rs:** Add and call the validate function
- **secret/create.rs:** Add and call the validate function
- **secret/mod.rs:** Add the secret key, value, and note max const's
- **secret/update.rs:** Add and call the validate function

## Before you submit


dani-garcia:
Ideally these checks should be moved to the `bitwarden` crate, inside the `create_secret`, `update_secret` and `delete_secret`functions, that way they also apply to anyone using the SDK and not just to the CLI.

dani-garcia (approved):

dani-garcia (dismissed):

Hinton (dismissed):
Most of the time the validation of create and update are identical. Do we want to tie those implementations together somehow?

coltonhurst:
> Most of the time the validation of create and update are identical. Do we want to tie those implementations together somehow?

I think that's a good idea. Due to the SM timeline, I can create a separate ticket and we can address the refactor timeline there.

Hinton (approved):

dani-garcia:
@coltonhurst This looks like it's covered by the now merged https://github.com/bitwarden/sdk/pull/810, so we can probably close this right?

coltonhurst:
> @coltonhurst This looks like it's covered by the now merged #810, so we can probably close this right?

yep! Closing it, thank you :)

PR #99, opened 2023-07-12
Set rangeStrategy to widen

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [x] Other
```

## Objective

Set `rangeStrategy` to `widen` for cargo.

## Before you submit

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -20,15 +20,17 @@
     },
     {
       "matchManagers": ["cargo"],
+      "rangeStrategy": "widen"
+    },
+    {
+      "matchManagers": ["cargo"],
       "matchUpdateTypes": ["patch"],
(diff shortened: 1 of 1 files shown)


.github/renovate.json
```diff
@@ -18,17 +18,19 @@
       "matchManagers": ["npm"],
       "matchUpdateTypes": ["minor", "patch"]
     },
+    {
+      "matchManagers": ["cargo"],
+      "rangeStrategy": "widen"
+    },
```
MGibson1, line 24: this would widen through breaking versions too, right?

I think that's what we want, but it does mean we likely have more manual updates to these PRs

MGibson1 (approved):

PR #100, opened 2023-07-12
Change versions to ranges for libraries

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [x] Other
```

## Objective

Change versions to ranges which should let renovate handle them. Replaces the fixed versions with ranges that match the behavior of  `^`.

## Before you submit


dani-garcia:
This is equivalent to what we have already, Cargo handles `1.2.3`, `^1.2.3` and `>=1.2.3, <2.0.0` the same, is this just a change needed for renovate to work correctly? Would using `^1.2.3` ranges work as well? I think it would definitely look nicer at least.

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio

Hinton:
@dani-garcia unfortunately that doesn't work. Example: https://github.com/bitwarden/sdk/pull/94/files

dani-garcia (approved):

PR #103, opened 2023-07-14
[PM-3029] Add support for shell completion

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
Add support for shell completion using the official crates from the Clap project. 

I've made a command for each to be generated at runtime, but we could potentially also generate them at build time, though that will require to move the `Cli` and `Commands` structs to a separate file. Build time generation might make more sense if we ever want to distribute `.deb` or `.rpm` packages.

## Code changes
- Changed `#[command(name)]` to `bws` to represent the binary name, used by the shell completions.
- Added a command:
  - `Completions`: Receives a single parameter for the type of shell to generate the completions for (Bash, zsh, pwsh, etc),  if not provided it will try autodetecting the current shell.

Note that both these changes seem to ignore the `hide` parameter set in the deprecated commands, so we might want to consider holding off merging this until they are removed.

## How to enable completions
This varies by shell, in `zsh`:
Edit the file `~/.zshrc` and append to the end the following line:
```
source <(/path/to/bws completions zsh)
```

## Screenshots
https://github.com/bitwarden/sdk/assets/725423/9be931e6-9093-447d-a365-13b838b9968c


Hinton:
Do we know what the convention is for shell completion and manpages for rust packages? Feels somewhat odd to have the CLI itself generate them.

dani-garcia:
I don't think there are specific conventions for rust packages, usually man pages are included in the system install packages (so .deb or .rpm in Linux), and it's the system package manager that moves them to the correct place. We could generate them at build time and include them in the release zip, but without an automatic installation step that might end up confusing users.

Shell autocompletion on the other hand I've noticed a lot of software handle the generation from the CLI, I assume because they can't control which shell the user is running, and the generated autocomplete code is different for all of them, for example:

- [Trivy](https://aquasecurity.github.io/trivy/v0.40/tutorials/shell/shell-completion/) and [K0s](https://docs.k0sproject.io/v1.23.6+k0s.2/shell-completion/) have a completion subcommand
- [Terraform](https://developer.hashicorp.com/terraform/cli/commands#shell-tab-completion) has an --install-autocomplete flag

Hinton:
Please split this into two PRs. We'll try and get shell completion merged while thinking of what we should do about man pages.

Hinton (commented):
Can you add this to the readme?

crates/bws/README.md
```diff
@@ -18,5 +18,29 @@ Or download a pre-built binary from the [Releases](https://github.com/bitwarden/
...
+echo 'source <(/path/to/bws completions zsh)' >> ~/.zshrc
+```
+
+### Bash
+
+Enable autocompletions for the current user:
+
+```zsh
+echo 'source <(/path/to/bws completions bash)' >> ~/.bashrc
+```
```
Hinton, line 43: Should this be in the readme or the help docs? The readme is only really visible on crates.io

crates/bws/src/main.rs
```diff
@@ -26,16 +27,16 @@ use render::{serialize_response, Color, Output};
...
 #[derive(Parser, Debug)]
-#[command(name = "Bitwarden Secrets CLI", version, about = "Bitwarden Secrets CLI", long_about = None)]
+#[command(name = "bws", version, about = "Bitwarden Secrets CLI", long_about = None)]
 struct Cli {
     // Optional as a workaround for https://github.com/clap-rs/clap/issues/3572
     #[command(subcommand)]
     command: Option<Commands>,
 
-    #[arg(short = 'o', long, global = true, value_enum, default_value_t = Output::JSON)]
+    #[arg(short = 'o', long, global = true, value_enum, default_value_t = Output::JSON, help="Select the output format for the commands", hide = true)]
```
Hinton, line 36:
```suggestion
    #[arg(short = 'o', long, global = true, value_enum, default_value_t = Output::JSON, help="Output format", hide = true)]
```

crates/bws/src/main.rs
```diff
@@ -26,16 +27,16 @@ use render::{serialize_response, Color, Output};
...
     output: Output,
 
-    #[arg(short = 'c', long, global = true, value_enum, default_value_t = Color::Auto)]
+    #[arg(short = 'c', long, global = true, value_enum, default_value_t = Color::Auto, help="Enable or disable the use of colors in the output")]
```
Hinton, line 39:
```suggestion
    #[arg(short = 'c', long, global = true, value_enum, default_value_t = Color::Auto, help="Use colors in the output")]
```

crates/bws/src/main.rs
```diff
@@ -245,6 +250,18 @@ async fn process_commands() -> Result<()> {
...
+        let Some(shell) = shell.or_else(Shell::from_env) else {
+            eprintln!("Couldn't autodetect a valid shell. Run `bws completions --help` for more info.");
+            std::process::exit(1);
+        };
+
+        let mut cmd = Cli::command();
+        let name = cmd.get_name().to_string();
+        clap_complete::generate(shell, &mut cmd, name, &mut std::io::stdout());
+        return Ok(());
+    }
```
Hinton, line 263: We're starting to have a few if let here, should we try and use a switch?

crates/bws/README.md
dani-garcia: The help docs should probably only be updated after we make a release with these changes merged in I think, otherwise it might be confusing for users of the existing release.

crates/bws/src/main.rs
dani-garcia: That's a good point, we should probably consider extracting them to a separate function in the future because the process_commands is already getting pretty big, but that's a bigger refactoring.

Hinton (approved):

PR #104, opened 2023-07-14
Improve errors in bitwarden_json

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
`bitwarden_json` handles the errors from `Client` calls correctly and returns a JSON like `{success: false, errorMessage: "whatever"}`.

This is not the case if the command sent to the `run_command` is malformed, in this case it would just return the error as a String. I've changed this so the error returned is in the same JSON format as the `Client` calls.

`parse_settings` would also silently start using the defaults if the provided settings where malformed, I've printed an error message in that case, but it might be best to change the return type of `Client::new`, though this means updating all the language bindings as well.

### DIFF
--- a/crates/bitwarden-json/src/client.rs
+++ b/crates/bitwarden-json/src/client.rs
@@ -2,7 +2,7 @@ use bitwarden::client::client_settings::ClientSettings;
 
 use crate::{
     command::{Command, ProjectsCommand, SecretsCommand},
-    response::ResponseIntoString,
+    response::{Response, ResponseIntoString},
 };
 
 pub struct Client(bitwarden::Client);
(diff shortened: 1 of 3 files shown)


crates/bitwarden-json/src/client.rs
```diff
@@ -71,8 +75,11 @@ impl Client {
 
     fn parse_settings(settings_input: Option<String>) -> Option<ClientSettings> {
         if let Some(input) = settings_input.as_ref() {
-            if let Ok(settings) = serde_json::from_str(input) {
-                return Some(settings);
+            match serde_json::from_str(input) {
+                Ok(settings) => return Some(settings),
+                Err(e) => {
+                    eprintln!("Failed to parse settings: {}", e);
```
Hinton, line 81: Do we know what happens with eprint in wasm or other environments?

dani-garcia: It should work fine in everything but wasm, which doesn't have a standard output by default. We can add the log crate to bitwarden-json and use that, which should work.

dani-garcia: Changed it to log::error, which works when used with `console_log` in the wasm crate.

Hinton (approved):

PR #109, opened 2023-07-17
[PM-2076] Properly handle error responses in send_identity_connect_request

### LINKED ISSUE #87: Failed to parse IdentityTokenResponse
### Steps To Reproduce

We keep getting `Internal error: Failed to parse IdentityTokenResponse` from the vault SDK ~17 hours ago. Seems to be an issue with deserializing the response from the Bitwarden API.

### Expected Result

Attempt to retrieve secrets from the vault shouldn't fail.

### Actual Result

An [internal error](https://github.com/bitwarden/sdk/blob/8a689bb18db4a0937654771b544820e74fc27866/crates/bitwarden/src/error.rs#L48) is returned instead.

### Screenshots or Videos

_No response_

### Additional Context

_No response_

### Operating System

Linux

### Operating System Version

_No response_

### Build Version

0.2.1

### Issue Tracking Info

Hinton:
## Type of change

```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Ensure any server error is handled properly in the SDK instead of silently ignored by converting it to an error using `reqwest` `error_for_status` fn.

Resolves #87 

A network error now resolves into 

```
Error:
   0: HTTP status server error (504 Gateway Timeout) for url (https://localhost:8080/identity/connect/token)
```

## Before you submit

### DIFF
--- a/crates/bitwarden/src/auth/api/request/mod.rs
+++ b/crates/bitwarden/src/auth/api/request/mod.rs
@@ -43,13 +43,13 @@ async fn send_identity_connect_request(
         request = request.header("Auth-Email", BASE64_ENGINE.encode(email.as_bytes()));
     }
 
-    let raw_response = request
+    let response = request
         .body(serde_qs::to_string(&body).unwrap())
         .send()
-        .await?
-        .error_for_status()?
-        .text()
         .await?;
 
-    parse_identity_response(&raw_response)
+    let status = response.status();
+    let text = response.text().await?;
+
+    parse_identity_response(status, &text)
 }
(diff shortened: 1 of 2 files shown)


dani-garcia (approved):
There might be some cases where we'd get better error messages by parsing the response body, for example when receiving a generic error like `400 Bad Request`. That said this should better than what we have now

dani-garcia (approved):
LGTM, eventually we might want to parse the output to differentiate specific errors, but this is much better already than "Failed to parse Response"

PR #113, opened 2023-07-20
[PM-3125] Initial setup on smaller mobile-only sdk

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
Create a smaller client surface for the new mobile apps, by hiding all the secrets manager stuff under a `secrets` feature flag, and making a new encryption-focused API hidden under a `mobile` feature flag.

As an initial pass, I've implemented the KDF functions needed for login and the basics needed to initialise the cryptography parts of the client.

Sidenote: We really need to split the bitwarden_json::Command struct into multiple files 😨

### DIFF
--- a/crates/bitwarden-json/src/command.rs
+++ b/crates/bitwarden-json/src/command.rs
@@ -1,14 +1,15 @@
-use bitwarden::auth::request::AccessTokenLoginRequest;
-
 #[cfg(feature = "secrets")]
-use bitwarden::secrets_manager::{
-    projects::{
-        ProjectCreateRequest, ProjectGetRequest, ProjectPutRequest, ProjectsDeleteRequest,
-        ProjectsListRequest,
-    },
-    secrets::{
-        SecretCreateRequest, SecretGetRequest, SecretIdentifiersRequest, SecretPutRequest,
-        SecretsDeleteRequest,
+use bitwarden::{
+    auth::request::AccessTokenLoginRequest,
+    secrets_manager::{
+        projects::{
+            ProjectCreateRequest, ProjectGetRequest, ProjectPutRequest, ProjectsDeleteRequest,
+            ProjectsListRequest,
+        },
+        secrets::{
+            SecretCreateRequest, SecretGetRequest, SecretIdentifiersRequest, SecretPutRequest,
+            SecretsDeleteRequest,
+        },
     },
 };
 
@@ -20,7 +21,7 @@ use bitwarden::{
 
 #[cfg(feature = "mobile")]
 use bitwarden::mobile::{
-    crypto::{InitOrgCryptoRequest, InitUserCryptoRequest},
+    crypto::InitCryptoRequest,
     kdf::{KdfParamRequest, PasswordHashRequest},
 };
 
@@ -52,6 +53,7 @@ pub enum Command {
     ///
     ApiKeyLogin(ApiKeyLoginRequest),
 
+    #[cfg(feature = "secrets")]
     /// Login with Secrets Manager Access Token
     ///
     /// This command is for initiating an authentication handshake with Bitwarden.
@@ -202,7 +204,7 @@ pub enum MobileKdfCommand {
     ///
     /// Returns: String
     ///
-    GetUserPasswordHash(PasswordHashRequest),
+    HashPassword(PasswordHashRequest),
 }
 
 #[cfg(feature = "mobile")]
@@ -210,11 +212,7 @@ pub enum MobileKdfCommand {
 #[serde(rename_all = "camelCase", deny_unknown_fields)]
 pub enum MobileCryptoCommand {
     /// > Requires having previously set the KDF parameters
-    /// Decrypts the users keys and initializes the user crypto, allowing for the encryption/decryption of the users items
+    /// Decrypts the users keys and initializes the user crypto, allowing for the encryption/decryption of the users vault
     ///
-    InitUserCrypto(InitUserCryptoRequest),
-    /// > Requires having previously initialized the user crypto
-    /// Decrypts the users organization keys, allowing for the encryption/decryption of the items in the users organizations
-    ///
-    InitOrgCrypto(InitOrgCryptoRequest),
+    InitCrypto(InitCryptoRequest),
 }
--- a/crates/bitwarden-json/src/command.rs
+++ b/crates/bitwarden-json/src/command.rs
@@ -20,7 +20,10 @@ use bitwarden::{
 };
 
 #[cfg(feature = "mobile")]
-use bitwarden::mobile::{crypto::InitCryptoRequest, kdf::PasswordHashRequest};
+use bitwarden::mobile::kdf::PasswordHashRequest;
+
+#[cfg(all(feature = "mobile", feature = "internal"))]
+use bitwarden::mobile::crypto::InitCryptoRequest;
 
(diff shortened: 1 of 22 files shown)


package.json
```diff
@@ -16,7 +16,7 @@
   "scripts": {
     "lint": "prettier --check .",
     "prettier": "prettier --write .",
-    "schemas": "rimraf ./support/schemas && cargo run --bin sdk-schemas --features internal && ts-node ./support/scripts/schemas.ts",
+    "schemas": "rimraf ./support/schemas && cargo run --bin sdk-schemas --features internal,secrets,mobile && ts-node ./support/scripts/schemas.ts",
```
Hinton, line 19: Should we use `--all-features` here?

crates/bws/Cargo.toml
```diff
@@ -18,7 +18,7 @@ keywords = ["bitwarden", "secrets-manager", "cli"]
 clap = { version = "4.3.0", features = ["derive", "env"] }
 tokio = { version = "1.28.2", features = ["rt-multi-thread", "macros"] }
 log = "0.4.18"
-bitwarden = { path = "../bitwarden", version = "0.2.1" }
+bitwarden = { path = "../bitwarden", version = "0.2.1", features = ["secrets"] }
```
Hinton, line 21: Napi also needs secrets right?

crates/bitwarden/src/mobile/client_kdf.rs
```diff
@@ -0,0 +1,23 @@
...
+}
+
+impl<'a> ClientKdf<'a> {
+    pub async fn set_kdf_params(&mut self, req: KdfParamRequest) -> Result<()> {
+        set_kdf_params(self.client, req).await
+    }
+
+    pub async fn get_user_password_hash(&mut self, req: PasswordHashRequest) -> Result<String> {
+        get_user_password_hash(self.client, req).await
+    }
```
Hinton, line 16: Do we need to persist the kdf params or could they be an input to `get_user_password_hash`?

And should we rename it to `hash_password`.

crates/bitwarden-json/src/command.rs
```diff
@@ -1,21 +1,29 @@
...
-    secrets_manager::{
-        projects::{
-            ProjectCreateRequest, ProjectGetRequest, ProjectPutRequest, ProjectsDeleteRequest,
-            ProjectsListRequest,
-        },
-        secrets::{
-            SecretCreateRequest, SecretGetRequest, SecretIdentifiersRequest, SecretPutRequest,
-            SecretsDeleteRequest,
-        },
+use bitwarden::auth::request::AccessTokenLoginRequest;
```
Hinton, line 1: Access token is a secrets manager concept.

crates/bitwarden-json/src/command.rs
```diff
@@ -166,3 +181,40 @@ pub enum ProjectsCommand {
...
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub enum MobileCryptoCommand {
+    /// > Requires having previously set the KDF parameters
+    /// Decrypts the users keys and initializes the user crypto, allowing for the encryption/decryption of the users items
+    ///
+    InitUserCrypto(InitUserCryptoRequest),
+    /// > Requires having previously initialized the user crypto
+    /// Decrypts the users organization keys, allowing for the encryption/decryption of the items in the users organizations
+    ///
+    InitOrgCrypto(InitOrgCryptoRequest),
```
Hinton, line 219: Should this be a single method? I don't think there are any valid scenarios where we only have one of these keys.

crates/bitwarden/Cargo.toml
```diff
@@ -13,7 +13,11 @@ edition = "2021"
 rust-version = "1.57"
 
 [features]
+default = ["secrets"]
```
Hinton, line 16: Is this for backwards compatibility? We should add a deprecation note and update our docs to suggest turning this on.

Hinton (commented):
Looks like a great start, some thoughts if we can make the hash stateless, and combine the init method.

crates/bitwarden/src/mobile/client_kdf.rs
dani-garcia: We use the KDF params both to create the master password hash for login, and to decrypt the symmetric encryption key in `InitUserCrypto`. 

The reason for the separation in two steps is for a vault unlock scenario, where you already have the vault and encryption keys locally. In those cases, you'd only need to call `set_kdf_params` and then initialize the crypto. But we could just require the KDF parameters as input for the crypto initialization as well.

Agreed on the rename.

crates/bitwarden-json/src/command.rs
dani-garcia: I made those to be one to one with how we handle things internally, but you are right that from a consumer pov the separation is not needed.

crates/bitwarden-json/src/command.rs
dani-garcia: It's hidden under the feature flag now.

crates/bitwarden/Cargo.toml
dani-garcia: Yeah, it's for backwards compatibility and should be removed eventually, I think. I've updated the bitwarden crate README and the docs to include the feature, we should add a deprecation entry to the changelog once that is merged as well.

crates/bitwarden/src/mobile/client_kdf.rs
Hinton: Hmm that's a valid point.

It does mean on say account switching they would need to call both `set_kdf_params` and `InitCrypto`, which might be a bit confusing.

crates/bitwarden/Cargo.toml
dani-garcia: Added the notice to the changelog

crates/bitwarden/src/mobile/client_kdf.rs
dani-garcia: I've removed the set_kdf_params now, both options seem to have some pros and cons, but at least this way we reduce the API surface a bit

Hinton (commented):
Looks good, can we hide the `InitCrypto` command behind the internal flag? It's not yet a stable interface.

dani-garcia:
Do we want to make it so both `mobile` and  `internal` need to be enabled for it to be accessible?

Hinton:
Yes.

crates/bitwarden/src/client/auth_settings.rs
```diff
@@ -31,7 +32,13 @@ pub enum Kdf {
 }
 
 impl AuthSettings {
+    #[cfg(feature = "internal")]
     pub fn new(response: PreloginResponseModel, email: String) -> Self {
+        use crate::util::{
+            default_argon2_iterations, default_argon2_memory, default_argon2_parallelism,
+            default_pbkdf2_iterations,
+        };
```
Hinton, line 40: What is the benefit of moving use into functions? Just take advantage of the existing `feature` check?

Hinton (approved):
Looks good!

crates/bitwarden/src/client/auth_settings.rs
dani-garcia: Yeah, pretty much. Otherwise it's another cfg that we need to add.

PR #114, opened 2023-07-20
Add changelogs

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [x] Other
```

## Objective

We should record a changelog for consumers of the crates that doesn't require a user to walk through gh releases.

## Before you submit

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,6 @@
+# Changelog
+
+The changelog for the crates are located in the individual crates.
+
+- [`bitwarden`](./crates/bitwarden/CHANGELOG.md)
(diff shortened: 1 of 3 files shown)


dani-garcia (approved):
LGTM. I assume we only want to include any public SDK user facing changes here? So internal refactors, minor dependency updates and anything under the `internal` and `mobile` feature flags should be excluded from here.

Should we add entries here in the same PR's that make the changes, or do we keep the modifications to the changelogs to separate PR's?

Hinton:
@dani-garcia yea only public api changes. Ideally in the same PR otherwise we'll forget it.

PR #115, opened 2023-07-21
UniFFI bindings for Bitwarden

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Code changes

## Before you submit


dani-garcia:
Initial version based on uniffi proc macros tested on Android. Some comments:
- Implemented all the current mobile sdk commands, and they are working from an Android app
- Needed to sprinkle a fair amount of `#[derive(uniffi::Record)]` all over the `bitwarden` crate, for all the input and output structs, and their contents.
- Had to create a `uniffi_support` file to handle conversion for DateTimes, UUIDs and CipherStrings, at the moment this requires implementing an unsafe trait, but we're depending on existing implementations to types in the standard library (like `String`, `u32` or `SystemTime`) and then converting the result, which should be safe.
- Created a new `Error` type in the `bitwarden_uniffi` to wrap `bitwarden::Error` because `uniffi` doesn't support errors defined in external dependencies, if we move this integration to the `bitwarden` crate that won't be an issue.
- Related to the error, the enum can't just be named `Error` because that gets translated to `Exception` in Kotlin and that breaks everything.
- The `uniffi` client is behind an Arc, so we put the `bitwarden` `Client` inside an RwLock to allow mutability
- Related to the last point, I've changed the new mobile API and removed the unnecessary mutability, the only place where we modify the client is when initializing the crypto.

Still a work in progress but results are pretty good so far.

This is the generated API used in a Kotlin app, the DateTime types get converted to a Java Instant, UUIDs and CipherStrings are converted to strings. The API is using `suspend fn` coroutines in place of the async calls, which seems reasonably idiomatic for someone with just a passing knowledge of Kotlin
image

Hinton:
Replaced by #140.

Issue #31, opened 2023-04-17
Make available for Raspberry Pi arm64


0lzi:
### Steps To Reproduce

1. No available binary 

Tried downloading the Apple aarch64 and get `cannot execute binary file: Exec format error`

### Expected Result

Expect a arch64 or armhf binary and be able to use bws commands

### Actual Result

no aarch64 / armhf binary

### Additional Context

Linux raspberrypi 5.15.84-v8+ #1613 SMP PREEMPT aarch64 GNU/Linux

### Operating System

Linux

### Operating System Version

rapperypi OS

### Shell

Bash

### Build Version

na

### Issue Tracking Info

- [X] I understand that work is tracked outside of Github. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

Hinton:
Hi @0lzi,

The darwin aarch64 build is unfortunately not compatible with linux, they use different standard libraries. I did look into enabling cross compiling for the linux build but ran into some issues that will require some more extensive work to resolve.

In the meanwhile it should be possible to compile the CLI yourself using `cargo install bws` if you have the rust toolchain installed.

PR #116, opened 2023-07-21
[PM-1953] Vendor openSSL dependency (static linking)

### LINKED ISSUE #75: Unable to find libssl.so.3
### Steps To Reproduce

1. Download latest linux release on Ubuntu 20.04
1. Execute bws

### Expected Result

No error

### Actual Result

bws: error while loading shared libraries: libssl.so.3: cannot open shared object file: No such file or directory

### Screenshots or Videos

_No response_

### Additional Context

openssl and libssl-dev system packages are installed

### Operating System

Linux

### Operating System Version

Ubuntu 20.04

### Shell

Bash

### Build Version

N/A (latest release build)

### Issue Tracking Info

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

- Cross-compile aarch64-unknown-linux-gnu using cross.
- Statically link openSSL by marking it as a vendored dependency.

Resolves #75, #31.

## Before you submit


.github/workflows/build-cli.yml
```diff
@@ -67,21 +70,27 @@ jobs:
...
-          sudo apt-get install gcc-aarch64-linux-gnu
-
       - name: Cache cargo registry
         uses: Swatinem/rust-cache@dd05243424bd5c0e585e4b55eb2d7615cdd32f1f # v2.5.1
         with:
           key: ${{ matrix.settings.target }}-cargo-${{ matrix.settings.os }}
 
+      - name: Install Cross (aarch64-unknown-linux-gnu)
+        if: ${{ matrix.settings.target == 'aarch64-unknown-linux-gnu' }}
+        run: cargo install cross
```
MGibson1, line 80: Is this cached as part of cargo caching?

crates/bws/Cargo.toml
MGibson1, line 1: can't say I love this formatter. Is it documented so we're consistent at least?

crates/bws/Cargo.toml
```diff
@@ -15,23 +15,31 @@ keywords = ["bitwarden", "secrets-manager", "cli"]
...
 toml = "0.7.4"
-comfy-table = "^7.0.1"
-chrono = { version = "0.4.26", features = ["clock", "std"], default-features = false }
 uuid = { version = "^1.3.3", features = ["serde"] }
 
 [dev-dependencies]
 tempfile = "3.5.0"
+
+[target.'cfg(target_os = "linux")'.dependencies]
+openssl = { version = "0.10", features = ["vendored"] }
```
MGibson1, line 45: Statically linking this probably activates more of the [OpenSSL license ](https://www.openssl.org/source/license-openssl-ssleay.txt) requirements -- namely, that you have to be able to get at the copyright from the binary and any marketing we do needs to credit OpenSSL

crates/bws/Cargo.toml
Hinton: It's the vscode TOML extension.

.github/workflows/build-cli.yml
Hinton: Yes

crates/bws/Cargo.toml
Hinton: I think so, we could look into rusttls which gets rid of the openssl dependency.

Hinton: OpenSSL 3 is also licensed under Apache so we should be fine.

vgrassia (dismissed):

crates/bws/Cargo.toml
MGibson1: yes, I had 1.1.1 in my head

MGibson1 (approved):

PR #120, opened 2023-07-24
[PM-1952] Replace action-rs with dtolnay

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Replace the `action-rs` github actions since they are no longer actively maintained.

- `toolchain` is replaced with `dtolnay/rust-toolchain`
- `cargo` uses the direct cargo commands instead.

## Before you submit

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -271,6 +271,7 @@ dependencies = [
  "bitwarden-api-api",
  "bitwarden-api-identity",
  "cbc",
+ "chrono",
  "getrandom 0.2.10",
  "hkdf",
  "hmac",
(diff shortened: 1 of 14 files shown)


MGibson1 (approved):

vgrassia:
Do we need an action to set up `rust`? I think it is included in the GitHub runner image by default.

Hinton:
> Do we need an action to set up `rust`? I think it is included in the GitHub runner image by default.

It is, but this is a more consist way to configure the rustup to specific versions with specific components. The same arugment could be made about node and jvm, most runners comes with 1 version of it.

vgrassia (approved):

PR #121, opened 2023-07-24
[PM-3131] Implement Folder decryption

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
Implement folder models and folder decryption for the mobile SDK.

I've exposed the CipherString type directly in the schemas to avoid doing a lot of `.parse::<CipherString>()?` all over the place. It's defined inside the `mobile` module to limit the scope, as it's not something we want exposed in the normal SDK.

### DIFF
--- a/crates/bitwarden/Cargo.toml
+++ b/crates/bitwarden/Cargo.toml
@@ -27,7 +27,7 @@ serde = { version = ">=1.0, <2.0", features = ["derive"] }
 serde_json = ">=1.0.96, <2.0"
 serde_qs = ">=0.12.0, <0.13"
 serde_repr = ">=0.1.12, <0.2"
-schemars = { version = ">=0.8, <0.9", features = ["uuid1"] }
+schemars = { version = ">=0.8, <0.9", features = ["uuid1", "chrono"] }
 log = ">=0.4.18, <0.5"
 assert_matches = ">=1.5.0, <2.0"
 thiserror = ">=1.0.40, <2.0"
@@ -52,6 +52,7 @@ getrandom = { version = ">=0.2.9", features = ["js"] }
 
 bitwarden-api-identity = { path = "../bitwarden-api-identity", version = "0.2.0" }
 bitwarden-api-api = { path = "../bitwarden-api-api", version = "0.2.0" }
+chrono = { version = "0.4.26", features = ["serde"] }
 
 [dev-dependencies]
 tokio = { version = "1.28.2", features = ["rt", "macros"] }
--- a/crates/bitwarden/Cargo.toml
+++ b/crates/bitwarden/Cargo.toml
@@ -46,13 +46,13 @@ rand = ">=0.8.5, <0.9"
 num-bigint = ">=0.4, <0.5"
 num-traits = ">=0.2.15, <0.3"
 uuid = { version = ">=1.3.3, <2.0", features = ["serde"] }
+chrono = { version = ">=0.4.26, <0.5", features = ["serde"], default-features = false }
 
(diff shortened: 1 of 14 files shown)


crates/bitwarden/Cargo.toml
```diff
@@ -52,6 +52,7 @@ getrandom = { version = ">=0.2.9", features = ["js"] }
 
 bitwarden-api-identity = { path = "../bitwarden-api-identity", version = "0.2.0" }
 bitwarden-api-api = { path = "../bitwarden-api-api", version = "0.2.0" }
+chrono = { version = "0.4.26", features = ["serde"] }
```
Hinton, line 55: Move up to the other dependencies. Use loose version and turn off default-features since we don't want `oldtime`.

```suggestion
chrono = { version = ">=0.4.26, <0.5", features = ["serde"], default-features = false }
```

crates/bitwarden/src/mobile/vault/folders/request.rs
Hinton, line 1: request and response files aren't scaleable. For platform we put each "command" in it's own file along with the request and response models for that command.

crates/bitwarden/src/mobile/vault/folders/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+    revision_date: DateTime<Utc>,
+}
+
+#[derive(Serialize, Deserialize, Debug, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct FolderView {
+    id: Uuid,
+    name: String,
+    revision_date: DateTime<Utc>,
+}
```
Hinton, line 32: I would like to move these up into a root `vault` module.

Hinton (commented):
Still reviewing but here are some initial comments

crates/bitwarden/src/mobile/vault/folders/mod.rs
dani-garcia: So just in a `mobile/vault/folders.rs` file, and keeping the `folders` folder for everything else?

crates/bitwarden/src/lib.rs
```diff
@@ -59,6 +59,7 @@ pub mod platform;
 #[cfg(feature = "secrets")]
 pub mod secrets_manager;
 mod util;
+pub mod vault;
```
Hinton, line 62: I think we can move the mobile check here?

dani-garcia: Oh yeah, that's true, changed in 1ddda6e

crates/bitwarden/src/vault/mod.rs
```diff
@@ -0,0 +1,3 @@
+mod folder;
+
+pub use folder::*;
```
Hinton, line 3: @MGibson1 had some hesitation against re-exporting wildcard since it's not trivial to know what gets re-exported.

Hinton (approved):

crates/bitwarden/src/vault/mod.rs
dani-garcia: I can agree with that yeah, specially for `pub use` I'd rather be specific, I've removed the wildcard now.

crates/bitwarden/src/mobile/vault/mod.rs
```diff
@@ -0,0 +1,5 @@
+mod client_folders;
+mod client_vault;
+mod folders;
+
+pub use folders::*;
```
Hinton, line 5: You also have pub use here, and in `folders/mod`

Hinton (approved):

PR #137, opened 2023-07-28
Update serde_repr to fix some warnings

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
The macro-generated code of `serde_repr` serialization implementation contains a lowercase struct, which generates a warning in rust-analyzer, seems like it doesn't show in CI, and it's not important, but I think it's a good idea to update the dependency to make the warning go away. 

https://github.com/dtolnay/serde-repr/releases/tag/0.1.16
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2551,9 +2551,9 @@ dependencies = [
 
 [[package]]
 name = "serde_repr"
-version = "0.1.15"
+version = "0.1.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


Hinton (commented):
Do we need to actually update the dependency version in cargo.toml? Since it's just a warning updating the version in the lockfile will resolve the issue for us without forcing any consumers to use the new version.

dani-garcia:
My thinking was that some consumers of the crate might also hit that warning and break their CI, but I'm pretty sure rust doesn't emit warnings for third party crates, so maybe it's not useful.

Hinton (approved):

PR #138, opened 2023-07-28
Fix invalid target_arch

dani-garcia:
## Type of change
```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Some `cfg`s were checking for a `target_arch` of `wasm`, this value is not valid, the only valid values are `wasm32` and `wasm64`.

You can check the possible values with this command:
```sh
rustc +nightly -Z unstable-options --print all-target-specs-json | jq '[ to_entries[] | { "target": .key, "arch": .value.arch, "target-family": .value."target-family" } | select(.arch | startswith("wasm")) ]'
```


Hinton (approved):
Do we need to care about wasm64?

PR #139, opened 2023-07-28
Fix some clippy lints

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Fix some of the lints that appear when running `cargo clippy`. Mostly removing unnecessary references, `into()`s and `clone()`s. There should be no API or behavior changes.


Hinton (dismissed):

crates/bitwarden/src/auth/api/response/identity_token_response.rs
```diff
@@ -14,30 +14,30 @@ pub enum IdentityTokenResponse {
     Authenticated(IdentityTokenSuccessResponse),
     Payload(IdentityTokenPayloadResponse),
     Refreshed(IdentityTokenRefreshResponse),
-    TwoFactorRequired(IdentityTwoFactorResponse),
+    TwoFactorRequired(Box<IdentityTwoFactorResponse>),
```
Hinton, line 17: Why did we need a box here?

crates/bitwarden/src/client/mod.rs
```diff
@@ -4,6 +4,7 @@ pub(crate) use client::*;
 pub(crate) mod access_token;
 #[cfg(any(feature = "internal", feature = "mobile"))]
 pub(crate) mod auth_settings;
+#[allow(clippy::module_inception)]
```
Hinton, line 7: Is it best practice for us to look into renaming these files?

Hinton (dismissed):
LGTM pending response on the box question.

crates/bitwarden/src/auth/api/response/identity_token_response.rs
dani-garcia: `IdentityTwoFactorResponse` is too big and so it bloats the size of the `IdentityTokenResponse` enum, as it's size is based on the biggest element it contains.

The size in bytes of the elements:
```rust
        assert_eq!(240, std::mem::size_of::<IdentityTokenSuccessResponse>());
        assert_eq!(128, std::mem::size_of::<IdentityTokenPayloadResponse>());
        assert_eq!(104, std::mem::size_of::<IdentityTokenRefreshResponse>());
        assert_eq!(656, std::mem::size_of::<IdentityTwoFactorResponse>());
        assert_eq!(120, std::mem::size_of::<IdentityCaptchaResponse>());
        assert_eq!(96, std::mem::size_of::<IdentityTokenFailResponse>());
```

crates/bitwarden/src/client/mod.rs
dani-garcia: This seems to be meant to avoid beginners accidentally defining the module twice by not fully understanding the module system, at least according to the lint help: https://rust-lang.github.io/rust-clippy/master/index.html#/module_inception

The `client` module there is private, so I think it's not a huge issue, but if we can come up with a better name that should be preferable IMO.

dani-garcia:
Had to make some changes for clippy to not complain about Cipher, as it was very big (2600 bytes). This was because we declare all the variants (login, identity, card) and they bloat the size even when they are unused. Ultimately I'd like to change it to an enum, but for the moment it's best to just box the result, to keep it as similar as possible as the server API responses.

Hinton (approved):

PR #140, opened 2023-07-28
Kotlin & Swift bindings for SDK

Hinton:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Provide bindings for Kotlin and Swift through [UniFFI](https://github.com/mozilla/uniffi-rs/).

## Code changes

## Before you submit


dani-garcia (dismissed):

.editorconfig
```diff
@@ -22,3 +22,7 @@ indent_size = 4
 [*.{kt,java,xml,gradle}]
 indent_style = space
 indent_size = 4
+
+[*.{xml}]
+# VS Code XML extension removes the final newline
+insert_final_newline = false
```
MGibson1, line 28: :grumble:, I want newlines _everywhere_

crates/bitwarden-uniffi/src/lib.rs
```diff
@@ -0,0 +1,20 @@
...
+        Self(bitwarden::Client::new(settings))
+    }
+
+    pub fn run_command(&self, input_str: String) -> String {
+        input_str
+    }
+
+    fn parse_settings(_settings_input: Option<String>) -> Option<ClientSettings> {
+        None
+    }
```
MGibson1, line 19: This is precisely the json client, I don't understand why we need a new struct to represent a uniffi client

crates/bitwarden-uniffi/swift/Tests/BitwardenSdkTests/BitwardenSdkTests.swift
```diff
@@ -0,0 +1,11 @@
+import XCTest
+@testable import BitwardenSdk
+
+final class BitwardenSdkTests: XCTestCase {
+    func testExample() throws {
+        // This is an example of a functional test case.
+        // Use XCTAssert and related functions to verify your tests produce the correct
+        // results.
+        XCTAssertEqual(BitwardenSdk().text, "Hello, World!")
```
MGibson1, line 9: Why does this work? we aren't defining anything like this in the sdk.

MGibson1 (commented):
I'm coming in to this with little to no knowledge of uniffi and mobile development in general, so I'm probably missing a lot.

I don't understand why this needs to be a new client rather than a tie to the JSON one, and I don't understand the folder structure. Swift and Android appear to be `language` examples, why aren't they in that directory? Do they actually make any calls to the SDK in this PR?

What is the interface in each of these languages? I don't see an idiomatic client like we've used for python, node, or C#.

Is the goal of this PR to show that swift and kotlin packages can be built with a uniffi-style pipeline? If so, the examples feel too integrated into the crate.

crates/bitwarden-uniffi/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "bitwarden-uniffi"
+version = "0.1.0"
+edition = "2021"
+rust-version = "1.57"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[lib]
+crate-type = ["staticlib", "cdylib"]
```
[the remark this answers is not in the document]
Hinton: We're actually using the dynamic libs in most cases. This should let us reuse the same lib in multiple binaries without having to bloat every executable.

crates/bitwarden-uniffi/android/README.md
```diff
@@ -0,0 +1,44 @@
...
+./build-schemas.sh
+```
+
+### Publish
+
+```bash
+export GITHUB_ACTOR=username
+export GITHUB_TOKEN=token
+
+./gradlew sdk:publish
```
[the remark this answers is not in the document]
Hinton: Maybe not. The readme won't really be visible externally though and is more instructions for the developers how to build the thing locally. In that sense knowing how to publish the package is a bit beneficial.

.editorconfig
Hinton: Yea me to. But it resulted in vscode adding a newline, then immediately removing it due to running formater.

crates/bitwarden-uniffi/src/lib.rs
Hinton: Yes it is. There is a follow up PR which changes that though. I wanted to keep this PR somewhat light and not introduce more work into it, it's essentially the minimum code we needed to get something working, you might notice that `run_command` doesn't actually run a command just echoes the input back.

Hinton:
> I don't understand why this needs to be a new client rather than a tie to the JSON one, and I don't understand the folder structure. Swift and Android appear to be language examples, why aren't they in that directory? Do they actually make any calls to the SDK in this PR?

I've moved the language stubs to the languages directory.

> What is the interface in each of these languages? I don't see an idiomatic client like we've used for python, node, or C#.

The idomatic client is generated from uniffi. An example is available in https://github.com/bitwarden/sdk-swift/blob/main/Sources/BitwardenSdk/BitwardenSDK.swift.

> Is the goal of this PR to show that swift and kotlin packages can be built with a uniffi-style pipeline? If so, the examples feel too integrated into the crate.

The PR sets up the plumbing needed build swift and kotlin artifacts.
- https://github.com/bitwarden/sdk/pull/149 builds on this PR and introduces the actual functionality in the uniffi sdk.
- https://github.com/bitwarden/sdk/pull/154 expands the kotlin app sample to call some methods in the sdk.

MGibson1 (dismissed):

MGibson1 (dismissed):

.github/workflows/build-android.yml
```diff
@@ -0,0 +1,126 @@
...
+        with:
+          name: android-${{ matrix.settings.target }}
+          path: ./target/${{ matrix.settings.target }}/release/libbitwarden_uniffi.so
+          if-no-files-found: error
+
+  combine:
+    runs-on: ubuntu-22.04
+    needs:
+      - build
+
```
vgrassia, line 60:
```suggestion
    needs: build
```

.github/workflows/build-android.yml
```diff
@@ -0,0 +1,126 @@
...
+    runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        settings:
+          - target: aarch64-linux-android
+          - target: armv7-linux-androideabi
+          - target: x86_64-linux-android
+          - target: i686-linux-android
+
```
vgrassia, line 26:
```suggestion
```

.github/workflows/build-android.yml
```diff
@@ -0,0 +1,126 @@
...
+          mv /home/runner/work/sdk/sdk/android-armv7-linux-androideabi/libbitwarden_uniffi.so ./armeabi-v7a/libbitwarden_uniffi.so
+          mv /home/runner/work/sdk/sdk/android-aarch64-linux-android/libbitwarden_uniffi.so ./arm64-v8a/libbitwarden_uniffi.so
+          mv /home/runner/work/sdk/sdk/android-i686-linux-android/libbitwarden_uniffi.so ./x86/libbitwarden_uniffi.so
+          mv /home/runner/work/sdk/sdk/android-x86_64-linux-android/libbitwarden_uniffi.so ./x86_64/libbitwarden_uniffi.so
+
+      - name: Generate bindings
+        working-directory: languages/kotlin
+        run: ./build-schemas.sh
+
+      - name: Login to Azure - Prod Subscription
```
vgrassia, line 107:
```suggestion
      - name: Login to Azure - CI Subscription
```

vgrassia (changes requested):

vgrassia (approved):

PR #142, opened 2023-07-31
Add caching to lint

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Code changes

## Before you submit

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -24,6 +24,9 @@ jobs:
         with:
           toolchain: stable
 
+      - name: Cache cargo registry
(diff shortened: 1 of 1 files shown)


dani-garcia (approved):
LGTM.

One small only somewhat related concern, the other day I noticed we seem to be hitting cache limits pretty much all the time: https://github.com/bitwarden/sdk/actions/caches

Not sure if there is much we can do here, but I'm worried we're going to end up in a situation where most of our caches end up evicted before they are used again, to the point that caching becomes counterproductive to the goal of reducing CI times.

PR #143, opened 2023-07-31
[PM-3136] Cipher encryption

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
Implement cipher encryption for the mobile SDK


crates/bitwarden/src/mobile/vault/client_ciphers.rs
```diff
@@ -0,0 +1,25 @@
...
+
+pub struct ClientCiphers<'a> {
+    pub(crate) client: &'a Client,
+}
+
+impl<'a> ClientCiphers<'a> {
+    pub async fn encrypt(&self, req: CipherEncryptRequest) -> Result<CipherEncryptResponse> {
+        let enc = self.client.get_encryption_settings()?;
+
+        let cipher = req.cipher.encrypt(enc, &None)?;
```
Hinton, line 13: Ciphers can belong to an organization, although they contain the organization_id themselves if they do.

crates/bitwarden/src/vault/cipher/cipher.rs
```diff
@@ -119,3 +123,34 @@ pub struct CipherListView {
...
+            id: self.id,
+            organization_id: self.organization_id,
+            folder_id: self.folder_id,
+            collection_ids: self.collection_ids,
+            name: self.name.encrypt(enc, org_id)?,
+            notes: self.notes.encrypt(enc, org_id)?,
+            r#type: self.r#type,
+            login: self.login.encrypt(enc, org_id)?,
+            identity: self.identity.encrypt(enc, org_id)?,
+            card: self.card.encrypt(enc, org_id)?,
```
Hinton, line 140: What happens if we get null here?

Hinton (commented):
Would be nice with some tests.

crates/bitwarden/src/vault/cipher/cipher.rs
dani-garcia: Option<T> implements Encryptable and Decryptable if T implements it, and just maps over the optional, so null stays null, and any other values get encrypted or decrypted as usual.

https://github.com/bitwarden/sdk/blob/4ae062e20e6d6bee2abd89af9dda105d19269bdc/crates/bitwarden/src/crypto.rs#L389-L399

Hinton (approved):

PR #145, opened 2023-07-31
[PM-3133] Cipher List Decryption

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
Implement cipher list decryption for the mobile SDK. Depends on #144

### DIFF
--- a/crates/bitwarden/src/mobile/vault/client_ciphers.rs
+++ b/crates/bitwarden/src/mobile/vault/client_ciphers.rs
@@ -5,8 +5,8 @@ use crate::{
 };
 
 use super::{
-    client_vault::ClientVault, CipherDecryptRequest, CipherDecryptResponse, CipherEncryptRequest,
-    CipherEncryptResponse,
+    client_vault::ClientVault, CipherDecryptListRequest, CipherDecryptListResponse,
+    CipherDecryptRequest, CipherDecryptResponse, CipherEncryptRequest, CipherEncryptResponse,
 };
 
 pub struct ClientCiphers<'a> {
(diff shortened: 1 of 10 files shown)


crates/bitwarden/src/mobile/vault/client_ciphers.rs
```diff
@@ -29,6 +29,17 @@ impl<'a> ClientCiphers<'a> {
...
         Ok(CipherDecryptResponse { cipher })
     }
+
+    pub async fn decrypt_list(
+        &self,
+        req: CipherDecryptListRequest,
+    ) -> Result<CipherDecryptListResponse> {
+        let enc = self.client.get_encryption_settings()?;
+
+        let ciphers = req.ciphers.decrypt(enc, &None)?;
```
Hinton, line 39: Does this understand itself which decrypt interface to use?

crates/bitwarden/src/vault/cipher/cipher.rs
```diff
@@ -185,3 +185,87 @@ impl Decryptable<CipherView> for Cipher {
...
+                    if number_len > 4 {
+                        if !sub_title.is_empty() {
+                            sub_title.push_str(", ");
+                        }
+
+                        // On AMEX cards we show 5 digits instead of 4
+                        let digit_count = match &number[0..2] {
+                            "34" | "37" => 5,
+                            _ => 4,
+                        };
```
Hinton, line 221: Should we use the card type instead of number for this? Not sure what the TS implementation uses.

crates/bitwarden/src/mobile/vault/client_ciphers.rs
dani-garcia: Yeah, the implementation for Vec just maps over the ciphers and calls the decrypt implementation of Cipher. Org id is ignored and each cipher uses its own.

https://github.com/bitwarden/sdk/blob/35c3c66d7bf7992624cb125688158553891f1437/crates/bitwarden/src/crypto.rs#L407-L412

crates/bitwarden/src/vault/cipher/cipher.rs
dani-garcia: The clients seem to use the number the same way, not sure if there's a reason for that (Maybe in case a user mistypes the card?)

https://github.com/bitwarden/clients/blob/fb74c2d6eeef9d974036e58dbba5dfe30e992dbe/libs/common/src/vault/models/view/card.view.ts#L58-L60

PR #146, opened 2023-07-31
Fix fingerprint implementation

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Input needs to be B64 encoded to prevent encoding issues. Output should be a response otherwise we have problems mapping it to errors.

## Before you submit


dani-garcia (approved):
LGTM, if there's a problem with returning a `String`, I think `hash_password` does the same so maybe we have to change that one as well.

PR #149, opened 2023-07-31
[PM-3376] Uniffi macro

Hinton:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Configure the UniFFI macro to generate the binding definitions. Wire up kdf and vault operations.

## Before you submit


crates/bitwarden-uniffi/android/build-schemas.sh
```diff
@@ -1 +1,7 @@
-cargo run -p uniffi-bindgen generate ../src/sdk.udl --language kotlin --out-dir sdk/src/main/java
+cargo build --package bitwarden-uniffi --release --all-features
+cargo run -p uniffi-bindgen generate \
+  ../../../target/aarch64-linux-android/release/libbitwarden_uniffi.so \
```
dani-garcia, line 3: On the previous line we're compiling for the host target, so I think it would make more sense to use the host build artifact, otherwise we depend on the user having compiled it themselves previously.
```suggestion
  ../../../target/release/libbitwarden_uniffi.so \
```

crates/bitwarden/src/uniffi_support.rs
```diff
@@ -0,0 +1,62 @@
+use std::{num::NonZeroU32, str::FromStr};
+
+use uuid::Uuid;
+
+use crate::{crypto::CipherString, error::Error, UniffiCustomTypeConverter};
+
+uniffi::custom_type!(NonZeroU32, u32);
+
+impl UniffiCustomTypeConverter for NonZeroU32 {
```
Hinton, line 9: @dani-garcia I re-wrote these to use the new macro from uniffi.

crates/bitwarden/src/mobile/vault/ciphers/cipher_decrypt.rs
```diff
@@ -5,12 +5,14 @@ use crate::vault::{Cipher, CipherView};
 
 #[derive(Serialize, Deserialize, Debug, JsonSchema)]
 #[serde(rename_all = "camelCase", deny_unknown_fields)]
+#[cfg_attr(feature = "mobile", derive(uniffi::Record))]
 pub struct CipherDecryptRequest {
-    pub cipher: Box<Cipher>,
+    pub cipher: Cipher,
```
Hinton, line 10: @dani-garcia I had to remove these box to have uniffi happy. I didn't want to implement custom transformers for everything. Not sure if there is another way to better handle this.

dani-garcia: The box here is mostly to fix Clippy's complaining about unbalanced enums, seems like we can just change `` in `bitwarden-json` to be defined like this:
```rust
pub enum MobileCiphersCommand {
    Encrypt(Box<CipherEncryptRequest>),
    Decrypt(Box<CipherDecryptRequest>),
    DecryptList(CipherDecryptListRequest),
}
```
And then just using `*cmd` to access the inner value of the box when matching against the commands, that seems to remove the clippy lint to me at least.

crates/bitwarden/src/vault/cipher/linked_id.rs
```diff
@@ -4,21 +4,24 @@ use serde_repr::{Deserialize_repr, Serialize_repr};
...
 #[derive(Clone, Copy, PartialEq, Eq, Serialize, Deserialize, Debug, JsonSchema)]
 #[serde(untagged)]
+#[cfg_attr(feature = "mobile", derive(uniffi::Enum))]
 pub enum LinkedIdType {
-    Login(LoginLinkedIdType),
-    Card(CardLinkedIdType),
-    Identity(IdentityLinkedIdType),
+    Login { val: LoginLinkedIdType },
+    Card { val: CardLinkedIdType },
+    Identity { val: IdentityLinkedIdType },
```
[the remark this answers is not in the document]
dani-garcia: I think the problem here is uniffi doesn't know how to represent enum tuple variants, only the struct variants. We should check what the generated code looks like in Kotlin and Swift this way, if it's too cumbersome to use we might just be better off combining them all into a big enum.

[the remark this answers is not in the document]
Hinton: We could perhaps implement our own converters for it?

crates/bitwarden/src/uniffi_support.rs
dani-garcia: Nice, looks much simpler now with the latest uniffi changes!

crates/bitwarden/src/vault/cipher/linked_id.rs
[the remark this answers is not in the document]
dani-garcia: If we want a simple option, I guess we can go with something like this, and just expose it as a number?
```rust
uniffi::custom_type!(LinkedIdType, u32);

impl crate::UniffiCustomTypeConverter for LinkedIdType {
    type Builtin = u32;

    fn into_custom(val: Self::Builtin) -> uniffi::Result<Self> {
        let val = serde_json::Value::Number(val.into());
        Ok(serde_json::from_value(val)?)
    }

    fn from_custom(obj: Self) -> Self::Builtin {
        serde_json::to_value(&obj)
            .expect("LinkedIdType should be serializable")
            .as_u64()
            .expect("Not a numeric enum value") as u32
    }
} 
```

crates/bitwarden/src/vault/cipher/linked_id.rs
```diff
@@ -10,15 +10,34 @@ pub enum LinkedIdType {
     Identity(IdentityLinkedIdType),
 }
 
+[#cfg(feature = "mobile")]
```
dani-garcia, line 13:
```suggestion
#[cfg(feature = "mobile")]
```

crates/bitwarden/src/vault/cipher/linked_id.rs
```diff
@@ -10,15 +10,34 @@ pub enum LinkedIdType {
...
+use crate::UniffiCustomTypeConverter;
+[#cfg(feature = "mobile")]
```
dani-garcia, line 15:
```suggestion
#[cfg(feature = "mobile")]
```

crates/bitwarden/src/vault/cipher/linked_id.rs
```diff
@@ -10,15 +10,34 @@ pub enum LinkedIdType {
...
+uniffi::custom_type!(LinkedIdType, u32);
+[#cfg(feature = "mobile")]
```
dani-garcia, line 17:
```suggestion
#[cfg(feature = "mobile")]
```

dani-garcia (approved):

PR #164, opened 2023-08-13
Scaffold generators

Hinton:
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Scaffold the password and passphrase generator. Wire up a command for password in the bw cli.

## Before you submit

### DIFF
--- a/crates/bw/src/main.rs
+++ b/crates/bw/src/main.rs
@@ -34,7 +34,7 @@ enum Commands {
     #[command(long_about = "Pull the latest vault data from the server")]
     Sync {},
 
-    #[command(long_about = "Pull the latest vault data from the server")]
+    #[command(long_about = "Password and passphrase generators")]
     Generate {
         #[command(subcommand)]
         command: GeneratorCommands,
@@ -93,11 +93,7 @@ struct PasswordGeneratorArgs {
     )]
     special: bool,
 
-    #[arg(
-        long,
-        default_value = "16",
-        help = "Minimum length of generated password"
-    )]
+    #[arg(long, default_value = "16", help = "Length of generated password")]
     length: u8,
 }
 
--- a/crates/bw/src/main.rs
+++ b/crates/bw/src/main.rs
@@ -1,4 +1,4 @@
-use bitwarden::client::client_settings::ClientSettings;
+use bitwarden::{client::client_settings::ClientSettings, tool::PasswordGeneratorRequest};
 use bitwarden_cli::{install_color_eyre, Color};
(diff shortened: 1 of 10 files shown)


crates/bw/src/main.rs
```diff
@@ -33,6 +33,12 @@ enum Commands {
 
     #[command(long_about = "Pull the latest vault data from the server")]
     Sync {},
+
+    #[command(long_about = "Pull the latest vault data from the server")]
```
dani-garcia, line 37: This comment needs to be updated

crates/bw/src/main.rs
```diff
@@ -62,6 +68,39 @@ enum ItemCommands {
...
+        long,
+        action,
+        help = "Include special characters (!@#$%^&*)"
+    )]
+    special: bool,
+
+    #[arg(
+        long,
+        default_value = "16",
+        help = "Minimum length of generated password"
```
dani-garcia, line 99: There's no minimum and maximum, the length of the generated password is exactly what's provided, so the comment should probably be "Length of generated password" right?

crates/bw/src/main.rs
```diff
@@ -62,6 +68,39 @@ enum ItemCommands {
...
+#[derive(Args, Clone)]
+struct PasswordGeneratorArgs {
+    #[arg(short = 'l', long, action, help = "Include lowercase characters (a-z)")]
+    lowercase: bool,
+
+    #[arg(short = 'u', long, action, help = "Include uppercase characters (A-Z)")]
+    uppercase: bool,
+
+    #[arg(short = 'n', long, action, help = "Include numbers (0-9)")]
+    numbers: bool,
```
dani-garcia, line 86: Not sure about these parameters, by default they are all set to false, so running `bw generate password` we won't have any character types selected, which would probably have to return an error. 

You'll have to go out of your way to do `bw generate password -l -u -n` to get an alphanumeric password.

I think having a reasonable default for `bw generate password` is a good idea. Maybe we should invert the flags and make them `no-lowercase`, `no-uppercase` and `no-numbers`?

crates/bw/src/main.rs
Hinton: Yep, codepilot 😠

crates/bw/src/main.rs
Hinton: My thinking is that if all are false the default applies. Which is how the CLI currently works.

dani-garcia (approved):

PR #168, opened 2023-08-17
Improve doc

Hinton:
## Type of change

```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Change links in the doc to be lowercase.

## Before you submit

### DIFF
--- a/languages/kotlin/doc.md
+++ b/languages/kotlin/doc.md
@@ -1,0 +1,522 @@
+# Bitwarden Mobile SDK
+
+Auto generated documentation for the Bitwarden Mobile SDK. For more information please refer to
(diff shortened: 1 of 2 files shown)


dani-garcia:
Do we want to keep the doc file in the repo? I feel like that should be .gitignored like we do with other auto generated stuff like schemas and uniffi generated code.

Hinton:
For now, we need it to be accessible for ease of use.

dani-garcia (approved):
Fair enough, we have to be careful to keep it updated in that case, we don't want the docs getting outdated and causing confusion.

LGTM as long as it's not a long term thing

PR #176, opened 2023-08-21
Remove autogenerated NAPI bindings

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
These files get generated each time you run `npm run build`, there doesn't seem to be a point to include them in the repo, unless I'm missing something.


Hinton (changes requested):
We never actually run npm run build in the CI. So the published files are using the binding.js.

We might want to look into changing that though. Can't remember if we adjusted anything in them though.

dani-garcia:
We run build on the build-napi workflow, but we don't copy those binding changes to the release-napi workflow: https://github.com/bitwarden/sdk/blob/fe632268b37c087e39b3012b67fafa8ffcb059c2/.github/workflows/build-napi.yml#L33
So we'd probably need to copy the bindings over when we copy the binaries. 
At the moment using the repo bindings seems to be fine for now but it might end up causing incompatibilities if the bindings and the generated binaries get too out of sync.

This is the diff between what is currently on the repo and what `npm run build` generates, seems the changes are to add support for a macos universal build, probably just coming from a napi-rs update, so it doesn't seem like we have modified them in any way:

```diff
diff --git a/crates/bitwarden-napi/binding.js b/crates/bitwarden-napi/binding.js
index c3cf320..ada74d6 100644
--- a/crates/bitwarden-napi/binding.js
+++ b/crates/bitwarden-napi/binding.js
@@ -1,4 +1,10 @@
-const { existsSync, readFileSync } = require("fs");
+/* tslint:disable */
+/* eslint-disable */
+/* prettier-ignore */
+
+/* auto-generated by NAPI-RS */
+
+const { existsSync, readFileSync } = require('fs')
 const { join } = require("path");

 const { platform, arch } = process;
@@ -11,7 +17,8 @@ function isMusl() {
   // For Node 10
   if (!process.report || typeof process.report.getReport !== "function") {
     try {
-      return readFileSync("/usr/bin/ldd", "utf8").includes("musl");
+      const lddPath = require("child_process").execSync("which ldd").toString().trim();
+      return readFileSync(lddPath, "utf8").includes("musl");
     } catch (e) {
       return true;
     }
@@ -95,6 +102,15 @@ switch (platform) {
     }
     break;
   case "darwin":
+    localFileExisted = existsSync(join(__dirname, "sdk-napi.darwin-universal.node"));
+    try {
+      if (localFileExisted) {
+        nativeBinding = require("./sdk-napi.darwin-universal.node");
+      } else {
+        nativeBinding = require("@bitwarden/sdk-napi-darwin-universal");
+      }
+      break;
+    } catch {}
     switch (arch) {
       case "x64":
         localFileExisted = existsSync(join(__dirname, "sdk-napi.darwin-x64.node"));
```

Hinton:
Closing this. We can re-explore this in the future.

PR #178, opened 2023-08-21
Refactor auth directory

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Refactor auth directory in the bitwarden crate to follow the same structure as the rest of the library.

## Code changes

## Before you submit


crates/bitwarden/src/auth/login/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+    bitwarden_api_identity::{
+        apis::accounts_api::accounts_prelogin_post,
+        models::{PreloginRequestModel, PreloginResponseModel},
+    },
+};
+
+pub mod response;
+
+mod password;
+pub use password::PasswordLoginResponse;
```
dani-garcia, line 16: Shouldn't these two also be behind the `internal` feature?

Hinton: Ideally, but for some reason the `PasswordLoginResponse` is used to unparse the other responses. I think the response parsing could use some more work but this PR is a good step in the right direction.

dani-garcia (approved):

PR #181, opened 2023-08-21
Fix panic when re-registering a logger

### LINKED ISSUE #134: Bitwarden NAPI initializes logger multiple times
From #112 

### Steps To Reproduce

Create multiple instances of the Bitwarden SDK.

### Expected Result

Multiple concurrent instances of Bitwarden Client.

### Actual Result

```
thread '<unnamed>' panicked at 'Builder::init should not be called after logger initialized: SetLoggerError(())', C:\Users\runneradmin\.cargo\registry\src\github.com-1ecc6299db9ec823\env_logger-0.10.0\src\lib.rs:816:14
```

### Screenshots or Videos

_No response_

### Additional Context

We need to look into a way to iniitalize the logger once, maybe using https://doc.rust-lang.org/stable/std/sync/struct.Once.html.

### Operating System

Windows, macOS, Linux

### Operating System Version

_No response_

### Build Version

0.3.0

### Issue Tracking Info

dani-garcia:
## Type of change
```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Fix panic when re-registering a logger. This should only happen if our logger got registered on a previous call, so there's no need to register it again.

Fixes #134, but it's also applicable to `bitwarden-c`

### DIFF
--- a/crates/bitwarden-c/src/c.rs
+++ b/crates/bitwarden-c/src/c.rs
@@ -24,7 +24,8 @@ pub async extern "C" fn run_command(
 // Init client, potential leak! You need to call free_mem after this!
 #[no_mangle]
 pub extern "C" fn init(c_str_ptr: *const c_char) -> *mut Client {
-    env_logger::init();
+    // This will only fail if another logger was already initialized, so we can ignore the result
+    let _ = env_logger::try_init();
     if c_str_ptr.is_null() {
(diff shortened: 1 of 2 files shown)


Hinton (approved):
Looks good, thanks for resolving this.

PR #182, opened 2023-08-22
Also build non sim ios

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Build non simulator ios artifact

## Before you submit

### DIFF
--- a/languages/swift/build.sh
+++ b/languages/swift/build.sh
@@ -8,6 +8,7 @@ mkdir tmp
 
 # Build native library
 cargo build --package bitwarden-uniffi --target aarch64-apple-ios-sim --release
+cargo build --package bitwarden-uniffi --target aarch64-apple-ios --release
 
 # Generate swift bindings
 cargo run -p uniffi-bindgen generate \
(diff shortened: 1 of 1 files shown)


languages/swift/build.sh
```diff
@@ -31,6 +32,8 @@ cat ./tmp/bindings/BitwardenFFI.modulemap ./tmp/bindings/BitwardenCoreFFI.module
 xcodebuild -create-xcframework \
   -library ../../target/aarch64-apple-ios-sim/release/libbitwarden_uniffi.dylib \
   -headers ./tmp/Headers \
+  -library ../../target/aarch64-apple-ios/release/libbitwarden_uniffi.dylib \
+  -headers ./tmp/Headers \
```
dani-garcia, line 36: Do we need to specify -headers twice? It's already there on the previous line.

Hinton: We do. It's headers per library, I guess so you can support different headers for different builds.

dani-garcia (approved):

PR #184, opened 2023-08-24
[PM-3130] Send decryption for mobile SDK

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
Implement send decryption for mobile SDK:
- `decrypt` decrypts the send metadata: name, notes, file name and text.
- `decrypt_file` decrypts the send file using the key in the provided send. The file needs to be downloaded and stored separately.
- `decrypt_buffer` same as the previous, but reads from and decrypts the file into memory.


crates/bitwarden/src/crypto.rs
```diff
@@ -137,6 +138,36 @@ impl FromStr for CipherString {
...
+        if buf.is_empty() {
+            return Err(CSParseError::NoType.into());
+        }
+        let enc_type = buf[0];
+
+        match enc_type {
+            0 => unimplemented!(),
+            1 | 2 => {
+                let iv = buf[1..17].try_into().unwrap();
+                let mac = buf[17..49].try_into().unwrap();
```
Hinton, line 153: Should we unwrap here? Could result in fatal errors

crates/bitwarden/src/vault/send.rs
```diff
@@ -0,0 +1,174 @@
...
+impl Send {
+    pub(crate) fn get_send_key(
+        &self,
+        enc: &EncryptionSettings,
+        org_id: &Option<Uuid>,
+    ) -> Result<EncryptionSettings> {
+        let key: Vec<u8> = enc.decrypt_bytes(&self.key, org_id)?;
+        let key = stretch_key(key.try_into().unwrap(), "send", Some("send"));
+        Ok(EncryptionSettings::new_single_key(key))
+    }
```
Hinton, line 120: Can we add a test for this?

Hinton:
Clippy seems unhappy as well.

Hinton (approved):

PR #188, opened 2023-08-25
Move KDF to auth

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Move password_hash from kdf to auth.

## Before you submit


dani-garcia (commented):
Looks good, but needs to update the android and ios sample apps as well

PR #190, opened 2023-08-25
SM-774: Write Help Text to 'stderr' When No Args Are Provided

coltonhurst:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

When no arguments are provided to `bws`, we should write the help text to `stderr` rather than `stdout` to be consistent with the existing `bws subcommand` behavior.

Potentially relevant `clap` GitHub discussions:

- [https://github.com/clap-rs/clap/pull/4248](https://github.com/clap-rs/clap/pull/4248)
- [https://github.com/clap-rs/clap/pull/3289](https://github.com/clap-rs/clap/pull/3289)
- [https://github.com/clap-rs/clap/issues/1788](https://github.com/clap-rs/clap/issues/1788)

## Code changes

- **main.rs:** Print with `eprintln`

## Before you submit


Hinton:
I think this changes `bws` to be more consistent with `bws subcommand`. So from that perspective I'm fine with this change. We don't want to change `--help` though.

@dani-garcia any thoughts?

We should document this in the changelog though. Not sure if this qualifies as a breaking change, I have a hard time imagining someone depending on the stdout output when not using any commands.

crates/bws/src/main.rs
```diff
@@ -246,7 +246,7 @@ async fn process_commands() -> Result<()> {
 
     let Some(command) = cli.command else {
         let mut cmd = Cli::command();
-        cmd.print_help()?;
+        eprintln!("{}", cmd.render_help());
```
dani-garcia, line 249:
```suggestion
        eprintln!("{}", cmd.render_help().ansi());
```

This technically breaks the styling that the clap help does, calling `.ansi()` will return the formatting. (We might want to do it only when color is enabled though)

Top is the current master, bottom is this PR without the ansi change:
image

Printing it like this loses the underscores and the bold headers.

coltonhurst: Good catch, fixed in: 5b08cfd

PR #198, opened 2023-08-28
[PM-3431] Generate registration keys

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Implement key generations required for registration.

## Before you submit


crates/bitwarden/src/client/encryption_settings.rs
```diff
@@ -119,7 +119,7 @@ impl EncryptionSettings {
     pub(crate) fn encrypt(&self, data: &[u8], org_id: &Option<Uuid>) -> Result<EncString> {
         let key = self.get_key(org_id).ok_or(CryptoError::NoKeyForOrg)?;
 
-        let dec = encrypt_aes256(data, key.mac_key, key.key)?;
+        let dec = encrypt_aes256_hmac(data, key.mac_key.unwrap(), key.key)?;
```
dani-garcia, line 122: Should we use something like `let mac_key = key.mac_key.ok_or(CryptoError::InvalidMac)?;` instead of `unwrap()`? Or can we be sure that the `mac_key` will always be `Some`?

crates/bitwarden/src/crypto/aes_ops.rs
```diff
@@ -39,16 +39,20 @@ pub fn decrypt_aes256_hmac(
...
+pub fn encrypt_aes256(data_dec: &[u8], key: GenericArray<u8, U32>) -> Result<EncString> {
+    let mut iv = [0u8; 16];
+    rand::thread_rng().fill_bytes(&mut iv);
+    let data = cbc::Encryptor::<aes::Aes256>::new(&key, &iv.into())
+        .encrypt_padded_vec_mut::<Pkcs7>(data_dec);
+
+    Ok(EncString::AesCbc256_B64 { iv, data })
+}
+
+pub fn encrypt_aes256_hmac(
```
dani-garcia, line 51: I guess we could now change this fn to something like this to avoid duplication: 
```rust
pub fn encrypt_aes256_hmac(
    data_dec: &[u8],
    mac_key: GenericArray<u8, U32>,
    key: GenericArray<u8, U32>,
) -> Result<EncString> {
    let enc = encrypt_aes256(data_dec, key)?;
    let EncString::AesCbc256_B64 { iv, data } = enc else {
        unreachable!()
    };
    let mac = validate_mac(&mac_key, &iv, &data)?;

}
```

Don't like the partial match very much though, what do you think?

crates/bitwarden/src/crypto/rsa.rs
```diff
@@ -0,0 +1,31 @@
...
+    RsaPrivateKey, RsaPublicKey,
+};
+
+use crate::{
+    crypto::{encrypt_aes256_hmac, EncString, SymmetricCryptoKey},
+    error::{Error, Result},
+    util::BASE64_ENGINE,
+};
+
+pub(super) fn make_key_pair(key: &SymmetricCryptoKey) -> Result<(String, EncString)> {
```
dani-garcia, line 13: I feel like returning a tuple with two similar values like String and EncString could lead to confusion about which one is which. What do you think about creating a KeyPair struct with `private_key` and `public_key` fields?

crates/bitwarden/src/auth/register.rs
```diff
@@ -0,0 +1,98 @@
...
+    pub email: String,
+    pub name: Option<String>,
+    pub password: String,
+    pub password_hint: Option<String>,
+}
+
+pub(super) async fn register(
+    client: &mut Client,
+    req: &RegisterRequest,
+) -> Result<RegisterResponse> {
```
dani-garcia, line 28: Is it useful to have this response with the master password hash and encryption keys? 

I assume immediately after registering the client is going to log in, and to reuse this information we would need to handle it as a special case, which doesn't seem worth the effort for a one time thing.

crates/bitwarden/src/crypto/aes_ops.rs
Hinton: Feels like it's more pain than gain honestly. We could extract the previous operations to a separate function.

```rust
fn partial() {
    let mut iv = [0u8; 16];
    rand::thread_rng().fill_bytes(&mut iv);
    let data = cbc::Encryptor::<aes::Aes256>::new(&key, &iv.into())
        .encrypt_padded_vec_mut::<Pkcs7>(data_dec);
  (iv, data)
}

File: crates/bitwarden/src/auth/register.rs [hunk 5]
```diff
@@ -0,0 +1,84 @@
...
+            kdf: Some(bitwarden_api_identity::models::KdfType::Variant0),
+            kdf_iterations: Some(default_pbkdf2_iterations().get() as i32),
+            kdf_memory: None,
+            kdf_parallelism: None,
+            reference_data: None, // TODO: Add
+        }),
+    )
+    .await?;
+
+    unimplemented!()
```
dani-garcia, line 58: Is there anything missing here?

dani-garcia (approved):

PR #200, opened 2023-08-28
Move AES and SymmetricCryptoKey to crypto module

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Move `SymmetricCryptoKey` and `AES` operations to the crypto module.

## Before you submit

### DIFF
--- a/crates/bitwarden/src/crypto/mod.rs
+++ b/crates/bitwarden/src/crypto/mod.rs
@@ -1,9 +1,8 @@
 //! Cryptographic primitives used in the SDK
 
+use crate::error::CryptoError;
 use aes::cipher::{generic_array::GenericArray, typenum::U64, Unsigned};
 use hmac::digest::OutputSizeUser;
-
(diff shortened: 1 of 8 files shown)


crates/bitwarden/src/crypto/mod.rs
```diff
@@ -112,6 +115,16 @@ pub(crate) fn stretch_key(secret: [u8; 16], name: &str, info: Option<&str>) -> S
     SymmetricCryptoKey::try_from(key.as_slice()).unwrap()
 }
 
+pub fn decrypt(cipher: &CipherString, key: &SymmetricCryptoKey) -> Result<Vec<u8>> {
```
Hinton, line 118: I think this should be moved into EncString?

dani-garcia: That sounds reasonable, though that might cause conflicts with the Decryptable trait because they have the same name.

dani-garcia (approved):

PR #205, opened 2023-08-29
Move crypto/mod.rs to separate files

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Splitting this into a few PR to better preserve the git history.

- `stretch_key` is used to derive shareable keys. This PR extracts the code to a new file to remove it from the `mod.rs`.
- The other methods are used in combination with the master_key.

## Before you submit


crates/bitwarden/src/crypto/master_key.rs
```diff
@@ -0,0 +1,140 @@
+//! Cryptographic primitives used in the SDK
+
+#[cfg(feature = "internal")]
```
dani-garcia, line 3: This file is already defined as this in crypto/mod.rs:
```
#[cfg(feature = "internal")]
mod master_key;
```

So any cfg internal in this file is not necessary now.

crates/bitwarden/src/crypto/mod.rs
```diff
@@ -1,23 +1,25 @@
...
+mod aes_ops;
+pub use aes_ops::{decrypt_aes256, encrypt_aes256};
 mod symmetric_crypto_key;
 pub use symmetric_crypto_key::SymmetricCryptoKey;
+mod shareable_key;
+pub(crate) use shareable_key::stretch_key;
+
+#[cfg(feature = "internal")]
+mod master_key;
+#[cfg(any(feature = "internal", feature = "mobile"))]
```
dani-garcia, line 19: Mobile implies internal so this can be simply `#[cfg(feature = "mobile")]`

dani-garcia (commented):
Looks good to me, just some comments about the feature flags

dani-garcia (dismissed):

dani-garcia (approved):

PR #207, opened 2023-08-29
Refactor master key operations

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Code changes

## Before you submit

### DIFF
--- a/crates/bitwarden/src/crypto/master_key.rs
+++ b/crates/bitwarden/src/crypto/master_key.rs
@@ -2,13 +2,24 @@
 
 #[cfg(feature = "internal")]
 use aes::cipher::typenum::U32;
+
+use super::SymmetricCryptoKey;
 use {
     crate::{client::auth_settings::Kdf, error::Result},
     aes::cipher::generic_array::GenericArray,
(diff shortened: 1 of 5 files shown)


crates/bitwarden/src/crypto/master_key.rs
Hinton, line 1: I wonder if all of this should move into a crypto module in auth instead. It's not generic crypto operations but rather related only to authentication.

dani-garcia (approved):

crates/bitwarden/src/crypto/master_key.rs
```diff
@@ -1,17 +1,59 @@
...
+            &self.0.key,
+            password,
+            purpose as u32,
+        )
+        .expect("hash is a valid fixed size");
+
+        Ok(BASE64_ENGINE.encode(hash))
+    }
 
-    use super::PbkdfSha256Hmac;
```
dani-garcia, line 13: The two feature flags on this file for "internal" can be removed, because it's already feature flagged in the mod file.

PR #211, opened 2023-08-31
[PM-3138] Send encryption

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
Add Send encryption support for the mobile API

### DIFF
--- a/crates/bitwarden/src/crypto/enc_string.rs
+++ b/crates/bitwarden/src/crypto/enc_string.rs
@@ -155,6 +155,31 @@ impl EncString {
             .into()),
         }
     }
+
+    #[cfg(feature = "mobile")]
+    pub(crate) fn to_buffer(&self) -> Result<Vec<u8>> {
+        let mut buf;
+
+        match self {
+            EncString::AesCbc256_B64 { iv, data } => {
+                buf = Vec::with_capacity(1 + 16 + data.len());
+                buf.push(self.enc_type());
+                buf.extend_from_slice(iv);
+                buf.extend_from_slice(data);
+            }
+            EncString::AesCbc128_HmacSha256_B64 { iv, mac, data }
+            | EncString::AesCbc256_HmacSha256_B64 { iv, mac, data } => {
+                buf = Vec::with_capacity(1 + 16 + 32 + data.len());
+                buf.push(self.enc_type());
+                buf.extend_from_slice(iv);
+                buf.extend_from_slice(mac);
+                buf.extend_from_slice(data);
+            }
+            _ => todo!(),
+        }
+
+        Ok(buf)
+    }
 }
 
 impl Display for EncString {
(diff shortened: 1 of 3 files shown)


crates/bitwarden/src/crypto/enc_string.rs
```diff
@@ -155,6 +155,31 @@ impl EncString {
...
+            }
+            EncString::AesCbc128_HmacSha256_B64 { iv, mac, data }
+            | EncString::AesCbc256_HmacSha256_B64 { iv, mac, data } => {
+                buf = Vec::with_capacity(1 + 16 + 32 + data.len());
+                buf.push(self.enc_type());
+                buf.extend_from_slice(iv);
+                buf.extend_from_slice(mac);
+                buf.extend_from_slice(data);
+            }
+            _ => todo!(),
```
Hinton, line 178: Some tests?

crates/bitwarden/src/mobile/vault/client_sends.rs
```diff
@@ -52,6 +52,35 @@ impl<'a> ClientSends<'a> {
...
+        let data = std::fs::read(decrypted_file_path).unwrap();
+        let encrypted = self.encrypt_buffer(send, &data).await?;
+        std::fs::write(encrypted_file_path, encrypted)?;
+        Ok(())
+    }
+
+    pub async fn encrypt_buffer(&self, send: Send, buffer: &[u8]) -> Result<Vec<u8>> {
+        let enc = self.client.get_encryption_settings()?;
+        let key = Send::get_key(&send.key, enc, &None)?;
+        let enc = EncryptionSettings::new_single_key(key);
```
Hinton, line 79: Should `Send::get_key` return a `EncryptionSettings::new_single_key` immediately?

dani-garcia: I think the reason to have the `get_key` not return `EncryptionSettings` directly was to make it more testable, but that doesn't mean we can't have a helper function like this, because we do it in a couple of places: 
```rust
    pub(crate) fn get_encryption(
        key: &EncString,
        enc: &EncryptionSettings,
        org_id: &Option<Uuid>,
    ) -> Result<EncryptionSettings> {
        let key = Send::get_key(key, enc, org_id)?;
        Ok(EncryptionSettings::new_single_key(key))
    }
```

crates/bitwarden/src/vault/send.rs
```diff
@@ -135,6 +135,15 @@ impl Send {
         let key = derive_shareable_key(key.try_into().unwrap(), "send", Some("send"));
         Ok(key)
     }
```
Hinton, line 137: Can this be made private now? I think all calls to `Send::get_key` can safely be replaced by `get_encryption`.

dani-garcia: Good point, changed now

Hinton (approved):

PR #212, opened 2023-09-01
Make MAC validation optional when decrypting

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
Change the parameters of `decrypt_aes256` to make the MAC optional and not validate it, needed for implementing AES encryption without HMAC

### DIFF
--- a/crates/bitwarden/src/crypto/aes_ops.rs
+++ b/crates/bitwarden/src/crypto/aes_ops.rs
@@ -12,21 +12,22 @@ use crate::{
 
 pub fn decrypt_aes256(
     iv: &[u8; 16],
-    mac: &[u8; 32],
+    mac: Option<&[u8; 32]>,
     data: Vec<u8>,
     mac_key: Option<GenericArray<u8, U32>>,
     key: GenericArray<u8, U32>,
 ) -> Result<Vec<u8>> {
-    let mac_key = match mac_key {
-        Some(k) => k,
-        None => return Err(CryptoError::InvalidMac.into()),
+    match (mac_key, mac) {
+        (Some(mac_key), Some(mac)) => {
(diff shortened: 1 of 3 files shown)


crates/bitwarden/src/crypto/aes_ops.rs
```diff
@@ -12,22 +12,23 @@ use crate::{
 
 pub fn decrypt_aes256(
     iv: &[u8; 16],
-    mac: &[u8; 32],
+    mac: Option<&[u8; 32]>,
```
Hinton, line 15: It would be nice to ensure the mac and mac-key is always set or not set. Do you think we should have two methods for this instead? Or some way to wrap both mac and key in an option

dani-garcia: I've made it that way to try to change the function definition too much, but it's definitely better as two functions and it's not used in that many places

crates/bitwarden/src/crypto/aes_ops.rs
```diff
@@ -42,6 +25,21 @@ pub fn decrypt_aes256(
...
 }
 
+pub fn decrypt_aes256_hmac(
+    iv: &[u8; 16],
+    mac: &[u8; 32],
+    data: Vec<u8>,
+    mac_key: GenericArray<u8, U32>,
+    key: GenericArray<u8, U32>,
+) -> Result<Vec<u8>> {
+    // Validate HMAC
```
Hinton, line 35: This code is fairly self descriptive even without the comment.

Hinton (approved):

Hinton (approved):

PR #213, opened 2023-09-01
Implement parsing for all EncString ciphers

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
Implement converting an EncString to and from a base64 encoded string and a buffer for all the ciphers we support. Added support for legacy EncStrings without a type header, just like the rest of the clients.
Marked RSA with HMAC as deprecated like the other clients. Also removed their mac fields, the other clients don't check them either.


crates/bitwarden/src/crypto/enc_string.rs
```diff
@@ -62,52 +51,66 @@ impl FromStr for EncString {
...
-            ("5" | "6", 2) => {
-                unimplemented!()
+            #[allow(deprecated)]
+            ("5" | "6", 1) => {
+                let data = from_b64_vec(parts[0])?;
+                if enc_type == "5" {
+                    Ok(EncString::Rsa2048_OaepSha256_HmacSha256_B64 { data })
+                } else {
+                    Ok(EncString::Rsa2048_OaepSha1_HmacSha256_B64 { data })
+                }
```
Hinton, line 113: Is this cleaner than just writing

```rust
("5", 1) => {
  Ok(EncString::Rsa2048_OaepSha256_HmacSha256_B64 { data: from_b64_vec(parts[0])? })
}
("6", 1) => {
  Ok(EncString::Rsa2048_OaepSha1_HmacSha256_B64 { data: from_b64_vec(parts[0])? })
}
```

dani-garcia: I kept it like it was previoulsy, but it's definitely not very useful now that most of the match branches are very short with the `from_b64_vec`. We could even split `("1" | "2", 3)` as well, as it's only three lines now.

crates/bitwarden/src/crypto/enc_string.rs
```diff
@@ -148,6 +162,27 @@ impl EncString {
                     Ok(EncString::AesCbc256_HmacSha256_B64 { iv, mac, data })
                 }
             }
+            3 | 4 => {
```
Hinton, line 165: I guess we could split these as well.

Hinton (approved):

Hinton (approved):

PR #215, opened 2023-09-01
[PM-6107] Use rayon to multithread encryption/decryption

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
Use rayon to multithread the encryption and decryption of ciphers. Note that this will have no benefit for  WASM, as in that case it falls back to a single thread model.

In my Mac I get 8-9x speedups when decrypting a big number of EncStrings.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -439,6 +439,7 @@ dependencies = [
  "pbkdf2",
  "rand 0.8.5",
  "rand_chacha 0.3.1",
+ "rayon",
  "rsa",
  "schemars",
  "serde",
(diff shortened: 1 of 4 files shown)


dani-garcia:
Done some more benchmarking, both running the native binary, WASM, and the iOS emulator, and with two different types of data:
- One big array with 2 million EncStrings, each with 20 characters
- One smaller array of objects that implement Encryptable and each contain 14 EncStrings, the total size of the array is 2 million / 14 to keep the total EncStrings the same between both. The goal of this test is to check that nesting parallel calls works as expected.
```
{
    name: EncString,
    password: EncString,
    inner: [
        {
            name: EncString,
            password: EncString,
            inner: [
                {
                    name: EncString,
                    password: EncString,
                    inner: []
                },
                {
                    name: EncString,
                    password: EncString,
                    inner: []
                }
            ]
        },
        ... another one, same as above
    ]
}
```

On all the benchmarks both data types performed identically, which tells us that there is no overhead when nesting parallel iterator calls.

The results for the different targets are as follows:

NATIVE:
Secuential: 4s
Parallel: 430-470ms
Improvement: 8-9x

WASM:
Secuential: 16-19s
Parallel: 16-19s
No improvement, because rayon doesn't support WASM threads. Though we can look into: https://github.com/GoogleChromeLabs/wasm-bindgen-rayon

iOS:
Secuential: 4.6s
Parallel: 500-600ms
Improvement: 8-9x

This is run on a M2Pro Mac with 8 performance cores and 4 efficiency cores.

Hinton (approved):

PR #216, opened 2023-09-04
Scaffold exporters

Hinton:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Scaffold exporters.

## Before you submit


crates/bitwarden/src/tool/exporters/client_exporter.rs
```diff
@@ -0,0 +1,36 @@
...
+};
+
+pub struct ClientExporters<'a> {
+    pub(crate) _client: &'a crate::Client,
+}
+
+impl<'a> ClientExporters<'a> {
+    pub async fn export_vault(
+        &self,
+        folders: Vec<FolderView>,
```
dani-garcia, line 15: I'm assuming when using the export, the application won't have the list of folders and ciphers already decrypted, at most it will have the list view models. Would it make more sense to ask for the encrypted version and do it all in a single FFI call?

crates/bitwarden/src/tool/exporters/mod.rs
```diff
@@ -0,0 +1,33 @@
...
+};
+
+mod client_exporter;
+
+#[derive(JsonSchema)]
+#[cfg_attr(feature = "mobile", derive(uniffi::Enum))]
+pub enum ExportFormat {
+    Csv,
+    Json,
+    AccountEncryptedJson, // TODO: Should we deprecate this option completely?
```
dani-garcia, line 15: I don't think this format is particularly useful (and mostly replaced by the password encrypted JSON) and will just cause issues when users try to import it into a new account or after rotating keys, so I'm in support of deprecating it.

crates/bitwarden/src/tool/exporters/client_exporter.rs
Hinton: Yes, I had that on my mind but forgot it during the weekend.

PR #217, opened 2023-09-04
Add Sends API in bitwarden-uniffi

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
Add missing Sends API to the bitwarden-uniffi crate.

Also fixed copy-pasted comment in the password history function

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,5 +1,8 @@
 target
-languages
+languages/*
+!/languages/kotlin
+languages/kotlin/*
+!/languages/kotlin/doc.md
 schemas
 /crates/bitwarden-napi/src-ts/bitwarden_client/schemas.ts
 about.hbs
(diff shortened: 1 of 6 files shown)


Hinton (commented):
Can you re-generate the doc file?

dani-garcia:
Done, I also updated the prettierignore so that `npm run prettier` works on the file.

Not sure why but when I ran the doc generation the ExportFormat definition got deleted from the doc, so I restored it by hand.

.prettierignore
```diff
@@ -1,5 +1,8 @@
 target
-languages
+languages/*
+!/languages/kotlin
+languages/kotlin/*
```
Hinton, line 4: Is this needed?

dani-garcia: It's a bit awkward but that seems to be how the ignore file works yeah:
https://github.com/prettier/prettier/issues/3328#issuecomment-350732773

We could always remove it and add an extra step on the Readme to format the doc file after it's generated.

Hinton (approved):

PR #218, opened 2023-09-04
Golang SDK

cvele:
## Type of change

```
- [ ] Bug fix
- [ x] New feature development
- [ ] Other
```

## Objective

The objective is to develop a Golang wrapper for the Bitwarden client that interacts with a native rust library. This is to ensure that a consistent interface is provided, and the details of the Bitwarden rust library can be abstracted away.

## Code changes
BitwardenClient.go

    Defines the BitwardenClient struct with dependencies and submodules for handling projects and secrets.
    Implements the NewBitwardenClient function that creates a new client.
    Implements methods for access token login and cleanup.

BitwardenLibrary_native.go

    Implements the BitwardenLibrary interface using cgo for interacting with the native C library.
    Provides methods for initialization, memory release, and running commands on the client.

BitwardenLibrary_custom.go

    An alternative implementation of the BitwardenLibrary interface, possibly for testing or alternative backends. The code is currently the same as BitwardenLibrary_native.go.

CommandRunner.go

    Implements the CommandRunner struct that takes care of running commands via the library interface.
    Centralizes command-running logic.

Projects.go

    Implements a Projects struct with methods to create, list, get, update, and delete projects.
    Utilizes the CommandRunner interface to run these commands.

Secrets.go

    Implements a Secrets struct with methods to create, list, get, update, and delete secrets.
    Utilizes the CommandRunner interface to run these commands.

*Notes*

- The code is designed to be modular, separating concerns into different components (like Projects, Secrets, and CommandRunner).
- Interfaces are used where possible to make it easier to mock these components for testing.
- The native C API is encapsulated neatly, making it easier to switch out or update the underlying library without affecting the rest of the code.


Hinton (commented):
I think we should look into moving the c interface behind an `internal` package since it's not "safe" to consume. And only expose managed wrappers.

Do we need to define a proper package instead of `main`? Can golang packages be distributed as a subdirectory of this repo or do we need to host the code in it's own repository.

languages/go/main.go
```diff
@@ -0,0 +1,101 @@
...
+
+	if _, err = bitwardenClient.Secrets.Delete([]string{secretID}); err != nil {
+		panic(err)
+	}
+
+	if _, err = bitwardenClient.Projects.Delete([]string{projectID}); err != nil {
+		panic(err)
+	}
+
+	bitwardenClient.Close()
```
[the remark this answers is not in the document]
cvele: Will look into it 👍

languages/go/README.md
```diff
@@ -0,0 +1,116 @@
...
+This SDK is designed to interact with Bitwarden services in Go. It includes implementations for managing projects and secrets, as well as a client interface to facilitate operations like login.
+
+## Prerequisites
+
+- Go installed
+- C environment to run CGO
+
+## Installation
+
+Download the SDK files and place them in your Go project directory.
```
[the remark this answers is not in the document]
cvele: In the context of using cgo with Go, it's a common practice to include the C header files (with the .h extension) and any pre-compiled static or dynamic libraries (with .a or .so extensions) within the Go project directory. To link these libraries, you can use #cgo directives in the Go pseudo-code as shown below:

```/*
#cgo CFLAGS: -I/path/to/header/files
#cgo LDFLAGS: -L/path/to/library/files -lmylib
#include "mylib.h"
*/
import "C"
```

The specific location for these files within your Go project can vary depending on requirements. Our code looks for them under /usr/lib and /usr/local/lib. Location can be customized (look at build.sh for details). 

As for automating the build process, you could employ go generate to trigger shell commands that facilitate the build. For instance:

```import "os/exec"

func init() {
    // Check for the presence of the Rust library, and build if necessary
    if !checkRustLib() {
        cmd := exec.Command("/bin/bash", "build.sh")
        err := cmd.Run()
        if err != nil {
            panic("failed to build Rust library: " + err.Error())
        }
    }
}
```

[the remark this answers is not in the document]
cvele: @Hinton  We could be committing a pre-built version of the Rust library within the Go codebase. This is a viable approach for ensuring version compatibility. By doing this, we can tie specific versions of the Rust library to corresponding Git tags of the Go project. This offers a couple of advantages:

- Ease of Use: Users cloning our repository would have immediate access to the correct version of the Rust library, eliminating the need for manual installation or build steps.

- Version Synchronization: Keeping the Rust library version aligned with Go codebase ensures compatibility and reduces the risk of version mismatch issues.

However, there are also some considerations to keep in mind:

- Repo Size: The size of the repository may increase, especially if the Rust library grows large or if we need to keep multiple versions.

- Platform Compatibility: Pre-compiled libraries are often platform-specific. We would either need to commit versions for all targeted platforms or provide a build mechanism for platforms we haven't pre-compiled for.

- Maintenance: Keeping the library updated in the repo would be done via CI/CD, and we would have to ensure it remains in sync with the Go codebase.

If these considerations are acceptable within the context, then including the pre-built Rust library could be a good strategy.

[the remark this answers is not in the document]
Hinton: I don't think we'll want to include the built libraries. They are ~5mb each (x6) which would balloon the repository size. Including the c header would probably be fine. We could also look into using a different repository for golang down the road, and bundle the libs using git lfs.

languages/go/main.go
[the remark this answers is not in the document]
cvele: In Go, resource management is generally explicit and usually left to the programmer to handle. Unlike languages with garbage collection for resources (like C# with its IDisposable interface and using statement), Go doesn't offer built-in destructors or "finalizers" for custom types that would automatically release resources.

That said we could create a wrapper
 function that takes a callback to operate on the client and closes it afterwards.
 
 ```func WithBitwardenClient(action func(client *BitwardenClient) error) error {
    client := NewBitwardenClient()
    defer client.Close()
    return action(client)
}
```

How does that sound ?

**Comment by cvele** [comment]
> I think we should look into moving the c interface behind an `internal` package since it's not "safe" to consume. And only expose managed wrappers.

👍 

> Do we need to define a proper package instead of `main`? Can golang packages be distributed as a subdirectory of this repo or do we need to host the code in it's own repository.

You don't necessarily need a separate repository for distributing your Go package. You can have it as a subdirectory in your existing repository. In your case, with the configuration as module github.com/bitwarden/sdk/languages/go, the package can reside as a subdirectory under this path.

Consumers of your package would import it like this:

File: languages/go/main.go (hunk 1 above)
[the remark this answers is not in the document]
Comment (Hinton) [reply languages/go/main.go]: In that case I think keeping the explicit Close makes sense. We should ensure all our examples include the `defer Close()`.

File: languages/go/project.go [hunk 3]
```diff
@@ -0,0 +1,118 @@
...
+
+type Projects struct {
+	CommandRunner CommandRunnerInterface
+}
+
+func NewProjects(commandRunner CommandRunnerInterface) *Projects {
+	return &Projects{CommandRunner: commandRunner}
+}
+
+func (p *Projects) Get(id string) (ResponseForProjectResponse, error) {
```
Hinton, line 24: Similar to the .net SDK, can we unwrap the responses and return the error in the library? It's somewhat annoying if consumers needs to handle two error types.

dani-garcia:
Some small comments:

- the names of the files in `internal/cinterface` are mispelled: `libary` -> `library`.
- with the move to `internal/cinterface`, the `build.sh` script needs to update it's path now, at the moment it's expecting `bitwarden_library.go` in the same directory.
- I think we should consider moving the example to a separate `example` directory to avoid mixing it with the library code. We are doing that as well in the .NET implementation. I've done that and made the example into a proper go module so I could test the changes locally (here's the commit 0eb862f50a24f85073f63a3daf71582a3004c656), what do you think about doing something like that? . In `ClientSettings` I had to add `&` to the arguments otherwise it wouldn't compile, not sure what's up with that.

Otherwise the code worked correctly in my testing and I see you're already working on the CI release action, great!

dirtycajunrice:
Can we get a condensed list of what is still blocking merge? The inclusion of this SDK is the only thing blocking bitwarden's inclusion in external secrets for k8s https://github.com/external-secrets/external-secrets/issues/2661

cvele:
> Some small comments:
> 
>     * the names of the files in `internal/cinterface` are mispelled: `libary` -> `library`.
> 
>     * with the move to `internal/cinterface`, the `build.sh` script needs to update it's path now, at the moment it's expecting `bitwarden_library.go` in the same directory.
> 
>     * I think we should consider moving the example to a separate `example` directory to avoid mixing it with the library code. We are doing that as well in the .NET implementation. I've done that and made the example into a proper go module so I could test the changes locally (here's the commit [0eb862f](https://github.com/bitwarden/sdk/commit/0eb862f50a24f85073f63a3daf71582a3004c656)), what do you think about doing something like that? . In `ClientSettings` I had to add `&` to the arguments otherwise it wouldn't compile, not sure what's up with that.
> 
> 
> Otherwise the code worked correctly in my testing and I see you're already working on the CI release action, great!

Sorry for a delay, but all issues have been addressed. Additionally removed hard dependancy on clientsettings from schemas.

cvele:
> Can we get a condensed list of what is still blocking merge? The inclusion of this SDK is the only thing blocking bitwarden's inclusion in external secrets for k8s [external-secrets/external-secrets#2661](https://github.com/external-secrets/external-secrets/issues/2661)

All of the comments that I have found have been resolved. Could you re-review?

support/scripts/schemas.ts
```diff
@@ -63,6 +63,18 @@ async function main() {
...
   writeToFile("./languages/csharp/Bitwarden.Sdk/schemas.cs", csharp.lines);
+  
+  writeToFile("./languages/csharp/schemas.cs", csharp.lines);
+
+  const go = await quicktype({
+    inputData,
+    lang: "go",
+    rendererOptions: {
+      package: "sdk",
+    },
```
Hinton, line 74: It seems we're only using the types and not the unmarshal/marshal helpers?

If so we should change this to only generate types and packages https://github.com/glideapps/quicktype/blob/master/packages/quicktype-core/src/language/Golang.ts#L25

languages/go/example/example.go
```diff
@@ -0,0 +1,85 @@
...
+
+	responseForAPIKeyLoginResponse := bitwardenClient.AccessTokenLogin(accessToken)
+	fmt.Println(responseForAPIKeyLoginResponse)
+
+	organizationID, err := uuid.FromString(organizationIDStr)
+	if err != nil {
+		panic(err)
+	}
+
+	projectResponse, err := bitwardenClient.Projects.Create(organizationID.String(), projectName)
```
Hinton, line 34: The variable names in the example should be updated to reflect that the responses are models not response models.

languages/go/internal/cinterface/bitwarden_library_custom.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"fmt"
+	"unsafe"
+)
+
+/*
+#cgo LDFLAGS: -lbitwarden_c
+#cgo linux LDFLAGS: -L/usr/local/lib -L/usr/lib
+#cgo darwin LDFLAGS: -L/usr/local/lib -L/usr/lib
```
Hinton, line 14: This seems to be the only difference for custom? Is there a reason to not always attempt to find the binary in a local`./lib` directory?

languages/go/example/example.go
bbujak: done

support/scripts/schemas.ts
bbujak: Added

languages/go/bitwarden_client.go
```diff
@@ -0,0 +1,63 @@
...
+	Projects      ProjectsInterface
+	Secrets       SecretsInterface
+}
+
+func NewBitwardenClient(apiURL *string, identityURL *string, userAgent *string) *BitwardenClient {
+	deviceType := DeviceType("SDK")
+	clientSettings := ClientSettings{
+		APIURL:      apiURL,
+		IdentityURL: identityURL,
+		UserAgent:   userAgent,
```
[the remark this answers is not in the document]
bbujak: done

languages/go/bitwarden_client.go
```diff
@@ -0,0 +1,63 @@
...
+func (c *BitwardenClient) AccessTokenLogin(accessToken string) ResponseForAPIKeyLoginResponse {
+	req := AccessTokenLoginRequest{AccessToken: accessToken}
+	command := Command{AccessTokenLogin: &req}
+
+	responseStr := c.commandRunner.RunCommand(command)
+
+	var response ResponseForAPIKeyLoginResponse
+	if err := json.Unmarshal([]byte(responseStr), &response); err != nil {
+		panic(err)
+	}
```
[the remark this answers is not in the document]
bbujak: Agreed , done

languages/go/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+To initialize the client, you need to import the SDK and create a new `BitwardenClient` instance.
+
+```go
+import "github.com/bitwarden/sdk/languages/go"
+
+settings := ClientSettings{
+  // Your settings here
+}
+client := NewBitwardenClient(settings)
```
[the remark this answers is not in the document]
bbujak: done

languages/go/README.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+---
+
+### Projects
+
+#### Create a Project
+
+```go
+response, err := client.Projects.Create("organization_id", "project_name")
```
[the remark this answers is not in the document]
bbujak: done

languages/csharp/schemas.cs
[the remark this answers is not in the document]
bbujak: deleted

support/scripts/schemas.ts
```diff
@@ -62,6 +62,15 @@ async function main() {
...
 
   writeToFile("./languages/csharp/schemas.cs", csharp.lines);
+
+  const go = await quicktype({
+    inputData,
+    lang: "go",
+    rendererOptions: {
+      "package": "main",
+    },
+  });
```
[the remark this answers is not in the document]
bbujak: done

languages/go/README.md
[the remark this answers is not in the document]
Hinton: `settings` is defined above and should be removed since it's not used.

languages/go/bitwarden_client.go
```diff
@@ -0,0 +1,67 @@
...
+	lib           cinterface.BitwardenLibrary
+	commandRunner CommandRunnerInterface
+	Projects      ProjectsInterface
+	Secrets       SecretsInterface
+}
+
+func NewBitwardenClient(apiURL *string, identityURL *string) (*BitwardenClient, error) {
+	deviceType := DeviceType("SDK")
+	var userAgent = "GOLANG-SDK"
+	clientSettings := ClientSettings{
```
Hinton, line 20: Please be consistent in variable initialization.
```suggestion
	deviceType := DeviceType("SDK")
	userAgent := "GOLANG-SDK"
	
	clientSettings := ClientSettings{
```

languages/go/bitwarden_client.go
```diff
@@ -0,0 +1,67 @@
...
+	return &BitwardenClient{
+		lib:           lib,
+		client:        client,
+		commandRunner: runner,
+		Projects:      NewProjects(runner),
+		Secrets:       NewSecrets(runner),
+	}, nil
+}
+
+func (c *BitwardenClient) AccessTokenLogin(accessToken string) (*ResponseForAPIKeyLoginResponse, error) {
```
Hinton, line 48: We should change this to only return an error if one exists, since the consumer has no use for the response.

```suggestion
func (c *BitwardenClient) AccessTokenLogin(accessToken string) (error) {
```

This also means checking that the response is successful. Similar to the .NET implementation, https://github.com/bitwarden/sdk/blob/master/languages/csharp/Bitwarden.Sdk/BitwardenClient.cs#L23-L31

languages/go/README.md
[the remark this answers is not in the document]
bbujak: removed

languages/go/project.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	var response ProjectResponse
+	if err := p.executeCommand(command, &response); err != nil {
+		return nil, err
+	}
+	return &response, nil
+}
+
+func (p *Projects) Delete(projectIDs []string) (*ProjectResponse, error) {
```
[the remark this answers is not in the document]
bbujak: replaced with ProjectsDeleteResponse

languages/go/bitwarden_client.go
bbujak: Made it consistent, changed userAgent to "Bitwarden GOLANG-SDK"

languages/go/bitwarden_client.go
bbujak: Implemented to return just error

languages/go/util.go
```diff
@@ -0,0 +1,32 @@
...
+		return fmt.Errorf("API error: unknown")
+	}
+
+	err = json.Unmarshal([]byte(responseStr), &v)
+	if err != nil {
+		return fmt.Errorf("failed to unmarshal response: %v", err)
+	}
+
+	return nil
+}
```
dani-garcia, line 32: I don't think this is correct, the responses we get from the API are of the format `{ success, errorMessage, data }`. We're handling the success and errorMessage correctly, but once that's done we try to decode the whole message into `&v`, while we should only be decoding the data field, something like this:
```go
func checkSuccessAndError(responseStr string, v interface{}) error {
	var wrapper struct {
		Success      bool             `json:"success"`
		ErrorMessage *string          `json:"errorMessage"`
		Data         *json.RawMessage `json:"data"`
	}

	err := json.Unmarshal([]byte(responseStr), &wrapper)
	if err != nil {
		return fmt.Errorf("failed to unmarshal wrapper response: %v", err)
	}

	if !wrapper.Success {
		if wrapper.ErrorMessage != nil {
			return fmt.Errorf("API error: %s", *wrapper.ErrorMessage)
		}
		return fmt.Errorf("API error: unknown")
	}

	err = json.Unmarshal(*wrapper.Data, &v)
	if err != nil {
		return fmt.Errorf("failed to unmarshal response: %v", err)
	}

	return nil
}
```

languages/go/bitwarden_client.go
```diff
@@ -0,0 +1,67 @@
...
+	if err != nil {
+		return err
+	}
+
+	var response ResponseForAPIKeyLoginResponse
+	if err := json.Unmarshal([]byte(responseStr), &response); err != nil {
+		return err
+	}
+
+	return nil
```
dani-garcia, line 62: We need to do checkSuccessAndError here as well:
```go
func (c *BitwardenClient) AccessTokenLogin(accessToken string) (*ResponseForAPIKeyLoginResponse, error) {
	req := AccessTokenLoginRequest{AccessToken: accessToken}
	command := Command{AccessTokenLogin: &req}

	responseStr, err := c.commandRunner.RunCommand(command)
	if err != nil {
		return nil, err
	}

	var response ResponseForAPIKeyLoginResponse
	if err = checkSuccessAndError(responseStr, &response); err != nil {
		return nil, err
	}

	return &response, nil
}
```

bbujak: do we want response retuned ?

dani-garcia: Good point, I didn't see that was changed to not return a response.

We only need to return an error yeah, but at least we should check that success is true in the response like we do with checkSuccessAndError, maybe like this?

```go
func (c *BitwardenClient) AccessTokenLogin(accessToken string) error {
	req := AccessTokenLoginRequest{AccessToken: accessToken}
	command := Command{AccessTokenLogin: &req}

	responseStr, err := c.commandRunner.RunCommand(command)
	if err != nil {
		return err
	}

	var response APIKeyLoginResponse
	if err = checkSuccessAndError(responseStr, &response); err != nil {
		return err
	}

	return nil
}
```

bbujak: APIKeyLoginResponse or ResponseForAPIKeyLoginResponse? I would say ResponseForAPIKeyLoginResponse

languages/go/util.go
bbujak: This , looks correct. I made the changes

languages/go/bitwarden_client.go
bbujak: Changed

bbujak: changed to APIKeyLoginResponse

go.work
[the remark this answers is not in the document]
Hinton: The `go.work`still exists.

.github/workflows/golang-release.yml
```diff
@@ -0,0 +1,69 @@
...
+jobs:
+  build_rust:
+    uses: ./.github/workflows/build-rust-cross-platform.yml
+
+  generate-schemas:
+    uses: ./.github/workflows/generate_schemas.yml
+
+  build:
+    needs: [build_rust, generate-schemas]
+    runs-on: ubuntu-latest
```
mimartin12, line 23:
```suggestion
    runs-on: ubuntu-22.04
```
We typically pin runner versions to a specific version, Ubuntu 22.04 can be used here.

.github/workflows/golang-release.yml
```diff
@@ -0,0 +1,69 @@
...
+
+      - name: Build
+        run: go build -v ./...
+
+      - name: Test
+        run: go test -v ./...
+
+  release:
+    needs: [build]
+    runs-on: ubuntu-latest
```
mimartin12, line 49:
```suggestion
    runs-on: ubuntu-22.04
```
Same as above comment.

.github/workflows/golang-release.yml
```diff
@@ -0,0 +1,69 @@
...
+        with:
+          go-version: ${{ env.GO_VERSION }}
+
+      - name: Cache dependencies
+        uses: actions/cache@v2
+        with:
+          path: ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
```
mimartin12, line 39: I am not familiar with using the `restore-keys` with this action, will this name be sufficient? Or is there a glob pattern expected here? Ex. `${{ runner.os }}-go-*`?

.github/workflows/golang-release.yml
vgrassia, line 23:
```suggestion
  build:
    name: Build
    needs:
      - build_rust
      - generate-schemas
    runs-on: ubuntu-22.04
```

.github/workflows/golang-release.yml
```diff
@@ -0,0 +1,69 @@
...
+
+  generate-schemas:
+    uses: ./.github/workflows/generate_schemas.yml
+
+  build:
+    needs: [build_rust, generate-schemas]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
```
vgrassia, line 26:
```suggestion
        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
```

.github/workflows/golang-release.yml
```diff
@@ -0,0 +1,69 @@
...
+
+      - name: Setup Go environment
+        uses: actions/setup-go@v2
```
vgrassia, line 29:
```suggestion
        uses: actions/setup-go@93397bea11091df50f3d7e59dc26a7711a8bcfbe # v4.1.0
```

.github/workflows/golang-release.yml
```diff
@@ -0,0 +1,69 @@
...
+        with:
+          go-version: ${{ env.GO_VERSION }}
+
+      - name: Cache dependencies
+        uses: actions/cache@v2
```
vgrassia, line 34:
```suggestion
        uses: actions/cache@704facf57e6136b1bc63b828d79edcd491f0ee84 # v3.3.2
```

.github/workflows/golang-release.yml
mimartin12, line 1: Many of these actions are very old and outdated. `checkout` and `setup-go` for example are shipping v4 now. These need to be updated.

Actions are required to be pinned to their tag commit that was released for a specific version for that action. For example, with the Checkout Action.

`actions/checkout@v4` should actually be `actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11` ([source](https://github.com/actions/checkout/releases/tag/v4.1.1))

.github/workflows/golang-release.yml
vgrassia, line 49:
```suggestion
  release:
    name: Release
    needs: build
    runs-on: ubuntu-22.04
```

mimartin12 (changes requested):

.github/workflows/golang-release.yml
```diff
@@ -0,0 +1,69 @@
...
+
+      - name: Test
+        run: go test -v ./...
+
+  release:
+    needs: [build]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
```
vgrassia, line 52:
```suggestion
        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
```

.github/workflows/golang-release.yml
```diff
@@ -0,0 +1,69 @@
...
+
+      - name: Setup Go environment
+        uses: actions/setup-go@v2
```
vgrassia, line 55:
```suggestion
        uses: actions/setup-go@93397bea11091df50f3d7e59dc26a7711a8bcfbe # v4.1.0
```

.github/workflows/golang-release.yml
```diff
@@ -0,0 +1,69 @@
...
+        with:
+          go-version: ${{ env.GO_VERSION }}
+
+      - name: Set release version
+        run: echo "VERSION=${{ github.event.inputs.version_number }}" >> $GITHUB_ENV
+
+      - name: Install Goreleaser
+        run: go install github.com/goreleaser/goreleaser@latest
```
vgrassia, line 63:
```suggestion
        run: go install github.com/goreleaser/goreleaser@v1.21.2
```

vgrassia (changes requested):

go.work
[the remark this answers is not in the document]
bbujak: deleted

.github/workflows/golang-release.yml
bbujak: changed

.github/workflows/golang-release.yml
bbujak: changed

.github/workflows/golang-release.yml
bbujak: changed

.github/workflows/golang-release.yml
bbujak: changed

.github/workflows/golang-release.yml
bbujak: changed

.github/workflows/golang-release.yml
bbujak: changed

.github/workflows/golang-release.yml
bbujak: changed

.github/workflows/golang-release.yml
bbujak: changed

.github/workflows/golang-release.yml
bbujak: changed

.github/workflows/golang-release.yml
bbujak: changed

languages/go/internal/cinterface/bitwarden_library_custom.go
dani-garcia: Personally I've always used the `./lib` directory to copy the `libbitwarden_c.dylib` file and then ran `go build`, so I've never used either the `bitwarden_library_custom.go` or the `build.sh` script, maybe they are not needed anymore?

.github/workflows/golang-release.yml
bbujak: This should work, if you look at [this](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key)

.github/workflows/golang-release.yml
bbujak: fixed

languages/go/internal/cinterface/bitwarden_library_custom.go
bbujak: I agrree, let's remove it

dani-garcia (dismissed):
Code looks good to me

languages/go/internal/cinterface/bitwarden_library.go
```diff
@@ -0,0 +1,58 @@
+//go:build !custom
+// +build !custom
+
```
Hinton, line 3: With the custom file gone, can we remove these comments?

Hinton (commented):
Minor nit, and please run prettier on the readme. `npm run prettier` in the project root.

languages/go/internal/cinterface/bitwarden_library.go
bbujak: Removed

dani-garcia:
The languages directories are excluded from prettier in `.prettierignore`, it seems. To format the README you can instead do:
```
cd languages/go
npx prettier --write README.md
```

We'll need to fix the `.prettierignore` in the future.

bbujak:
> The languages directories are excluded from prettier in `.prettierignore`, it seems. To format the README you can instead do:
> 
> ```
> cd languages/go
> npx prettier --write README.md
> ```
> 
> We'll need to fix the `.prettierignore` in the future.

That did the trick

dani-garcia (dismissed):

Hinton (approved):

vgrassia (approved):

dani-garcia (approved):

larivierec:
hello I know the PR is already closed but I have a small question.
Is the go package going to include the rust generated binary?

i see a mention of `libbitwarden_c`.
this is the main reason i'm asking

Hinton:
@larivierec Our plan is for go users to download the c-lib manually since including it in the repository would be unmaintainable and quickly balloon the repository size.

carnei-ro:
@cvele hey man, I'm having a bad time trying to compile the example project located at `./languages/go/example`

```
# github.com/bitwarden/sdk/languages/go
../command_runner.go:10:21: undefined: Command
../project.go:4:47: undefined: ProjectResponse
../project.go:5:32: undefined: ProjectsResponse
../project.go:6:26: undefined: ProjectResponse
../project.go:7:65: undefined: ProjectResponse
../project.go:8:32: undefined: ProjectsDeleteResponse
../secrets.go:4:80: undefined: SecretResponse
../secrets.go:5:32: undefined: SecretIdentifiersResponse
../secrets.go:6:25: undefined: SecretResponse
../secrets.go:7:97: undefined: SecretResponse
../secrets.go:7:97: too many errors
```

any tips?

Hinton:
@carnei-ro you need to run `npm run schemas` in the repository root to generate the schema definitions.

Skarlso:
I was trying to do this too and built the whole thing and did the vpr thing, but I'm getting 

```
ld: symbol(s) not found for architecture x86_64
```

I'm new to cgo, so I'm not sure what's up. Any ideas anyone? I know this is a closed PR, but I tried the rest of the places, and there are no answers, guides, docs, anything on how to deal with the SDK.

Skarlso:
ah.... finally, success

```
export CGO_LDFLAGS="-framework CoreFoundation"
➜  example git:(main) ✗ go build example.go
➜  example git:(main) ✗ ls -l
total 22792
-rwxr-xr-x@ 1 skarlso  staff  11653184 Jun 12 13:31 example
-rw-r--r--@ 1 skarlso  staff      2802 Jun 12 12:25 example.go
-rw-r--r--@ 1 skarlso  staff       188 Jun 12 12:25 go.mod
-rw-r--r--@ 1 skarlso  staff       187 Jun 12 12:25 go.sum
➜  example git:(main) ✗ ./example
panic: API error: Access token is not in a valid format: Doesn't contain a decryption key
... don't care about this, the binary works.
```

Skarlso:
In fact, there was no need for the `cp -vpr`. Just simply `export CGO_LDFLAGS="-framework CoreFoundation"` and the thing built out of the box.

PR #222, opened 2023-09-08
Asymmetric encryption support for EncString

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
EncString now has a generic type param to indicate whether it's using Symm/Asymm encryption. `EncString` is now the equivalent of `BaseEncString<SymmetricKey>` and `AsymmetricEncString` is `BaseEncString<AsymmetricKey>`. 

Note that both `EncString` and `AsymmetricEncString` are specific versions of the `BaseEncString`, which mean they both contain all the 7 variants we support, this is just to avoid code duplication while implementing parsing, but some of that might be able to be moved to the generic type.

### DIFF
--- a/crates/bitwarden/src/client/client.rs
+++ b/crates/bitwarden/src/client/client.rs
@@ -11,7 +11,7 @@ use {
             TwoFactorEmailRequest,
         },
         client::auth_settings::AuthSettings,
-        crypto::EncString,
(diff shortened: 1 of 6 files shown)


dani-garcia:
Replaced by #448

PR #229, opened 2023-09-12
SM-874: Fix Python Integration

coltonhurst:
## Type of change

```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

The Python integration is out of date and is not currently working. This PR fixes the Python integration by adding the ability to log in via access token, which is currently the supported way to interact with the Bitwarden Secrets Manager SDK. This PR also adds a `ProjectsClient` to `bitwarden_client.py` for easy use, and updates the error handling on Project or Secret deletes in the SDK.

## Code changes

- **crates/sdk-schemas/src/main.rs:** We need the `AccessTokenLoginResponse` struct to be generated so access token auth is available to integrations
- **languages/python/BitwardenClient/bitwarden_client.py:**
  - Import the following from `.schemas`:
    - `AccessTokenLoginRequest`
    - `AccessTokenLoginResponse`
    - `ResponseForAccessTokenLoginResponse`
  - Define a new function `access_token_login` to support authenticating with an access token
  - Add the `project_ids` parameter to the `create` and `update` functions for `SecretsClient`
  - Add a `ProjectsClient` class and `projects()` to the `BitwardenClient`
  - Removed the password login methods, as they are not supported
  - Removed the imports that are not needed, and add one that are (like `sys`, etc.)
- **languages/python/README.md:** Update the readme instructions, give more examples
- **languages/python/login.py:** Renamed `login.py` to `example.py`
- **languages/python/example.py:** Update the example to showcase auth with an access token

## Before you submit


languages/python/login.py
```diff
@@ -4,21 +4,18 @@
...
+    "identityUrl": "http://localhost:33656",
+    "userAgent": "Python",
 }))
 
 logging.basicConfig(level=logging.DEBUG)
 
-result = client.password_login("test@bitwarden.com", "asdfasdf")
-print(result)
-print(client.get_user_api_key("asdfasdf"))
+result = client.access_token_login("access token here")
```
Thomas-Avery, line 16:
```
if result.success == False:
    sys.exit(result.error_message)
```

I would recommend we add some error handling for failed auth.

Thomas-Avery (commented):
Looking good. Couple of minor suggestions.

languages/python/login.py
```diff
@@ -4,21 +4,18 @@
 from BitwardenClient.schemas import client_settings_from_dict
 
 client = BitwardenClient(client_settings_from_dict({
-    "api_url": "http://localhost:4000",
-    "identity_url": "http://localhost:33656",
-    "user_agent": "Python",
+    "apiUrl": "http://localhost:4000",
+    "deviceType": "SDK",
```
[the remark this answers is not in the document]
coltonhurst: Updated 👍

Thomas-Avery (approved):
LGTM

languages/python/BitwardenClient/bitwarden_client.py
```diff
@@ -18,6 +18,12 @@ def password_login(self, email: str, password: str) -> ResponseForPasswordLoginR
         )
         return ResponseForPasswordLoginResponse.from_dict(result)
```
Hinton, line 19: Can you remove this method while at it. We don't support password login.

languages/python/login.py
Hinton, line 1: Should rename this to example or something.

coltonhurst: Renamed to `example.py` in [4212204](https://github.com/bitwarden/sdk/pull/229/commits/42122044c1c2731c561bc95a62d2a7f2cb67543b)

languages/python/BitwardenClient/bitwarden_client.py
coltonhurst: Removed those three functions, and the imports that were no longer being used in [4212204](https://github.com/bitwarden/sdk/pull/229/commits/42122044c1c2731c561bc95a62d2a7f2cb67543b)

Thomas-Avery (approved):

Hinton (approved):

crates/bitwarden/src/secrets_manager/projects/delete.rs
Hinton, line 1: Please introduce this in a separate PR. We'll want to update the c# bindings.

coltonhurst: Moved the changes to: https://github.com/bitwarden/sdk/pull/304

Thomas-Avery (approved):

languages/python/BitwardenClient/bitwarden_client.py
```diff
@@ -12,33 +12,22 @@ def __init__(self, settings: ClientSettings = None):
...
             self.inner = bitwarden_py.BitwardenClient(settings_json)
 
-    def password_login(self, email: str, password: str) -> ResponseForPasswordLoginResponse:
-        result = self._run_command(
-            Command(password_login=PasswordLoginRequest(email, password))
-        )
-        return ResponseForPasswordLoginResponse.from_dict(result)
-
-    def get_user_api_key(self, secret: str, is_otp: bool = False) -> ResponseForUserAPIKeyResponse:
+    def access_token_login(self, access_token: str) -> AccessTokenLoginResponse:
```
dani-garcia, line 15: We're not returning anything on `access_token_login` on other language integrations, beyond an error type in the languages that require it.

This is because the return values are not really applicable to secrets manager:
```rust
pub struct AccessTokenLoginResponse {
    pub authenticated: bool,
    /// TODO: What does this do?
    pub reset_master_password: bool,
    /// Whether or not the user is required to update their master password
    pub force_password_reset: bool,
    two_factor: Option<TwoFactorProviders>,
}
```

languages/python/example.py
```diff
@@ -0,0 +1,47 @@
...
+}))
+
+# Add some logging & set the org id
+logging.basicConfig(level=logging.DEBUG)
+organization_id = "org_id_here"
+
+# Attempt to authenticate with the Secrets Manager Access Token
+result = client.access_token_login("access_token_here")
+
+if result.success == False:
```
dani-garcia, line 22: This check should be applied to all commands, I think the best way would be to update `BitwardenClient`:

```py
    def _run_command(self, command: Command) -> Any:
        response_json = self.inner.run_command(json.dumps(command.to_dict()))
        response = json.loads(response_json)

        if response["success"] == False:
            raise Exception(response["errorMessage"])

        return response
```

That way it won't be needed to be handled by clients directly, they'll just get an exception.

dani-garcia (commented):
Two small comments, otherwise I'd say it looks good

dani-garcia (dismissed):

dani-garcia (approved):

PR #230, opened 2023-09-13
Use the same urls for debug and release

Hinton:
## Type of change

```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Use the same urls for debug and release to avoid confusion.

## Before you submit


dani-garcia (approved):
I agree that having separate configs for debug and release is confusing, i've tripped on it a couple of times

PR #233, opened 2023-09-13
Offer Maven package at SDK repository

withinfocus:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Publishes the Maven artifact to the SDK repository vs. the spinoff private one just for that purpose.

## Code changes

- **languages/kotlin/sdk/build.gradle:** Path change.

## Before you submit

### DIFF
--- a/.github/workflows/build-android.yml
+++ b/.github/workflows/build-android.yml
@@ -13,6 +13,21 @@ defaults:
     shell: bash
 
 jobs:
+  setup:
+    name: Setup
+    runs-on: ubuntu-22.04
+    outputs:
+      package_version: ${{ steps.retrieve-version.outputs.package_version }}
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac # v4.0.0
+
+      - name: Get Package Version
+        id: retrieve-version
+        run: |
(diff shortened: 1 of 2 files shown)


Hinton:
We need to do a few more changes for this to work.

- Rename sdk to contain mobile or kotlin.
- Remove PAT
- Ideally update snapshot naming to be `last-version-snapshot`. since snapshots keep all old versions resulting in large sizes. We'll want a way to delete older snapshots.
- Configure artifacts cleanup for non tagged versions.

withinfocus:
> We need to do a few more changes for this to work.
> 
> * Rename sdk to contain mobile or kotlin. Since it's going to collide with the java version.

Done.

> * Remove PAT

Not sure why we'd remove the token access as this is needed to publish to Packages.

> * Ideally update snapshot naming to be `last-version-snapshot`. since snapshots keep all old versions resulting in large sizes. We'll want a way to delete older snapshots.

Pushed a commit that uses non-snapshot versioning for `master` pushes, otherwise use the version and branch name with a snapshot otherwise. Is this what you're wanting?

> * Configure artifacts cleanup for non tagged versions.

Do we want to automate this across the whole repo?

withinfocus:
https://github.com/bitwarden/sdk/packages/1945815

Hinton:
Replaced with https://github.com/bitwarden/sdk/pull/235

withinfocus:
You can always make commits directly to this branch vs. new branch and PR creation.

PR #235, opened 2023-09-14
Change maven artifactId to sdk-android

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Code changes

## Before you submit

### DIFF
--- a/.github/workflows/build-android.yml
+++ b/.github/workflows/build-android.yml
@@ -101,23 +101,10 @@ jobs:
         working-directory: languages/kotlin
         run: ./build-schemas.sh
 
-      - name: Login to Azure - CI Subscription
-        uses: Azure/login@92a5484dfaf04ca78a94597f4f19fea633851fa2 # v1.4.7
-        with:
-          creds: ${{ secrets.AZURE_KV_CI_SERVICE_PRINCIPAL }}
-
-      - name: Retrieve bot secrets
-        id: retrieve-bot-secrets
-        uses: bitwarden/gh-actions/get-keyvault-secrets@8fccdaea2810d5dfb13753b6dd93bbd3a63ff82e
-        with:
-          keyvault: bitwarden-ci
-          secrets: "github-pat-bitwarden-devops-bot-packages-scope"
-
       - name: Publish
         uses: gradle/gradle-build-action@ef76a971e2fa3f867b617efd72f2fbd72cf6f8bc # v2.8.0
         with:
           arguments: sdk:publish
           build-root-directory: languages/kotlin
         env:
-          GITHUB_ACTOR: "bitwarden-devops-bot"
-          GITHUB_TOKEN: ${{ steps.retrieve-bot-secrets.outputs.github-pat-bitwarden-devops-bot-packages-scope }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/languages/kotlin/app/build.gradle
+++ b/languages/kotlin/app/build.gradle
@@ -48,7 +48,7 @@ android {
 }
 
 dependencies {
-    // implementation 'com.bitwarden.sdk'
+    // implementation 'com.bitwarden.sdk-android'
     implementation project(':sdk')
 
     implementation 'androidx.core:core-ktx:1.8.0'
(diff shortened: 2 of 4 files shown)


.github/workflows/build-android.yml
```diff
@@ -101,23 +101,10 @@ jobs:
...
-
       - name: Publish
         uses: gradle/gradle-build-action@ef76a971e2fa3f867b617efd72f2fbd72cf6f8bc # v2.8.0
         with:
           arguments: sdk:publish
           build-root-directory: languages/kotlin
         env:
-          GITHUB_ACTOR: "bitwarden-devops-bot"
-          GITHUB_TOKEN: ${{ steps.retrieve-bot-secrets.outputs.github-pat-bitwarden-devops-bot-packages-scope }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
withinfocus, line 110: ℹ️ OK, so you wanted to use the local token.

languages/kotlin/app/build.gradle
```diff
@@ -48,7 +48,7 @@ android {
 }
 
 dependencies {
-    // implementation 'com.bitwarden.sdk'
+    // implementation 'com.bitwarden.sdk-android'
```
withinfocus, line 51: ⚠️ This is in the `kotlin` directory which is why I named my PR that way, but if `android` makes sense then fine by me.

languages/kotlin/sdk/build.gradle
```diff
@@ -42,14 +41,25 @@ publishing {
...
         maven(MavenPublication) {
             groupId = 'com.bitwarden'
-            artifactId = 'sdk'
+            artifactId = 'sdk-android'
 
-            def details = versionDetails()
+            // Determine the version from the git history.
+            //
+            // PRs: use the branch name.
+            // Master: Grab it from `crates/bitwarden/Cargo.toml`
```
withinfocus, line 49: 🎨 I'd like to discuss the approach on my PR to use non-snapshots for `master` builds and to also include the version before the branch name otherwise. These seem like best practices with Maven versioning. Do you want to use another workflow for non-snapshot builds?

languages/kotlin/app/build.gradle
Hinton: I think it does, the artifact is android specific. We should probably consider renaming the directory.

languages/kotlin/sdk/build.gradle
withinfocus: Oh also, when we entertain more automated versioning (perhaps not in this repo, but it's happening elsewhere) this could very much be driven inside the Actions workflows so that's why I copied over what the Rust builds were doing. Using that and passing in the version as a step dependency seems like the long-term solution to me vs. individual implementations figuring it out.

Hinton: Everything except tagged versions should be snapshots right? A non snapshot version should be static and never modified.

withinfocus: Per recommendations at https://stackoverflow.com/a/11414236/553125 it seems like a good idea to have both version and branch name.

withinfocus (changes requested):

languages/kotlin/sdk/build.gradle
Hinton: After reading the comments to that answer and the other answers.

Can you verify that both maven and gradle won't prefer 0.3.0-sdk-android-pas-maven over 0.3.0? The specification is a bit confusing https://maven.apache.org/pom.html#version-order-specification.

```
$ java -jar maven-artifact-3.9.4.jar 0.3.0-sdk-android-ps-maven 0.3.0

Display parameters as parsed by Maven (in canonical form and as a list of tokens) and comparison result:
1. 0.3.0-sdk-android-ps-maven -> 0.3-sdk-android-ps-maven; tokens: [0, 3, [sdk, [android, [ps, [maven]]]]]
   0.3.0-sdk-android-ps-maven > 0.3.0
2. 0.3.0 -> 0.3; tokens: [0, 3]
```

withinfocus (approved):

PR #236, opened 2023-09-14
Use static libs for the iOS Swift framework

matt-livefront:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [X] Other
```

## Objective

Running an iOS app linked against the SDK on a device results in a crash because the dynamic library can't be found.

> dyld[11647]: Library not loaded: libbitwarden_uniffi.dylib

In looking through the Apple docs, I found the following:

> Avoid using dynamic library files (.dylib files) for dynamic linking. An XCFramework can include dynamic library files, but only macOS supports these libraries for dynamic linking. Dynamic linking on iOS, watchOS, and tvOS requires the XCFramework to contain .framework bundles.

> For each library and each platform, build a binary static libary file (.a file) that includes slices for all the possible architectures the platform uses.

https://developer.apple.com/documentation/xcode/creating-a-multi-platform-binary-framework-bundle#Avoid-issues-when-using-alternate-build-systems

This updates the build script to build the framework with static libraries instead. I've tested on a device and it seems to be working.

## Code changes

- **build.sh:** Updates the script to use static libraries when building the XCFramework.

## Before you submit

### DIFF
--- a/languages/swift/build.sh
+++ b/languages/swift/build.sh
@@ -15,9 +15,9 @@ cargo build --package bitwarden-uniffi --target aarch64-apple-ios --release
 cargo build --package bitwarden-uniffi --target x86_64-apple-ios --release
 
 # Create universal libraries
-lipo -create ../../target/aarch64-apple-ios-sim/release/libbitwarden_uniffi.dylib \
-  ../../target/x86_64-apple-ios/release/libbitwarden_uniffi.dylib  \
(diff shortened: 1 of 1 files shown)


Hinton (approved):
We'll revisit this if needed. It should be possible to embed `.dylib` within `.framework` if need be. It sounds like things might be dynamically linked either way though.

PR #241, opened 2023-09-18
Refactor LoginMethod & AuthSettings

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Refactor `LoginMethod` to contain details from `AuthSettings`. I believe this more accurately represents the available state for Service Accounts, Organization keys and other login methods like TDE where KDF aren't applicable.

To facilitate these the following changes were made:

- `request_prelogin` is now directly called instead of implicitly in `determine_password_hash`.
- Since `AuthSettings` is removed, the helper for `derive_user_password_hash` is gone, and the master key needs to be manually created. I believe this moves us in a direction where we avoid re-building it constantly. 

## Before you submit


crates/bitwarden/src/auth/login/api_key.rs
```diff
@@ -28,24 +26,26 @@ pub(crate) async fn api_key_login(
...
-
         let access_token_obj = decode_token(&r.access_token)?;
 
         // This should always be Some() when logging in with an api key
         let email = access_token_obj
             .email
             .ok_or(Error::Internal("Access token doesn't contain email"))?;
 
-        let _ = determine_password_hash(client, &email, &input.password).await?;
+        let kdf = request_prelogin(client, email.clone()).await?.try_into()?;
```
dani-garcia, line 36:
```suggestion
        let kdf = client.prelogin(email.clone()).await?;
```

crates/bitwarden/src/auth/login/two_factor.rs
```diff
@@ -19,7 +19,10 @@ pub(crate) async fn send_two_factor_email(
     client: &mut Client,
     input: &TwoFactorEmailRequest,
 ) -> Result<()> {
-    let password_hash = determine_password_hash(client, &input.email, &input.password).await?;
+    let kdf = request_prelogin(client, input.email.clone())
+        .await?
+        .try_into()?;
```
dani-garcia, line 24:
```suggestion
    let kdf = client.prelogin(input.email.clone()).await?;
```

crates/bitwarden/src/mobile/kdf.rs
```diff
@@ -10,10 +11,7 @@ pub async fn hash_password(
     password: String,
     kdf_params: Kdf,
 ) -> Result<String> {
-    let auth_settings = AuthSettings {
-        email,
-        kdf: kdf_params,
-    };
-    let hash = auth_settings.derive_user_password_hash(&password)?;
-    Ok(hash)
+    let master_key = MasterKey::derive(password.as_bytes(), email.as_bytes(), &kdf_params)?;
```
dani-garcia, line 14: We have this code repeated in multiple places now:
 - Here
 - In `make_register_keys`
 - In `determine_password_hash`
 - In `get_secret_verification_request`

Should we think of a way to avoid that repetition?

crates/bitwarden/src/client/encryption_settings.rs
```diff
@@ -28,30 +28,39 @@ impl std::fmt::Debug for EncryptionSettings {
 impl EncryptionSettings {
     #[cfg(feature = "internal")]
     pub(crate) fn new(
-        auth: &AuthSettings,
+        login_method: &LoginMethod,
```
dani-garcia, line 31: This should probably take a `UserLoginMethod` directly, it doesn't make sense for `ServiceAccount`, which would use `new_single_key` instead

crates/bitwarden/src/mobile/kdf.rs
Hinton: Maybe. I think it's a good indication that we should try and pass along master key where it makes sense. I think we're fine with the duplication for now.

crates/bitwarden/src/auth/login/api_key.rs
Hinton: Long term I would like to eliminate the explicit consumer prelogin call. Maybe that results in us needing an AuthBuilder or something.

dani-garcia (approved):

dani-garcia (approved):

PR #256, opened 2023-09-25
Convert symmetric crypto keys to use marker traits

MGibson1:
traits and structs are markers for the purpose of the key

## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Work on starting to make usage of crypto keys explicit at compile time

## Code changes

## Before you submit


Hinton:
After discussion with @MGibson1 we will continue using the new type pattern since it provides the benefit of not exposing the inner implementation. With the tradeoff of needing to duplicate traits on the newtype.

PR #261, opened 2023-09-29
[PM-3434] Password generator

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
Implement password generation
The `min_` fields were set to bool incorrectly I assume, so I changed them to u8.

At the moment if all the minimums turn out larger than the length, I just expand the length, but that seems wrong. I feel like it would be best to return an error in that case instead of changing any values behind the user, thoughts?


crates/bitwarden/src/tool/generators/password.rs
```diff
@@ -21,10 +22,10 @@ pub struct PasswordGeneratorRequest {
     pub length: Option<u8>,
 
     pub avoid_ambiguous: Option<bool>, // TODO: Should we rename this to include_all_characters?
-    pub min_lowercase: Option<bool>,
-    pub min_uppercase: Option<bool>,
-    pub min_number: Option<bool>,
-    pub min_special: Option<bool>,
+    pub min_lowercase: Option<u8>,
```
Hinton, line 25: Whops 😅

crates/bitwarden/src/tool/generators/password.rs
```diff
@@ -40,8 +41,85 @@ pub struct PassphraseGeneratorRequest {
...
+    let avoid_ambiguous = input.avoid_ambiguous.unwrap_or(false);
+    let lower_chars = gen_chars(&LOWER_CHARS, &LOWER_CHARS_AMBIGUOUS, avoid_ambiguous);
+    let upper_chars = gen_chars(&UPPER_CHARS, &UPPER_CHARS_AMBIGUOUS, avoid_ambiguous);
+    let number_chars = gen_chars(&NUMBER_CHARS, &NUMBER_CHARS_AMBIGUOUS, avoid_ambiguous);
+    let all_chars = lower_chars
+        .iter()
+        .chain(&upper_chars)
+        .chain(&number_chars)
+        .chain(&SPECIAL_CHARS)
+        .collect::<Vec<_>>();
```
Hinton, line 78: Would it make sense to extract this to a function and test it to ensure the correct chars are available?

Hinton (commented):
This looks good from a quick glance. I think we should try and break up the deterministic portions of the code and unit tests those. Add a few tests ensuring the generated password contains the needed characters.

Hinton (dismissed):

audrey-jensen:
> At the moment if all the minimums turn out larger than the length, I just expand the length, but that seems wrong. I feel like it would be best to return an error in that case instead of changing any values behind the user, thoughts?

Bitwarden desktop, web, and browser update the minimum password length when the minimum digit value changes. The behavior works as follows:

* When the min digit value is greater than 9, the value is set to 9.
* When the min special character quantity is greater than 9, the control sets its value to 9.
* When the min digit quantity is set, `[minimum password length] = [minimum digit quantity] + 2`. 
* When the min special character quantity is set, if its value is greater than `[minimum password length] - [minimum digit quantity]`, the value is set to the difference between these quantities.
* The absolute minimum password size is 5.

This behavior should be preserved in your implementation.

crates/bitwarden/src/tool/generators/password.rs
```diff
@@ -40,10 +41,249 @@ pub struct PassphraseGeneratorRequest {
...
+    }
+    for _ in 0..min_special {
+        buf.push(*chars.special.choose(&mut rng).expect("slice is not empty"));
+    }
+    for _ in min_length..length {
+        buf.push(*chars.all.choose(&mut rng).expect("slice is not empty"));
+    }
+
+    buf.shuffle(&mut rng);
+    Ok(buf.iter().collect())
```
audrey-jensen, line 179: 📝  Here's where you use `Distribtion<char>.sample_iter`. Create each iterator, then take the number you need. Chain the takes together, and collect them in a Vec. Shuffle the vec, then convert to a string.

:+1: Put another way, this code is doing exactly what it should, given that you're working with arrays.

crates/bitwarden/src/tool/generators/password.rs
```diff
@@ -40,10 +41,249 @@ pub struct PassphraseGeneratorRequest {
...
+        assert_eq!(set.lower, Vec::new());
+        assert!(to_string(&set.upper).contains(&to_string(UPPER_CHARS)));
+        assert!(to_string(&set.upper).contains(&to_string(UPPER_CHARS_AMBIGUOUS)));
+        assert_eq!(set.number, Vec::new());
+        assert_eq!(set.special, Vec::new());
+    }
+
+    #[test]
+    fn test_password_gen() {
+        let mut rng = rand_chacha::ChaCha8Rng::from_seed([0u8; 32]);
```
audrey-jensen, line 238: :+1: Very clever, using a seed to ensure the test has a fixed result.

crates/bitwarden/src/tool/generators/password.rs
```diff
@@ -40,10 +41,249 @@ pub struct PassphraseGeneratorRequest {
...
+mod test {
+    use rand::SeedableRng;
+
+    use super::*;
+
+    // We convert the slices to Strings to be able to use `contains`
+    // This wouldn't work if the character sets were ordered differently, but that's not the case for us
+    fn to_string(chars: &[char]) -> String {
+        chars.iter().collect()
+    }
```
audrey-jensen, line 196: ❓ I wonder how this would look if instead you used [HashSet::difference](https://doc.rust-lang.org/std/collections/hash_set/struct.HashSet.html#method.difference). Main reason is because it gets away from the whole ordering business. I have no clue what would happen with execution time.

crates/bitwarden/src/tool/generators/password.rs
```diff
@@ -40,10 +41,249 @@ pub struct PassphraseGeneratorRequest {
...
+
+    #[test]
+    fn test_password_characters_all() {
```
audrey-jensen, line 199: 💭 My gut says the unit test names should be more descriptive, but I have no idea how to make that compatible w/ rust's idioms.

audrey-jensen (changes requested):
* 👍 (:+1:) or similar for great changes
* ❓ (:question:) - BLOCKS REVIEW - for questions
* ⚠️ (:warning:) - BLOCKS APPROVAL - for more significant problems or concerns needing attention
* 📝 (:memo:) for notes or general info
* 💭 (:thought_balloon:) for open-ended inquiry that's not quite a confirmed issue and could potentially benefit from discussion
* 🎨 (:art:) for suggestions / improvements

crates/bitwarden/src/tool/generators/password.rs
```diff
@@ -40,10 +41,249 @@ pub struct PassphraseGeneratorRequest {
...
+    'V', 'W', 'X', 'Y', 'Z',
+];
+const LOWER_CHARS_AMBIGUOUS: &[char] = &['l'];
+const LOWER_CHARS: &[char] = &[
+    'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'm', 'n', 'o', 'p', 'q', 'r', 's', 't',
+    'u', 'v', 'w', 'x', 'y', 'z',
+];
+const NUMBER_CHARS_AMBIGUOUS: &[char] = &['0', '1'];
+const NUMBER_CHARS: &[char] = &['2', '3', '4', '5', '6', '7', '8', '9'];
+const SPECIAL_CHARS: &[char] = &['!', '@', '#', '$', '%', '^', '&', '*'];
```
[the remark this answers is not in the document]
dani-garcia: Implementing Distribution directly seems a bit too much if we do have to do the rejection sampling ourselves, but looking through the rand docs i found the `Slice` distribution which might just be perfect for us to use: https://docs.rs/rand/latest/rand/distributions/struct.Slice.html

crates/bitwarden/src/tool/generators/password.rs
dani-garcia: I haven't seen any recommended practices about testing naming conventions, but beyond splitting the tests into more functions with more descriptive names, the assert macros also have messages that they can print when they fail:
```rust
#[test]
fn test_password_characters_all() {
        let set = PasswordGeneratorCharSet::new(true, true, true, true, true);
        assert_eq!(set.lower, LOWER_CHARS, "set contains lower chars");
        assert_eq!(set.upper, UPPER_CHARS, "set contains upper chars");
        ...
    }
```

This way at least the error message will include a bit more detail: 
```
running 1 test
thread 'tool::generators::password::test::test_password_characters_all' panicked at crates/bitwarden/src/tool/generators/password.rs:201:9:
assertion `left == right` failed: set contains lower chars
  left: []
 right: ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'm', 'n', 'o', 'p', 'q', 'r', 's', 't', 'u', 'v', 'w', 'x', 'y', 'z']

```

crates/bitwarden/src/tool/generators/password.rs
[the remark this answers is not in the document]
audrey-jensen: You can create a [Uniform distribition](https://docs.rs/rand/latest/rand/distributions/struct.Uniform.html) or implement it using [Rng::gen_range](https://docs.rs/rand/latest/rand/trait.Rng.html#method.gen_range).

I whipped up a prototype that wraps the slices in `PwCharSet` and has a method to combine them. This represents the classes directly as their arrays. A more flexible system could keep a catalog, which would enable named mutually exclusive classes to be intersected and so forth. This implementation builds it fresh each time.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1c44366b5fef4ebaaf054d5d6b8021d1

crates/bitwarden/src/tool/generators/password.rs
```diff
@@ -40,10 +41,249 @@ pub struct PassphraseGeneratorRequest {
...
+    let min_special = get_minimum(input.min_special, input.special);
+
+    // Check that the minimum lengths aren't larger than the password length
+    let min_length = min_lowercase + min_uppercase + min_number + min_special;
+    let length = input.length.unwrap_or(DEFAULT_PASSWORD_LENGTH);
+    if min_length > length {
+        return Err(Error::Internal(
+            "Password length can't be less than the sum of the minimums",
+        ));
+    }
```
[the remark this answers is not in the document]
Hinton: I played around with that during my review, the issue is that the min_length is the minimum from min_lowercase and 1.

I had to do the calculations twice, or convert the struct to a new struct which seemed a bit clunky.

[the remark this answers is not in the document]
audrey-jensen: > or convert the struct to a new struct which seemed a bit clunky. 

In a language other than rust, I'd agree. Converting between structures in rust, however, is largely an expected circumstance. Consider, for example, [the number of structures involved in iteration](https://doc.rust-lang.org/std/iter/index.html#structs).

crates/bitwarden/src/tool/generators/password.rs
audrey-jensen: I've found the unit test guidance for .Net to be absolutely great when it comes to language-agnostic advice. [Here's what it says](https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices#naming-your-tests) about unit test names:

> The name of your test should consist of three parts:
>
>  * The name of the method being tested.
>  * The scenario under which it's being tested.
>  * The expected behavior when the scenario is invoked.
>
> ...Just by looking at the suite of unit tests, you should be able to infer the behavior of your code without even looking at the code itself. Additionally, when tests fail, you can see exactly which scenarios don't meet your expectations.

There's lots of ways to structure this data and everyone has their own opinion about the specifics, so I don't want to deep dive into exactly how to express this in rust code. My main goal is for the test methods to be more thoroughly described. There's one other benefit: using separate functions for each test isolates stateful operations, preventing them from affecting future tests.

> ...beyond splitting the tests into more functions with more descriptive names, the assert macros also have messages that they can print when they fail

This works only so long as your individual asserts reflect the higher-order goals of the test. Many tests require more than one assertion, which is why I'm focused on the function name.

Last but not least, rustdoc can [execute code examples as tests](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html). I'll leave figuring out if we'd like to leverage that for the future.

crates/bitwarden/src/tool/generators/password.rs
dani-garcia: Good point! I've changed it to use HashSet::is_superset and I think it makes more clear what the goal of the test is, plus now ordering is not an issue.

Tests still seem to run almost instantly, so I don't think it's going to be a problem.

crates/bitwarden/src/tool/generators/password.rs
```diff
@@ -40,10 +79,252 @@ pub struct PassphraseGeneratorRequest {
...
+
+    #[test]
+    fn test_password_characters_all() {
+        let set = PasswordGeneratorCharSet::new(true, true, true, true, true);
+        assert_eq!(set.lower, LOWER_CHARS);
+        assert_eq!(set.upper, UPPER_CHARS);
+        assert_eq!(set.number, NUMBER_CHARS);
+        assert_eq!(set.special, SPECIAL_CHARS);
+    }
+    #[test]
```
Hinton, line 247:
```suggestion

    #[test]
```

crates/bitwarden/src/tool/generators/password.rs
```diff
@@ -40,10 +79,252 @@ pub struct PassphraseGeneratorRequest {
...
+        let set = PasswordGeneratorCharSet::new(true, true, true, true, false);
+        assert!(to_set(&set.lower).is_superset(&to_set(LOWER_CHARS)));
+        assert!(to_set(&set.lower).is_superset(&to_set(LOWER_CHARS_AMBIGUOUS)));
+        assert!(to_set(&set.upper).is_superset(&to_set(UPPER_CHARS)));
+        assert!(to_set(&set.upper).is_superset(&to_set(UPPER_CHARS_AMBIGUOUS)));
+        assert!(to_set(&set.number).is_superset(&to_set(NUMBER_CHARS)));
+        assert!(to_set(&set.number).is_superset(&to_set(NUMBER_CHARS_AMBIGUOUS)));
+        assert_eq!(set.special, SPECIAL_CHARS);
+    }
+    #[test]
```
Hinton, line 258:
```suggestion

    #[test]
```

crates/bitwarden/src/tool/generators/password.rs
```diff
@@ -40,10 +79,252 @@ pub struct PassphraseGeneratorRequest {
...
+    fn test_password_characters_lower() {
+        let set = PasswordGeneratorCharSet::new(true, false, false, false, true);
+        assert_eq!(set.lower, LOWER_CHARS);
+        assert_eq!(set.upper, Vec::new());
+        assert_eq!(set.number, Vec::new());
+        assert_eq!(set.special, Vec::new());
+    }
+    #[test]
```
Hinton, line 266:
```suggestion

    #[test]
```

audrey-jensen (dismissed):
Just do what you're going to do.

crates/bitwarden/src/tool/generators/password.rs
```diff
@@ -1,30 +1,73 @@
...
 /// - uppercase
 /// - numbers
 ///
 /// The default length is 16.
-#[derive(Serialize, Deserialize, Debug, JsonSchema, Default)]
+#[derive(Serialize, Deserialize, Debug, JsonSchema)]
 #[serde(rename_all = "camelCase", deny_unknown_fields)]
 #[cfg_attr(feature = "mobile", derive(uniffi::Record))]
 pub struct PasswordGeneratorRequest {
+    /// When set to true, the generated password will contain lowercase characters (a-z).
```
Hinton, line 20: This seems a bit verbose, what do you think about `Include lowercase characters (a-z)`?

crates/bitwarden/src/tool/generators/password.rs
```diff
@@ -40,10 +83,342 @@ pub struct PassphraseGeneratorRequest {
...
+    buf.extend(set.sample_iter(&mut rng).take(*qty));
+
+    let (set, qty) = &options.lower;
+    buf.extend(set.sample_iter(&mut rng).take(*qty));
+
+    let (set, qty) = &options.number;
+    buf.extend(set.sample_iter(&mut rng).take(*qty));
+
+    let (set, qty) = &options.special;
+    buf.extend(set.sample_iter(&mut rng).take(*qty));
```
Hinton, line 262: Could we iterate over this?

```rust
let opts = [&options.all, &options.upper, &options.lower, &options.number, &options.special];
for (set, qty) in opts {
  buf.extend(set.sample_iter(&mut rng).take(*qty));
}
```

crates/bitwarden/src/tool/generators/password.rs
```diff
@@ -1,32 +1,403 @@
...
 #[cfg_attr(feature = "mobile", derive(uniffi::Record))]
 pub struct PasswordGeneratorRequest {
+    /// Include lowercase characters (a-z).
     pub lowercase: bool,
+    /// Include uppercase characters (A-Z).
     pub uppercase: bool,
+    /// Include numbers (0-9).
     pub numbers: bool,
+    /// Include special characters.
+    /// The supported characters are: ! @ # $ % ^ & *
```
Hinton, line 27:
```suggestion
    /// Include special characters: `! @ # $ % ^ & *`.
```

crates/bitwarden/src/tool/generators/password.rs
```diff
@@ -1,32 +1,403 @@
...
 use serde::{Deserialize, Serialize};
 
-/// Password generator request. If all options are false, the default is to
+use crate::error::{Error, Result};
+
+/// Password generator request options. If all options are false, the default is to
 /// generate a password with:
 /// - lowercase
 /// - uppercase
 /// - numbers
```
Hinton, line 13: Is this still true? Or should we remove the "default"?

PR #262, opened 2023-09-29
[PM-3435] Passphrase generator

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
Added passphrase generator, including `bw` cli support


crates/bitwarden/src/tool/generators/password.rs
```diff
@@ -44,6 +45,51 @@ pub(super) fn password(_input: PasswordGeneratorRequest) -> Result<String> {
...
+        }
+
+        if insert_number_idx == Some(idx) {
+            passphrase.push_str(&rand.gen_range(0..=9).to_string());
+        }
+
+        if idx != words - 1 {
+            passphrase.push(separator)
+        }
+    }
```
Hinton, line 92: **suggestion:** I think we can break this up into a few more easily understood functions that can be more easily tested.

```rust
/// Generate random words from the dictionary
fn gen_words(mut rng: impl RngCore, num_words: u8) -> Vec<String> {
    (0..num_words)
        .map(|_| {
            EFF_LONG_WORD_LIST
                .choose(&mut rng)
                .expect("slice is not empty")
                .to_string()
        })
        .collect()
}
```

This removes the need for a loop, adding a number means updating the index of the affected string, and the word string is a join using the separator as the joiner.

Hinton (changes requested):
This review uses [Conventional commits](https://conventionalcomments.org/).

I think this is an excellent start, the `password.rs` file is getting a bit large and we can benefit from separating the two generators. We can break up the generation in a few functions that are easily unit tested which will increase our confidence in the code long term.

crates/bitwarden/src/tool/generators/passphrase.rs
```diff
@@ -0,0 +1,112 @@
...
+fn gen_words(mut rng: impl RngCore, num_words: u8) -> Vec<String> {
+    (0..num_words)
+        .map(|_| {
+            EFF_LONG_WORD_LIST
+                .choose(&mut rng)
+                .expect("slice is not empty")
+                .to_string()
+        })
+        .collect()
+}
```
Hinton, line 52: **suggestion**: Let's add a quick test for this function using https://docs.rs/rand/latest/rand/rngs/mock/struct.StepRng.html.

dani-garcia: I've looked into it, but it seems that the `choose` function expects the numbers to be uniformly distributed in the range of `0..2^64` to give proper results, so a mock rng like StepRng which just returns an incremental counter ends up with the following results, which don't seem that interesting to test:
```
let rng = StepRng::new(0, 1);
println!("{:?}", gen_words(rng, 8));
// Prints: 
["abacus", "abacus", "abacus", "abacus", "abacus", "abacus", "abacus", "abacus"]
```

I've tried to play with the parameters of StepRng a bit but it doesn't help that much either and I get the same results most of the time.

We might be able to use a different RNG, like `rand_chacha`, and just seed it with a constant initial value? This would return the same sequence of values each time, but they'd be more randomly distributed https://docs.rs/rand_chacha/latest/rand_chacha/struct.ChaCha8Rng.html#method.from_seed

Hinton: I think that sounds good, we can add `rand_chacha` as a test dependency.

Hinton (dismissed):
Looks great, thanks.

Hinton (approved):

crates/bitwarden/src/tool/generators/passphrase.rs
```diff
@@ -0,0 +1,161 @@
...
+        .iter_mut()
+        .for_each(|w| *w = capitalize_first_letter(w));
+}
+
+fn capitalize_first_letter(s: &str) -> String {
+    let mut c = s.chars();
+    match c.next() {
+        None => String::new(),
+        Some(f) => f.to_uppercase().collect::<String>() + c.as_str(),
+    }
```
audrey-jensen, line 72: ⚠️ At first I thought direct string manipulation would be more straightforward, but forgot that unicode case conversion can grow the length of the string. I'd consider adding a comment here explaining the iterator approach solves that issue.

💭 At the same time, I still wonder whether there's a way to eliminate the match.

audrey-jensen (changes requested):
* ❌, ⚠️  - BLOCKS APPROVAL - for more significant problems or concerns needing attention
* 💭 - for open-ended inquiry that's not quite a confirmed issue and could potentially benefit from discussion

crates/bitwarden/src/tool/generators/passphrase.rs
```diff
@@ -0,0 +1,161 @@
...
+/// The default separator is `-` and default number of words is 3.
+#[derive(Serialize, Deserialize, Debug, JsonSchema, Default)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+#[cfg_attr(feature = "mobile", derive(uniffi::Record))]
+pub struct PassphraseGeneratorRequest {
+    pub num_words: Option<u8>,
+    pub word_separator: Option<String>,
+    pub capitalize: Option<bool>,
+    pub include_number: Option<bool>,
+}
```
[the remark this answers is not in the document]
dani-garcia: Done, I've specifically not mentioned any rust types and concepts in the comments as these request types will be exported for the SDK bindings for other languages.

crates/bitwarden/src/tool/generators/passphrase.rs
```diff
@@ -0,0 +1,161 @@
...
+
+const DEFAULT_PASSPHRASE_NUM_WORDS: u8 = 3;
+const DEFAULT_PASSPHRASE_SEPARATOR: char = ' ';
+
+pub(super) fn passphrase(input: PassphraseGeneratorRequest) -> Result<String> {
+    passphrase_with_rng(rand::thread_rng(), input)
+}
+
+fn passphrase_with_rng(mut rng: impl RngCore, input: PassphraseGeneratorRequest) -> Result<String> {
```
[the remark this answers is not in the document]
dani-garcia: I've added some checks for the number of words that we didn't have before, so I guess this is technically solved 😄. I'm returning an error instead of clamping the number of words to the 3..20 range like the TypeScript code does. This to me seems more appropriate for an SDK, as there's no feedback to the user if we were to clamp the value. The UI should be the one limiting the range to the users, what do you think?

crates/bitwarden/src/tool/generators/passphrase.rs
dani-garcia: Sounds good, note that there is also this option that works assuming all characters are ASCII (which for the words list would work):
```rust
fn capitalize_first_letter_ascii(s: &mut str) {
    if let Some(r) = s.get_mut(0..1) {
        r.make_ascii_uppercase();
    }
}
```
I don't think it's much more readable though

crates/bitwarden/src/tool/generators/passphrase.rs
```diff
@@ -10,28 +13,51 @@ use serde::{Deserialize, Serialize};
...
-    let separator = input
-        .word_separator
-        .and_then(|s| s.chars().next())
-        .unwrap_or(DEFAULT_PASSPHRASE_SEPARATOR);
     let capitalize = input.capitalize.unwrap_or(false);
     let include_number = input.include_number.unwrap_or(false);
 
+    let num_words = input.num_words.unwrap_or(DEFAULT_PASSPHRASE_NUM_WORDS);
+    if !(MINIMUM_PASSPHRASE_NUM_WORDS..=MAXIMUM_PASSPHRASE_NUM_WORDS).contains(&num_words) {
+        return Err(Error::Internal("'num_words' must be between 3 and 20"));
```
Hinton, line 48: Can we use the constants in the error?

dani-garcia: Not with the current Internal error, which is a static str slice, I think we should just update the Internal variant to String, but that's a bigger change than I wanted to include in this PR

Hinton: Can you create a ticket for that work?

crates/bitwarden/src/tool/generators/passphrase.rs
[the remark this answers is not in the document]
audrey-jensen: So long as the intention is maintained--explaining when a value is/should be present--then all good.

crates/bitwarden/src/tool/generators/passphrase.rs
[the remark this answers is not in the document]
audrey-jensen: I'm interpreting your comment as "The SDK shouldn't control the user experience", and I agree that doesn't make much sense. That said, I'd wonder two things.

1. Should there be a wrapping library that *does* control the user experience. That is, should there be a `bitwarden_services` library or similar that is responsible for user-level validation?
2. Should the SDK limit the number of words at all? (This one's rhetorical--it shouldn't, since that would apply an arbitrary limit to the user experience.)

crates/bitwarden/src/tool/generators/passphrase.rs
```diff
@@ -10,28 +13,51 @@ use serde::{Deserialize, Serialize};
...
+    /// Character separator between words in the generated passphrase.
+    /// If the value is set, it cannot be empty.
+    /// The default value when unset is `-`
     pub word_separator: Option<String>,
+    /// When set to true, capitalize the first letter of each word in the generated passphrase.
+    /// The default value when unset is `false`
     pub capitalize: Option<bool>,
+    /// When set to true, include a number at the end of one of the words in the generated passphrase.
+    /// The default value when unset is `false`
     pub include_number: Option<bool>,
```
audrey-jensen, line 29: ❓ I'd not use option here, since the value isn't really optional. I'd instead instantiate the type with the defaults pre-populated. If the user overrides them, then the default isn't used. That said, this is an FFI type, so I'm guessing the reason that strategy isn't used is because it doesn't map cleanly to the targeted language?

crates/bitwarden/src/tool/generators/passphrase.rs
```diff
@@ -10,28 +13,51 @@ use serde::{Deserialize, Serialize};
 #[serde(rename_all = "camelCase", deny_unknown_fields)]
 #[cfg_attr(feature = "mobile", derive(uniffi::Record))]
 pub struct PassphraseGeneratorRequest {
```
audrey-jensen, line 15: ⚠️ `PassphraseGeneratorRequest` is missing type-level documentation.

crates/bitwarden/src/tool/generators/passphrase.rs
```diff
@@ -65,6 +91,9 @@ fn capitalize_words(words: &mut [String]) {
 }
 
 fn capitalize_first_letter(s: &str) -> String {
+    // Unicode case conversion can change the length of the string, so we can't capitalize in place.
+    // Instead we extract the first character and convert it to uppercase. This returns
+    // an iterator which we collect into a string, and then append the rest of the input.
```
audrey-jensen, line 96: 💖

crates/bitwarden/src/tool/generators/passphrase.rs
```diff
@@ -10,28 +13,51 @@ use serde::{Deserialize, Serialize};
...
 const DEFAULT_PASSPHRASE_NUM_WORDS: u8 = 3;
-const DEFAULT_PASSPHRASE_SEPARATOR: char = ' ';
+const DEFAULT_PASSPHRASE_SEPARATOR: &str = " ";
+
+const MINIMUM_PASSPHRASE_NUM_WORDS: u8 = 3;
+const MAXIMUM_PASSPHRASE_NUM_WORDS: u8 = 20;
 
 pub(super) fn passphrase(input: PassphraseGeneratorRequest) -> Result<String> {
     passphrase_with_rng(rand::thread_rng(), input)
 }
```
audrey-jensen, line 40: ⚠️ Here's another undocumented public member. I'm not gonna point them all out, but they should all be documented--at least within tools-owned code.

crates/bitwarden/src/tool/generators/passphrase.rs
audrey-jensen: That's where I went as well. I'd avoid it since it doesn't generalize. The main thing I like about it isn't the `if let` (though I am a fan), but that it avoids allocations.

crates/bitwarden/src/tool/generators/passphrase.rs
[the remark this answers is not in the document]
dani-garcia: To me it's not so much about not controlling the user experience, but more about just rejecting invalid options instead of silently changing them under the user to make them valid.

If a SDK user requests a 40 word passphrase, we have two ways of going about it that I consider correct:
- Either not limiting the option at all, and returning a 40 word passphrase.
- Or erroring out and mentioning we limit the generation to 20 characters.

On the other hand, disregarding the user's request for a 40 word passphrase, silently changing the value to 20 and returning a 20 word passphrase is just incorrect to me.

I think we should enforce the limits at the SDK level, mostly to avoid possible inconsistencies in all the clients using the SDK. 
At the same time, the UI should do validation to make sure that users can't select an invalid combination of options, if that can be exposed as a generic library for all the clients that'd be great

crates/bitwarden/src/tool/generators/passphrase.rs
dani-garcia: Yeah as you mention, just having the default on the struct would be the best solution for a Rust library, but we're limited in that sense because the Default impl isn't exposed through the FFI layers at the moment.

crates/bitwarden/src/tool/generators/passphrase.rs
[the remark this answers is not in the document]
audrey-jensen: > I think we should enforce the limits at the SDK level, mostly to avoid possible inconsistencies in all the clients using the SDK.

If there are technical limitations to generating passphrases, then setting SDK bounds is completely cool. I would not, however, do so arbitrarily in the generator. This is a policy decision, and should be applied by the SDK's caller. As to why? Consider what happens if circumstances change to make >20-word passphrases reasonable. You'll need to re-release the SDK and all downstream products to reverse this change.

That said, I'm not entirely against this, so long as the generator isn't enforcing the policy. For example, you can introduce a "policy free" struct that is used by this method, implement a conversion trait, and enforce the policy in the trait's impl. That way, the generator doesn't need to change when the policy does.

I've created an example of how this can work [on the rust playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=76aa3acd3a4bbfafa8b5ca32167364d0) using `TryFrom<...>`. While I'm not enthused about keeping the policy compiled in the SDK, using the trait-based solution introduces the flexibility required to migrate policy if needed.

crates/bitwarden/src/tool/generators/passphrase.rs
audrey-jensen: Ok, so I think the issue I'm running into here is that we're using the FFI type directly in the generator code. I think the solution here is to introduce a type that encapsulates the generator's requirements. The request should be mapped into that type, and the mapping process applies the defaults. I added an example to `passphrase_with_rng`.

I applied a pretty similar change in #261, but did so to implement `Distribution<char>`, which seems a bit overkill here.

crates/bitwarden/src/tool/generators/passphrase.rs
```diff
@@ -10,28 +13,51 @@ use serde::{Deserialize, Serialize};
...
 
 const DEFAULT_PASSPHRASE_NUM_WORDS: u8 = 3;
-const DEFAULT_PASSPHRASE_SEPARATOR: char = ' ';
+const DEFAULT_PASSPHRASE_SEPARATOR: &str = " ";
+
+const MINIMUM_PASSPHRASE_NUM_WORDS: u8 = 3;
+const MAXIMUM_PASSPHRASE_NUM_WORDS: u8 = 20;
 
 pub(super) fn passphrase(input: PassphraseGeneratorRequest) -> Result<String> {
```
audrey-jensen, line 41: ⚠️ `passphrase` should also be documented, since it is public code.

crates/bitwarden/src/tool/generators/passphrase.rs
audrey-jensen: I don't follow what you're saying. Could you point to the code you're talking about?

crates/bitwarden/src/tool/generators/passphrase.rs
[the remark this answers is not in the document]
Hinton: > Consider what happens if circumstances change to make >20-word passphrases reasonable. You'll need to re-release the SDK and all downstream products to reverse this change.

The current architecture is centered around the reverse. We want to enforce consistency across clients, one way to do this is by moving these concerns into the SDK. A future iteration could be to explore if we should expose available UI options so that updating the SDK would also change the presentation layer.

crates/bitwarden/src/tool/generators/passphrase.rs
dani-garcia: The public API function is here: 
https://github.com/bitwarden/sdk/blob/8d54b988deda55fd7af5c369db308b7428c0c747/crates/bitwarden/src/tool/generators/client_generator.rs#L17-L38

Which just calls the function we are talking about, which is pub(super), so only accessible within the generators module.

This is the pattern we've been following to avoid having a super-long client_generator file.

crates/bitwarden/src/tool/generators/passphrase.rs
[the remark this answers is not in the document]
audrey-jensen: The 20-word limitation is built in to the SDK. This is about making sure we're using rust types to implement the default so that we can easily extend it later.

If this were C# code, it would be akin to introducing an interface for mocking purposes.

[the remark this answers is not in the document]
Hinton: I view the validation inside the generate function to be typical out of bounds checks. Consider it part of the generators specification, it can generate words between 3 and 20 character, it needs a valid separator.

I would expect C# to behave similarly, I should not be able to pass in wrong values without getting back an error.

I do think we can remove the optional aspects from the request, since the UI constraints will always have options for this. I have a commit for this, if you can verify it satisfies the concern.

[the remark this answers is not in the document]
audrey-jensen: [Lines 50-52](https://github.com/bitwarden/sdk/blob/16207a6127133e70f37ec247ea79a2aaa5472959/crates/bitwarden/src/tool/generators/passphrase.rs#L50-L52) are the problem. If there's a conflict with policy set in the administrative console, they should be ignored. 

Using `TryFrom<T>` and an intermediate options type seems like the proper solution from multiple perspectives:

* It does everything you need it to do without changing callsites.
* It provides a wide path forward for policy overrides.
* It isolates the decision making code as you've requested elsewhere in the PR.
* It reduces the scope of unit tests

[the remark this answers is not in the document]
Hinton: A policy cannot and will not be a allowed to override the generators inherit constraints. The generator specification is explicit in that it can generate a passphrase between 3 and 20 words, it can capitalize, include numbers and supports different separators.

Passing in a policy that says `min_words = 21` will always fail, either the policy validation will ask the user to change their settings, or fail due to the generator attempting to generate to long passphrase.

As for the `TryFrom` pattern, I don't believe it's a good fit here for several reasons.

1. We will not allow consumers to implement the `TryFrom` themselves. Doing so would let them violate the policies or the generators built in constraints.
2. We would need a wrapper struct `PolicyOptions(options, Some(policy))` just to handle `TryFrom` with multiple values.

We do not yet know how policies will be represented which is why I dislike "future proofing" code. I would expect policies to each be their own struct which can implement functions for validation, which would be a layer before the generators own validations.

crates/bitwarden/src/tool/generators/passphrase.rs
audrey-jensen: ⚠️ Cool cool. Please link from the inner `passphrase(...)` to this one using a [name link](https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html#linking-to-items-by-name).

crates/bitwarden/src/tool/generators/passphrase.rs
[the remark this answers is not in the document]
audrey-jensen: > A policy cannot and will not be a allowed to override the generators inherit constraints.

The generator could output gigabytes of output, given enough memory. Say that every word is 1kb in size. 20 words is 20kb. This isn't an inherent constraint, it's arbitrary. And it's an arbitrary constraint that *today* is overridable in the admin console, and that I confirmed must continue to work.

> We will not allow consumers to implement the TryFrom themselves.

Consumers within the SDK boundary or outside of it? Because this doesn't change the FFI at all. And if it's consumers within the SDK, then this argument is silly. Anyone on the tools team could come in and make this change tomorrow. In fact, the reason I want to introduce indirection here--whether using `TryFrom<T>` or some other code that runs before the method is called--is so that I can implement the policy override.

> I would expect policies to each be their own struct which can implement functions for validation, which would be a layer called by the generator, either before or after performing it's own boundary validation.

Then I expect the exact same thing from your code.

This whole conversation stemmed from my desire to have an *infallible* generator. Dani wants to check the arguments, and I gave him a way to do so. We're now stuck going back and forth about the mechanism. I don't care about that, it just solved the problem elegantly.

-----

I care that:

1. The generator function does not check policy.
2. That the arbitrary limits applied to the consumer product can be overridden by administrators.

`TryFrom<T>` is unacceptable, so let's move on. Choose your own design that accomplishes the following:

❌ Pass a type that requires no validation into the generator.
❌ Do not pass the FFI type into the generator.
❌ Write the generator to be infallible and/or only fail by panicking.

crates/bitwarden/src/tool/generators/passphrase.rs
audrey-jensen: Resolved by https://github.com/bitwarden/sdk/commit/16207a6127133e70f37ec247ea79a2aaa5472959.

crates/bitwarden/src/tool/generators/passphrase.rs
audrey-jensen: My bad. I missed that. Thanks! :)

audrey-jensen (dismissed):
Just do what you're going to do.

crates/bitwarden/src/tool/generators/passphrase.rs
```diff
@@ -0,0 +1,240 @@
...
+
+/// Implementation of the random passphrase generator. This is not accessible to the public API.
+/// See [`ClientGenerator::passphrase`](crate::ClientGenerator::passphrase) for the API function.
+///
+/// # Arguments:
+/// * `options`: Valid parameters used to generate the passphrase. To create it, use
+///     [`PassphraseGeneratorRequest::validate_options`](PassphraseGeneratorRequest::validate_options).
+pub(super) fn passphrase(options: ValidPassphraseGeneratorOptions) -> String {
+    passphrase_with_rng(rand::thread_rng(), options)
+}
```
Hinton, line 84: I think the input here should be the `PassphraseGeneratorRequest`. The `ValidPassphraseGeneratorOptions` is a internal struct within the passphrase module so it doesn't make much sense having it as part of the public interface. This also removes the need to have a "private" module.

Hinton (approved):

PR #268, opened 2023-10-05
[DEVOPS-1525] swift workflow for package publishing

mimartin12:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Build and deploy workflows for the `BitwardenFFI. xcframework` package.
## Code changes

- **.github/workflows/build-swift.yml:** Workflow that builds `rc` and `hotfix-rc-swift` branches, creating the versioned `BitwardenFFI.xcframework` package.
- **.github/workflows/.github/workflows/release-swift.yml** Workflow to create a GitHub release from `rc` or `hotfix-rc-swift` branches. Creating a GitHub tag and release for `BitwardenFFI.xcframework`. Included is a job to push changes to the `sdk-swift` repository and create a release tag.

## Before you submit


.github/workflows/build-swift.yml
```diff
@@ -0,0 +1,143 @@
+---
+name: Build Swift Package
+
+on:
+  push:
+    branches:
+      - "rc"
+      - "hotfix-rc-swift"
```
Hinton, line 8: What about master? We currently don't "release" the swift sdk so the pressing concern is to get master snapshots automated since we have to do those manually currently.

.github/workflows/build-swift.yml
```diff
@@ -0,0 +1,143 @@
...
+    paths:
+      - "languages/swift/**"
```
Hinton, line 10: We expect to rarely edit these files.

.github/workflows/build-swift.yml
```diff
@@ -0,0 +1,143 @@
...
+
+      - name: Build xcframework
+        working-directory: ./languages/swift
+        run: |
+          xcodebuild -create-xcframework \
+            -library ./target/libbitwarden_uniffi.a-aarch64-apple-ios/libbitwarden_uniffi.a \
+            -headers ./Headers \
+            -library ${{ github.workspace	}}/universal-ios-sim/release/libbitwarden_uniffi.a \
+            -headers ./Headers \
+            -output ./BitwardenFFI.xcframework
```
Hinton, line 136: We'll want to keep the `build.sh` for building locally and keeping these in sync can be slightly annoying.

Is there any reason we always run raw commands in the workflows instead of calling bash scripts? Splitting the current bash script would be one way we can re-use the same logic in the workflow and avoid maintaining two separate flows.

.github/workflows/build-swift.yml
```diff
@@ -0,0 +1,143 @@
...
+    env:
+      _PKG_VERSION: ${{ needs.version.outputs.package_version }}
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac # v4.0.0
+
+      - name: Install rust
+        uses: dtolnay/rust-toolchain@5cb429dd810e16ff67df78472fa81cf760f4d1c0 # stable
+        with:
+          toolchain: 1.70.0
```
Hinton, line 90: We generally use `stable`.

.github/workflows/release-swift.yml
```diff
@@ -0,0 +1,148 @@
...
+      - name: Setup Git
+        working-directory: sdk-swift
+        run: |
+          git config --local user.email "${{ env._BOT_EMAIL }}"
+          git config --local user.name "${{ env._BOT_NAME }}"
+
+      - name: Update files
+        run: |
+          cp --verbose -rf sdk/languages/swift/README.md sdk-swift/README.md
+          cp --verbose -rf sdk/languages/swift/Package.swift sdk-swift/Package.swift
```
Hinton, line 115: We need to update package.swift with the new url.

.github/workflows/build-swift.yml
mimartin12: We can call the Bash script here without issues I think. I will make that change.

.github/workflows/build-swift.yml
mimartin12: What is a better path to trigger this workflow on?

.github/workflows/release-swift.yml
mimartin12: Oh, you are right!
I was looking at this and I noticed that we need to add this entire codeblock, with different values of course.
```swift
        .binaryTarget(
            name: "BitwardenFFI",
            url: "https://bwlivefronttest.blob.core.windows.net/sdk/6b8b84f-BitwardenFFI.xcframework.zip",
            checksum: "28851ab2735ecfa357c4ee4b73282e3f7b3fbb2e9258af9ddd05f160f0d774c0"),
```
Is there a good way to do this that you know of? I am trying to think of a way without spilcing up the file with something like `sed` or something similar.

The `sed` command I used is here, it works, but doesn't maintain formatting.

```bash
sed -i '' 's|.binaryTarget(name: "BitwardenFFI", path: "BitwardenFFI.xcframework")|.binaryTarget(\
    name: "BitwardenFFI",\
    url: "https://bwlivefronttest.blob.core.windows.net/sdk/6b8b84f-BitwardenFFI.xcframework.zip",\
    checksum: "28851ab2735ecfa357c4ee4b73282e3f7b3fbb2e9258af9ddd05f160f0d774c0"),|' Package.swift
```
_values are just for testing_

.github/workflows/build-swift.yml
mimartin12: I was having a discussion around that in slack, in the CloudOps channel. I want to figure this out as well.

For all GitHub release artifacts, those will need to be sourced from the deployment branch. Which is `rc` or `hotifx-rc-swift`. GitHub release artifacts have to come from a tag, and then we create a GitHub release with the artifact attached.

To get `master` builds into a public place to source them in `package.swift`. Those would be stored in an Azure storage account. 

Or if the idea was to release straight from `master` automatically, and have every build be a "GitHub Release" we can go that route for hosting the artifact.

Hinton: We don't want every merge into master to be a new release.

.github/workflows/build-swift.yml
Hinton: I think triggering it on everything might make more sense. Changes to the rust code will change the output of the workflow.

.github/workflows/release-swift.yml
Hinton: Not on top of my mind.

.github/workflows/build-swift.yml
withinfocus: My request / feedback in Slack was to get full releases working first with all this, then to look at what `master` artifacts look like (perhaps in parallel). I understand the GitHub release is noisy for every merge and that's not the path, but I want to keep our maintenance burden as low as possible since GitHub Packages doesn't exist for Swift.

.github/workflows/build-swift.yml
mimartin12: Done!

.github/workflows/build-swift.yml
mimartin12: Ok, so releases from merges into `master` isn't the way we want to go.

Do either of you all like the route of `rc` or `hotfix`/`hotfix-rc` as the release branch, just like how it's currently structured in the workflow?

withinfocus: Assuming the need to get artifacts for every master merge is unneeded at the moment, I like the unification of process of using `rc` or similar to prep and ship a release with artifacts.

Hinton: Our requirement **today** is snapshot releases. We don't have any plans on publishing stable releases for the swift SDK for a fairly long time.

mimartin12:
I made some changes and brought in some implementation similar to https://github.com/bitwarden/sdk/blob/main/.github/workflows/release-go.yml. This makes it so that the release is created here for the `sdk-swift` repo.

.github/workflows/build-swift.yml
```diff
@@ -0,0 +1,69 @@
+---
+name: Build Swift Package
+
+on:
+  push:
+    branches:
+      - "rc"
+      - "hotfix-rc"
+      - "master"
```
vgrassia, line 9:
```suggestion
      - "main"
```

.github/workflows/build-swift.yml
```diff
@@ -0,0 +1,69 @@
...
+      - "DEVOPS-1525-swift-workflow-for-package-publishing"
```
vgrassia, line 10:
```suggestion
```

Was this for testing?

.github/workflows/release-swift.yml
```diff
@@ -0,0 +1,174 @@
...
+    name: Validate Branch - Set Version
+    runs-on: ubuntu-22.04
+    outputs:
+      version: ${{ steps.version.outputs.version }}
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@8ade135a41bc03ea155e62e844d188df1ea18608 # v4.1.0
+
+      - name: Branch check
+        if: ${{ github.event.inputs.release_type != 'Dry Run' }}
```
vgrassia, line 32:
```suggestion
        if: ${{ inputs.release_type != 'Dry Run' }}
```

.github/workflows/release-swift.yml
```diff
@@ -0,0 +1,174 @@
...
+          cp --verbose -rf sdk/languages/swift/Sources sdk-swift/Sources
+          cp --verbose -rf sdk/languages/swift/Tests sdk-swift/Tests
+
+      - name: Push changes
+        working-directory: sdk-swift
+        run: |
+          git add .
+          git commit -m "Update Swift SDK to ${{ github.sha }}"
+
+          if [[ "${{ github.event.inputs.release_type }}" == "Dry Run" ]]; then
```
vgrassia, line 108:
```suggestion
          if [[ "${{ inputs.release_type }}" == "Dry Run" ]]; then
```

.github/workflows/release-swift.yml
```diff
@@ -0,0 +1,174 @@
...
+            echo "[!] Dry Run - Skipping push"
+            echo "==================================="
+            git ls-files -m
+            exit 0
+          else 
+            git push origin main
+          fi
+
+      - name: Create release tag on SDK Swift repo
+        if: ${{ github.event.inputs.release_type != 'Dry Run' }}
```
vgrassia, line 119:
```suggestion
        if: ${{ inputs.release_type != 'Dry Run' }}
```

.github/workflows/release-swift.yml
```diff
@@ -0,0 +1,174 @@
...
+          keyvault: ${{ env._KEY_VAULT }}
+          secrets: "github-pat-bitwarden-devops-bot-repo-scope"
+
+      - name: Download BitwardenEFI artifact
+        uses: bitwarden/gh-actions/download-artifacts@main
+        with:
+          workflow: build-swift.yml
+          artifacts: BitwardenFFI-${{ env._PKG_VERSION }}.xcframework
+          workflow_conclusion: success
+          branch: ${{ github.event.inputs.release_type == 'Dry Run' && 'master' || github.ref_name }}
```
vgrassia, line 160:
```suggestion
          branch: ${{ inputs.release_type == 'Dry Run' && 'main' || github.ref_name }}
```

.github/workflows/build-swift.yml
```diff
@@ -0,0 +1,69 @@
...
+        uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac # v4.0.0
+
+      - name: Get Package Version
+        id: retrieve-version
+        run: |
+          VERSION=$(grep -o '^version = ".*"' crates/bws/Cargo.toml | grep -Eo "[0-9]+\.[0-9]+\.[0-9]+")
+          echo "package_version=$VERSION" >> $GITHUB_OUTPUT
+
+  build:
+    runs-on: macos-13
```
vgrassia, line 29:
```suggestion
    name: Build
    runs-on: macos-13
```

vgrassia (changes requested):

.github/workflows/build-swift.yml
mimartin12: Yep, I can remove it for now, but I may add it back if there are some changes.

.github/workflows/build-swift.yml
```diff
@@ -0,0 +1,69 @@
...
+
+          # SHA Short
+          echo "short-sha=$(git rev-parse --short HEAD)" >> $GITHUB_OUTPUT
+
+      - name: Upload BitwardenFFI.xcframework artifact
+        uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
+        with:
+          name: BitwardenFFI-${{ env._VERSION }}-${{ steps.build.outputs.short-sha }}.xcframework
+          path: languages/swift/BitwardenFFI.xcframework
+          if-no-files-found: error
```
Hinton, line 69: This seems to upload every commit from main in our artifactory, but it's never used?

mimartin12: Oh, I see what you mean. Yes, currently the trigger here https://github.com/bitwarden/sdk/pull/268/files/e336370aa3e6b19b4167b200c2e0df48285b4c2d#diff-055e2da2434f07c0a866c511bfa44f89960b26b18915681bd5a97ed5a8689b9fR9 is going to create an artifact that will not be used in the release workflow. If there is no value in having a GitHub artifact from `main`, I can remove it.

Hinton: From my perspective the main value is automatically pushing commits from main to the sdk-swift repository. As this repository will get more activity depending on devops to trigger development releases whenever a new feature gets merged will be tedious

mimartin12: https://github.com/bitwarden/sdk/pull/268/commits/a15a57aac19eb951861546a9ef0649149b3e0314
Does this now, I removed release process items around it, and made it simply push to the repository when a build happens on `main`. This adheres to the requirement to use GitHub releases to store the artifact.

.github/workflows/build-swift.yml
```diff
@@ -0,0 +1,102 @@
...
+          rustup target install aarch64-apple-ios
+          rustup target install x86_64-apple-ios
+
+      - name: Build
+        shell: bash
+        working-directory: languages/swift
+        id: build
+        run: |
+          chmod +x build.sh
+          ./build.sh
```
Hinton, line 59: This file should already have the executable permission, if not we should re-commit it with `+x`

.github/workflows/release-swift.yml
```diff
@@ -0,0 +1,157 @@
...
+        working-directory: sdk-swift
+        run: |
+          git add .
+          git commit -m "Update Swift SDK to ${{ needs.validate.outputs.sha }}"
+          git push origin main
+
+      - name: Create release tag on SDK Swift repo
+        working-directory: sdk-swift
+        run: |
+          git tag ${{ env._PKG_VERSION }}-${{ needs.validate.outputs.short_sha }}
```
Hinton, line 113: I don't think we want to tag "non stable" releases.

.github/workflows/release-swift.yml
```diff
@@ -0,0 +1,157 @@
...
+          git push origin ${{ env._PKG_VERSION }}-${{ needs.validate.outputs.short_sha }}
+
+  github-release:
```
Hinton, line 116: We should only do github releases for stable.

.github/workflows/build-swift.yml
mimartin12: image
Looks like it is!

.github/workflows/release-swift.yml
mimartin12: How can we do this? How do we define stable releases of the SDK today?

Hinton: When we desire external parties to use the new version of the sdk, generally whenever secrets manager wants a new release.

PR #278, opened 2023-10-11
[PM-4270] Individual cipher key encryption

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
Implement individual vault item encryption and encryption. Note that at the moment it will not migrate items that don't have a key when encrypting.

### DIFF
--- a/crates/bitwarden/src/vault/cipher/cipher.rs
+++ b/crates/bitwarden/src/vault/cipher/cipher.rs
@@ -43,6 +43,7 @@ pub struct Cipher {
     pub folder_id: Option<Uuid>,
     pub collection_ids: Vec<Uuid>,
 
+    /// More recent ciphers uses individual encryption keys to encrypt the other fields of the Cipher.
     pub key: Option<EncString>,
 
     pub name: EncString,
@@ -135,8 +136,8 @@ pub struct CipherListView {
 
 impl KeyEncryptable<Cipher> for CipherView {
     fn encrypt_with_key(self, key: &SymmetricCryptoKey) -> Result<Cipher> {
-        let key_owned = Cipher::get_cipher_key(key, &self.key)?;
-        let key = key_owned.as_ref().unwrap_or(key);
+        let ciphers_key = Cipher::get_cipher_key(key, &self.key)?;
+        let key = ciphers_key.as_ref().unwrap_or(key);
 
         Ok(Cipher {
             id: self.id,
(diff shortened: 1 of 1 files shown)


crates/bitwarden/src/vault/cipher/cipher.rs
```diff
@@ -43,6 +43,8 @@ pub struct Cipher {
     pub folder_id: Option<Uuid>,
     pub collection_ids: Vec<Uuid>,
 
+    pub key: Option<EncString>,
```
Hinton, line 46: Can we add a comment here what the key is.

```suggestion
    /// More recent ciphers uses individual encryption keys to encrypt the other fields of the Cipher.
    pub key: Option<EncString>,
```

crates/bitwarden/src/vault/cipher/cipher.rs
```diff
@@ -77,6 +79,8 @@ pub struct CipherView {
...
```
Hinton, line 82: Do we want to expose the key in a view?

crates/bitwarden/src/vault/cipher/cipher.rs
```diff
@@ -131,11 +135,15 @@ pub struct CipherListView {
 
 impl KeyEncryptable<Cipher> for CipherView {
     fn encrypt_with_key(self, key: &SymmetricCryptoKey) -> Result<Cipher> {
+        let key_owned = Cipher::get_cipher_key(key, &self.key)?;
+        let key = key_owned.as_ref().unwrap_or(key);
```
Hinton, line 139: **suggestion**: I think we can improve the clarify here with some better variable names.

```suggestion
        let ciphers_key = Cipher::get_cipher_key(key, &self.key)?;
        let key = ciphers_key.as_ref().unwrap_or(key);
```

crates/bitwarden/src/vault/cipher/cipher.rs
```diff
@@ -190,6 +202,19 @@ impl KeyDecryptable<CipherView> for Cipher {
...
+        inner_key: &Option<EncString>,
+    ) -> Result<Option<SymmetricCryptoKey>> {
+        inner_key
+            .as_ref()
+            .map(|k| {
+                let key: Vec<u8> = k.decrypt_with_key(key)?;
+                SymmetricCryptoKey::try_from(key.as_slice())
+            })
+            .transpose()
+    }
```
Hinton, line 216: **suggestion:** Can we add a comment what this function is doing?

crates/bitwarden/src/vault/cipher/cipher.rs
dani-garcia: If we want to encrypt it back with the same Key, yeah. Though I guess we can generate a new key each time we encrypt a CipherView into a Cipher, if we wanted.

dani-garcia: Should we add a TODO or open a ticket for the future?

Hinton (approved):

PR #285, opened 2023-10-13
[PM-3436] Username generator

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
Implement username generation.

Some of the methods are shared with password and passphrase generation (like `capitalize_first_letter`, `random_lowercase_string` and the username word generation). To avoid dependencies between the PRs, I included them here again, once we're done with all the generators I'll unify them.


crates/bitwarden-uniffi/src/tool/mod.rs
```diff
@@ -35,6 +38,18 @@ impl ClientGenerators {
             .passphrase(settings)
             .await?)
     }
+
+    /// **API Draft:** Generate Username
```
Hinton, line 42: I assume this isn't a draft any longer?

crates/bitwarden/src/client/client.rs
```diff
@@ -132,6 +132,11 @@ impl Client {
         &self.__api_configurations
     }
 
+    #[cfg(feature = "mobile")]
+    pub(crate) fn get_http_client(&self) -> &reqwest::Client {
+        &self.__api_configurations.api.client
```
Hinton, line 137: Do we want to expose the api configurations http client? I think it pre-sets some headers that may be sensitive.

crates/bitwarden/src/tool/generators/username.rs
```diff
@@ -0,0 +1,237 @@
...
+        /// This is not used in all services, and is only used for display purposes
+        website: Option<String>,
+    },
+}
+
+#[derive(Serialize, Deserialize, Debug, JsonSchema)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+#[cfg_attr(feature = "mobile", derive(uniffi::Record))]
+pub struct UsernameGeneratorRequest {
+    pub r#type: UsernameGeneratorType,
```
Hinton, line 87: Should we just rename this to `strategy`? To avoid the name collision. Or make the input the `UsernameGeneratorType` directly.

crates/bitwarden/src/tool/generators/username.rs
```diff
@@ -0,0 +1,237 @@
+use crate::{error::Result, util::capitalize_first_letter, wordlist::EFF_LONG_WORD_LIST};
+use rand::{seq::SliceRandom, Rng, RngCore};
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Debug, JsonSchema)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+#[cfg_attr(feature = "mobile", derive(uniffi::Enum))]
+pub enum AddressType {
```
Hinton, line 9: I believe this should be `SubaddressType`? It's a bit confusing when it gets appended to the email. Or mayby `AppendType`.

crates/bitwarden/src/tool/generators/username.rs
```diff
@@ -0,0 +1,237 @@
...
+#[derive(Serialize, Deserialize, Debug, JsonSchema)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+#[cfg_attr(feature = "mobile", derive(uniffi::Enum))]
+pub enum UsernameGeneratorType {
+    /// Generates a single word username
+    Word {
+        /// When set to true, capitalizes the first letter of the word. Defaults to false
+        capitalize: Option<bool>,
+        /// When set to true, includes a 4 digit number at the end of the word. Defaults to false
+        include_number: Option<bool>,
```
Hinton, line 56: Should we just require these instead of making them optional? Similar to the other generators.

crates/bitwarden/src/tool/generators/username_forwarders/fastmail.rs
```diff
@@ -0,0 +1,106 @@
...
+        {
+            // TODO: Once we have a more flexible type of error, we can return this error_description
+            return Err(Error::Internal("Unknown Fastmail error occurred."));
+        };
+    } else if method_response == Some("error") {
+        let _description = response_value
+            .and_then(|r| r.get("description"))
+            .and_then(|r| r.as_str());
+
+        // TODO: Once we have a more flexible type of error, we can return this error_description
```
Hinton, line 80: Suggestion, make a new error type for forwarders, inline fastmail and inline the forwarders error in our regular error?

crates/bitwarden/src/tool/generators/username_forwarders/addyio.rs
```diff
@@ -0,0 +1,63 @@
...
+) -> Result<String> {
+    if api_token.is_empty() {
+        return Err(Error::Internal("Invalid addy.io API token."));
+    }
+    if domain.is_empty() {
+        return Err(Error::Internal("Invalid addy.io domain."));
+    }
+    if base_url.is_empty() {
+        return Err(Error::Internal("Invalid addy.io url."));
+    }
```
Hinton, line 19: This is a bit ugly. Would we benefit from making a NonEmptyString with a try/from into a string?

crates/bitwarden/src/tool/generators/username_forwarders/firefox.rs
```diff
@@ -0,0 +1,54 @@
...
+        enabled: bool,
+        generated_for: Option<String>,
+        description: String,
+    }
+
+    let description = website
+        .as_ref()
+        .map(|w| format!("{w} - "))
+        .unwrap_or_default();
+    let description = format!("{description}Generated by Bitwarden.");
```
Hinton, line 27: We seem to be duplicating this in many of the handlers. Should we pre-process this as part of the input?

crates/bitwarden/src/tool/generators/username.rs
```diff
@@ -0,0 +1,237 @@
...
+    let num = rng.gen_range(0..=9999);
+    format!("{num:0>4}")
+}
+
+fn random_lowercase_string(mut rng: impl RngCore, length: usize) -> String {
+    const CHARSET: &[u8] = b"abcdefghijklmnopqrstuvwxyz1234567890";
+    (0..length)
+        .map(|_| (*CHARSET.choose(&mut rng).expect("slice is not empty")) as char)
+        .collect()
+}
```
Hinton, line 190: Should we try and build up some standard distribution for use with `rng`?

crates/bitwarden/src/tool/generators/username.rs
```diff
@@ -0,0 +1,237 @@
...
+                    base_url,
+                } => addyio::generate(http, api_token, domain, base_url, website).await,
+                DuckDuckGo { token } => duckduckgo::generate(http, token).await,
+                Firefox { api_token } => firefox::generate(http, api_token, website).await,
+                Fastmail { api_token } => fastmail::generate(http, api_token, website).await,
+                ForwardEmail { api_token, domain } => {
+                    forwardemail::generate(http, api_token, domain, website).await
+                }
+                SimpleLogin { api_key } => simplelogin::generate(http, api_key, website).await,
+            }
```
Hinton, line 124: Slightly clunky, would it make sense to extract this to a trait, or are we limited by the rust toolchain again? Would be nice to just do `service.generate()`

crates/bitwarden/src/tool/generators/username.rs
Hinton, line 1: Can we add some comments to these functions?

Hinton (commented):
Looks good, a few comments. Would be nice if we either in this PR or a followup write some integration tests for the external forwarders, i.e. mock the response.

We should consider if there is an easy way to have e2e tests for this to, and call the external services.

crates/bitwarden-uniffi/src/tool/mod.rs
dani-garcia: The username and password generators also have the API draft notice in this file, we would probably need to remove those as well. I can change them all on this PR or leave it for another one after.

crates/bitwarden/src/client/client.rs
dani-garcia: We don't really set any headers on the client at the moment:
https://github.com/bitwarden/sdk/blob/7ec2183ab8194a05c994c9c67909dec538da1096/crates/bitwarden/src/client/client.rs#L87-L92

We probably had some in the past, because as it is, that `headers` map is not very useful. At the moment all the other headers are added to each request individually. We also have cookies disabled on the client so I think it should be reasonably safe.

crates/bitwarden/src/tool/generators/username_forwarders/addyio.rs
```diff
@@ -0,0 +1,63 @@
...
+    struct Request {
+        domain: String,
+        description: String,
+    }
+
+    let response = http
+        .post(format!("{base_url}/api/v1/aliases"))
+        .header(CONTENT_TYPE, "application/json")
+        .bearer_auth(api_token)
+        .header("X-Requested-With", "XMLHttpRequest")
```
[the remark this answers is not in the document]
dani-garcia: The official docs do include it yeah https://app.addy.io/docs/#aliases-POSTapi-v1-aliases

crates/bitwarden/src/tool/generators/username_forwarders/addyio.rs
dani-garcia: I've added these as they also were there on the current clients code. I'm wondering if they make any sense at all here, though. Using `"!&%$"` as domain is just as invalid as `""`, maybe we can just remove the checks entirely and let the API request fail. What do you think?

crates/bitwarden/src/tool/generators/username_forwarders/firefox.rs
dani-garcia: I've extracted it to a function at least, as not all the services use it. I've noticed that at the moment Firefox is using a slightly different message. I don't know if that matters, but I defined a separate function for it.

crates/bitwarden/src/tool/generators/username.rs
dani-garcia: Yeah, sounds good, done

crates/bitwarden/src/tool/generators/username.rs
dani-garcia: Moved the type directly to the request

crates/bitwarden/src/tool/generators/username.rs
dani-garcia: Good idea, done

crates/bitwarden/src/tool/generators/username.rs
dani-garcia: This is used in both the Subaddress and the Catch all variants, so calling it SubaddressType might not be great. Something like AppendType could work yeah

crates/bitwarden/src/tool/generators/username.rs
dani-garcia: Extracted to a generate function in ForwarderServiceType

crates/bitwarden/src/tool/generators/username_forwarders/addyio.rs
Hinton: Sounds good to me.

crates/bitwarden/src/tool/generators/username.rs
```diff
@@ -0,0 +1,244 @@
...
+        AppendType::Random => random_lowercase_string(&mut rng, 8),
+        AppendType::WebsiteName { website } => website,
+    };
+
+    format!("{}@{}", email_start, domain)
+}
+
+fn random_lowercase_string(mut rng: impl RngCore, length: usize) -> String {
+    let dist = rand::distributions::Slice::new(b"abcdefghijklmnopqrstuvwxyz1234567890")
+        .expect("Non-empty slice");
```
Hinton, line 191: Let's extract it to a constant, `LOWERCASE_ALPHANUMERICAL`?

crates/bitwarden/src/tool/generators/username_forwarders/forwardemail.rs
```diff
@@ -0,0 +1,129 @@
...
+
+        message: Option<String>,
+        error: Option<String>,
+    }
+    let status = response.status();
+    let response: Response = response.json().await?;
+
+    if let Some(message) = response.message {
+        return Err(Error::ResponseContent { status, message });
+    }
```
Hinton, line 77: Is this correct? Seems weird to throw an error on `message`?

Hinton (commented):
Generally looks good, would have liked if the forwarder tests could verify the error conditions to.

crates/bitwarden/src/tool/generators/username_forwarders/mod.rs
```diff
@@ -0,0 +1,22 @@
...
+    format!("{description}Generated by Bitwarden.")
+}
+
+fn format_description_ff(website: &Option<String>) -> String {
+    let description = website
+        .as_ref()
+        .map(|w| format!("{w} - "))
+        .unwrap_or_default();
+    format!("{description}Generated by Bitwarden.")
+}
```
Hinton, line 22: @kspearrin do you remember why Firefox forwarder uses a different description than the other formatters?

kspearrin: I do not, sorry.

crates/bitwarden/src/tool/generators/username_forwarders/forwardemail.rs
dani-garcia: The API response when there's an error is like this:
```
{
  "statusCode": 401,
  "error": "Unauthorized",
  "message": "Invalid API token."
}
```

So the `message` field is just an explanation of the `error` field. Not a great name but that's what the API gives us.

Hinton (approved):

crates/bitwarden/src/tool/generators/username_forwarders/mod.rs
Hinton: 👍, let's keep it for now but with the knowledge we can revisit this in the future.

PR #294, opened 2023-10-16
Lock SDK to 6.0.100 and ignore with Renovate updates

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [x] Other
```

## Objective

Disable renovate from bumping the dotnet-sdk. We have prefer latest which should be sufficient.

## Before you submit

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -44,10 +44,7 @@
       "groupName": "gh minor",
       "matchManagers": ["github-actions"],
       "matchUpdateTypes": ["minor", "patch"]
-    },
-    {
-      "matchPackageNames": ["dotnet-sdk"],
-      "enabled": false
     }
-  ]
+  ],
+  "ignoreDeps": ["dotnet-sdk"]
 }
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -44,6 +44,10 @@
       "groupName": "gh minor",
       "matchManagers": ["github-actions"],
       "matchUpdateTypes": ["minor", "patch"]
+    },
+    {
+      "matchPackageNames": ["dotnet-sdk"],
+      "enabled": false
     }
   ]
 }
--- a/languages/csharp/global.json
+++ b/languages/csharp/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "6.0.413",
+    "version": "6.0.100",
     "rollForward": "latestFeature"
   }
 }


withinfocus:
Is https://github.com/bitwarden/sdk/pull/286 an example? I don't see the harm really despite the `rollForward`.

Hinton:
@withinfocus yes. It's an annoyance since the dotnet tooling will complain about missing SDK version. Requiring developers to always be on the latest version can be somewhat tedious.

withinfocus:
You mean the runners not being able to pull them? On a weekly cadence that seems rare since these get out in a day or two.

Your point is certainly valid since this will still roll up to the latest release, but _when_ that's available could actually be of importance to us and our team / processes. Microsoft may issue an SDK update here for security reasons and we want to force the updates and conversation if we can; the potentially week-long delay for the update is already large and we want to push upgrades wherever as possible as quickly as possible. If we weren't in the security software space I think this change would be an easier sell, but I am taking pause here because in our seat we want to push harder.

Hinton:
No, I'm referring to the developer workflow. A developer should not have to update their SDK every time a patch released is issued. We should enforce that the deployed code is built with appropriate releases though. But that's outside the scope of the `global.json`.

withinfocus:
Hmm not sure I agree -- we want our developers updated too, and this is ease to automate. With one regular command that I run regularly locally I get these updates (https://github.com/isen-ng/homebrew-dotnet-sdk-versions).

Hinton:
I generally try and avoid taps and installs it directly from Microsoft to ensure the integrity.

withinfocus:
Relates to https://github.com/bitwarden/server/pull/3478 -- we are opting for an easier developer experience as of now.

.github/renovate.json
```diff
@@ -45,5 +45,6 @@
       "matchManagers": ["github-actions"],
       "matchUpdateTypes": ["minor", "patch"]
     }
-  ]
+  ],
```
withinfocus, line 48: ℹ️ If lockfiles show up for the .NET SDK here (I hope they don't) then we may need to force a Renovate constraint.

languages/csharp/global.json
```diff
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "6.0.413",
+    "version": "6.0.100",
```
withinfocus, line 3: ℹ️ I may take a quick look at an SDK upgrade to 8, unless you foresee an issue.

withinfocus (approved):

languages/csharp/global.json
Hinton: As long as the package remains compatible with all actively supported .NET installs. We may need to explore adding .net standard support in the future.

PR #295, opened 2023-10-17
Mock a TOTP implementation

Hinton:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Mock a TOTP implementation

## Before you submit

### DIFF
--- a/crates/bitwarden-uniffi/src/vault/mod.rs
+++ b/crates/bitwarden-uniffi/src/vault/mod.rs
@@ -1,6 +1,7 @@
 use std::sync::Arc;
 
 use bitwarden::vault::TotpResponse;
+use chrono::{DateTime, Utc};
 
 use crate::Client;
 
@@ -46,7 +47,7 @@ impl ClientVault {
     /// - A base32 encoded string
     /// - OTP Auth URI
     /// - Steam URI
-    pub async fn generate_totp(&self, key: String) -> TotpResponse {
+    pub async fn generate_totp(&self, key: String, time: Option<DateTime<Utc>>) -> TotpResponse {
         self.0 .0.read().await.vault().generate_totp(key).await
     }
 }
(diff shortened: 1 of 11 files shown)


crates/bitwarden-uniffi/src/vault/mod.rs
```diff
@@ -37,4 +39,14 @@ impl ClientVault {
...
+
+    /// Generate a TOTP code from a provided key.
+    ///
+    /// The key can be either:
+    /// - A base32 encoded string
+    /// - OTP Auth URI
+    /// - Steam URI
+    pub async fn generate_totp(&self, key: String) -> TotpResponse {
+        self.0 .0.read().await.vault().generate_totp(key).await
+    }
```
Hinton, line 51: I wasn't sure where to put this, it realistically only for vault ciphers but it's more similar to cryptography then anything else.

crates/bitwarden-uniffi/src/vault/mod.rs
```diff
@@ -37,4 +39,14 @@ impl ClientVault {
...
         Arc::new(sends::ClientSends(self.0.clone()))
     }
+
+    /// Generate a TOTP code from a provided key.
+    ///
+    /// The key can be either:
+    /// - A base32 encoded string
+    /// - OTP Auth URI
+    /// - Steam URI
+    pub async fn generate_totp(&self, key: String) -> TotpResponse {
```
dani-garcia, line 49: Should we also add the time here, in case any client wants to generate the previous/next code? Or maybe as a separate function?

crates/bitwarden-uniffi/src/vault/mod.rs
dani-garcia: I think it's fine in vault, we can even put it in the ciphers subclient inside vault too.

dani-garcia (approved):

dani-garcia (approved):

PR #296, opened 2023-10-19
Return a non-zero exit code when no subcommands are supplied

tangowithfoxtrot:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Return a non-zero exit code when no subcommands are supplied. This matches the behavior in many other CLIs (`bw`, `git`, `ssh`, etc.).

## Code changes

- **`crates/bws/src/main.rs`:** Exit with a non-zero exit code

## Before you submit


dani-garcia (approved):
Nice, thanks!

coltonhurst (approved):
Looks good!

PR #297, opened 2023-10-19
Split Encryptable/Decryptable trait into encryption and key retrieval

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
At the moment, the `Encryptable`/`Decryptable` trait handles both encryption/decryption and key retrieval. The goal of this PR is to split this trait into two parts, one part that only handles encryption/decryption (and thus will be able to be moved into a separate `bitwarden-crypto` crate, if we want), and another part that handles retrieving the corresponding encryption key for each model object. 

## Code changes
What was the `Encryptable`/`Decryptable` trait has been split into multiple parts:
- `KeyEncryptable`/`KeyDecryptable`: This receives only the encryption key and handles encryption/decryption only. The name of the interface was chosen to match the current `EncString::decrypt_with_key`, which this replaces.
- `LocateKey`, which is implemented by types that know the type of key they need. For example a `Folder` always uses the user key and a `Cipher` uses the user key or one of the organization keys depending on its `organization_id` field.
- `Encryptable`/`Decryptable`, this is now a thin compatibility layer between the previously mentioned two, to keep the existing API for the moment and keep the changes in this PR to a minimum. In the future we might decide that it's not necessary.

With these changes, any models should implement `KeyEncryptable`/`KeyDecryptable` only, and `LocateKey` if applicable, but never `Encryptable`/`Decryptable` directly.

Some of these changes will be more fleshed out in the future to avoid cluttering this PR, so I expect this won't look perfect:
- Both `LocateKey` and `Encryptable` have an interface like `enc: &EncryptionSettings, org_id: &Option<Uuid>`. The organization parameter is mostly superfluous now and will be removed in the future: Objects that implement `LocateKey` do not use it, while objects which don't implement it should be provided the keys directly.
- `EncString` is implementing `LocateKey` now, which goes against the explanation above. This is to avoid refactoring some secrets manager code, which I'll do on a later PR.
-  `EncryptionSettings` will change in the future, specially the `get_key` implementation, to be more generic. It also contains an `encrypt` function that shouldn't be there, but is used in a couple of spots of the code base and I'll remove on a later PR.
- `KeyEncryptable`/`KeyDecryptable` only handles `SymmetricCryptoKey`, this will be made more generic over key type.


crates/bitwarden/src/crypto/key_encryptable.rs
MGibson1, line 1: **nit:** the `with_key` suffix on all these methods isn't needed, it's implied by the presence of a required `SymmetricCryptoKey` in the parameters (and by the trait name, but hopefully that's temporary)

crates/bitwarden/src/crypto/encryptable.rs
```diff
@@ -12,15 +27,17 @@ pub trait Decryptable<Output> {
     fn decrypt(&self, enc: &EncryptionSettings, org_id: &Option<Uuid>) -> Result<Output>;
 }
 
-impl<T: Encryptable<Output>, Output> Encryptable<Option<Output>> for Option<T> {
-    fn encrypt(self, enc: &EncryptionSettings, org_id: &Option<Uuid>) -> Result<Option<Output>> {
-        self.map(|e| e.encrypt(enc, org_id)).transpose()
+impl<T: KeyEncryptable<Output> + LocateKey, Output> Encryptable<Output> for T {
```
MGibson1, line 30: nice

MGibson1 (commented):
This looks good, nothing pressing here. Let me know you're thoughts on these small nits and we can move forward

crates/bitwarden/src/client/encryption_settings.rs
```diff
@@ -134,18 +120,17 @@ impl EncryptionSettings {
...
 
         let test_string = "encrypted_test_string".to_string();
-        let cipher = test_string.clone().encrypt(&settings, &None).unwrap();
+        let cipher = test_string.clone().encrypt_with_key(&key).unwrap();
 
-        let decrypted_str = cipher.decrypt(&settings, &None).unwrap();
+        let decrypted_str: String = cipher.decrypt_with_key(&key).unwrap();
         assert_eq!(decrypted_str, test_string);
     }
 }
```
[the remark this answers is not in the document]
dani-garcia: Moved over in bb358a38fb203e307571c1aa50d9562b01b60562

MGibson1 (approved):

PR #298, opened 2023-10-20
Map from Requests to Domain models

Hinton:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Implements many of the translators from Request models to Domain models. Wires us up for hooking-up the sync response handling.

## Before you submit


crates/bitwarden/src/vault/cipher/attachment.rs
```diff
@@ -39,7 +39,7 @@ impl KeyEncryptable<Attachment> for AttachmentView {
             size: self.size,
             size_name: self.size_name,
             file_name: self.file_name.encrypt_with_key(key)?,
-            key: self.key.encrypt_with_key(key)?,
+            key: self.key.map(|k| k.encrypt_with_key(key)).transpose()?,
```
dani-garcia, line 42: Encryptable is implemented for &[u8], so you should be able to do something like this:
```suggestion
            key: self.key.as_deref().encrypt_with_key(key)?,
```

crates/bitwarden/src/crypto/enc_string.rs
```diff
@@ -165,6 +165,12 @@ impl FromStr for EncString {
 }
 
 impl EncString {
+    /// Synthetic sugar for mapping `Option<String>` to `Result<Option<EncString>>`
+    #[cfg(feature = "internal")]
+    pub(crate) fn try_from(s: Option<String>) -> Result<Option<EncString>, Error> {
```
dani-garcia, line 170: Should we name this something like try_from_optional so it can't be confused with the actual TryFrom trait?

crates/bitwarden/src/platform/sync.rs
```diff
@@ -80,20 +90,35 @@ impl SyncResponse {
 
         Ok(SyncResponse {
             profile: ProfileResponse::process_response(profile, enc)?,
+            folders: response
```
dani-garcia, line 93: This function is doing the map, into_iter, try_into, collect dance a lot, we can refactor that to a function like this, what do you think?

```rust
        fn try_into_iter<In, InItem, Out, OutItem>(iter: In) -> Result<Out, InItem::Error>
        where
            In: IntoIterator<Item = InItem>,
            InItem: TryInto<OutItem>,
            Out: FromIterator<OutItem>,
        {
            iter.into_iter().map(|i| i.try_into()).collect()
        }

        Ok(SyncResponse {
            profile: ProfileResponse::process_response(profile, enc)?,
            folders: try_into_iter(response.folders.ok_or(Error::MissingFields)?)?,
            collections: try_into_iter(response.collections.ok_or(Error::MissingFields)?)?,
            ciphers: try_into_iter(ciphers)?,
            domains: response.domains.map(|d| (*d).try_into()).transpose()?,
            policies: try_into_iter(response.policies.ok_or(Error::MissingFields)?)?,
            sends: try_into_iter(response.sends.ok_or(Error::MissingFields)?)?,
        })
```

Hinton: I like it.

dani-garcia (approved):

PR #305, opened 2023-10-24
[BEEEP] - Docker image for `bws`

tangowithfoxtrot:
## Type of change

Add a `bws` Dockerfile.

- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

This allows us to publish a `bitwarden/bws` docker image. Multi-arch images can be built with the following command:

`docker buildx build -f ./crates/bws/Dockerfile --push -t bitwarden/bws --platform linux/amd64,linux/arm64 .`

## Code changes

- **`Dockerfile`:** add file

## Before you submit


.github/workflows/build-cli-docker.yml
```diff
@@ -0,0 +1,166 @@
...
+        type: boolean
+        default: true
+  workflow_dispatch:
+    inputs:
+      sdk_branch:
+        description: "Server branch name to deploy (examples: 'master', 'rc', 'feature/sm')"
+        type: string
+        default: master
+  pull_request:
+    paths:
```
michalchecinski, line 23: Adding `Dockerfile` here would be beneficial to check if changes in dockerfile result in building an image in a PR.

.github/workflows/build-cli-docker.yml
```diff
@@ -0,0 +1,166 @@
+---
+name: Build bws Docker image
+
+on:
+  push:
+    paths:
+      - "crates/bws/**"
```
michalchecinski, line 7: Please add `Dockerfile` here to build new image after changes in `Dockerfile`.

.github/workflows/build-cli-docker.yml
```diff
@@ -0,0 +1,166 @@
...
+  build-docker:
+    name: Build Docker image
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout Repository - workflow_call
+        if: ${{ inputs.is_workflow_call == true }}
+        uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac # v4.0.0
+        with:
+          repository: bitwarden/sdk
+          ref: master
```
michalchecinski, line 40: We don't need to check out different repos on different inputs. The second checkout step is sufficient 😃

```suggestion
```

.github/workflows/build-cli-docker.yml
```diff
@@ -0,0 +1,166 @@
...
+
+      - name: Checkout Repository - workflow_dispatch
+        if: ${{ inputs.is_workflow_call != true }}
```
michalchecinski, line 43:
```suggestion
```

Check for trigger is not needed 😃

.github/workflows/build-cli-docker.yml
```diff
@@ -0,0 +1,166 @@
...
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout Repository - workflow_call
+        if: ${{ inputs.is_workflow_call == true }}
+        uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac # v4.0.0
+        with:
+          repository: bitwarden/sdk
+          ref: master
+
+      - name: Checkout Repository - workflow_dispatch
```
michalchecinski, line 42:
```suggestion
      - name: Checkout Repository
```

.github/workflows/build-cli-docker.yml
michalchecinski, line 7:
```suggestion
      - "crates/bws/**"
      - "Dockerfile"
```

.github/workflows/build-cli-docker.yml
```diff
@@ -0,0 +1,166 @@
...
+    paths:
+      - "crates/bws/**"
+  workflow_call:
+    inputs:
+      sdk_branch:
+        type: string
+        default: master
+      is_workflow_call:
+        type: boolean
+        default: true
```
michalchecinski, line 15: Do you plan to trigger this workflow from another one? If not, this is not needed 😃

```suggestion
```

michalchecinski (changes requested):

michalchecinski (dismissed):

.github/workflows/build-cli-docker.yml
```diff
@@ -0,0 +1,153 @@
...
+
+jobs:
+  build-docker:
+    name: Build Docker image
+    runs-on: ubuntu-22.04
+    steps:
+
+      - name: Checkout Repository
+        uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac # v4.0.0
+
```
michalchecinski, line 32:
```suggestion
      - name: Check Branch to Publish
        env:
          PUBLISH_BRANCHES: "master,rc,hotfix-rc"
        id: publish-branch-check
        run: |
          REF=${GITHUB_REF#/refs/heads/}

          IFS="," read -a publish_branches <<< $PUBLISH_BRANCHES

          if [[ "${publish_branches[*]}" =~ "${REF}" ]]; then
            echo "is_publish_branch=true" >> $GITHUB_ENV
          else
            echo "is_publish_branch=false" >> $GITHUB_ENV
          fi
          
```

.github/workflows/build-cli-docker.yml
```diff
@@ -0,0 +1,167 @@
...
+        uses: docker/setup-qemu-action@e81a89b1732b9c48d79cd809d8d81d79c4647a18 # v2.1.0
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@4b4e9c3e2d4531116a6f8ba8e71fc6e2cb6e6c8c # v2.5.0
+
+      ########## Login to Docker registries ##########
+      - name: Login to Azure - Prod Subscription
+        uses: Azure/login@92a5484dfaf04ca78a94597f4f19fea633851fa2 # v1.4.7
+        with:
+          creds: ${{ secrets.AZURE_CI_SERVICE_PRINCIPAL }}
```
michalchecinski, line 58: Maybe this secret will work? 🤔

```suggestion
          creds: ${{ secrets.AZURE_KV_CI_SERVICE_PRINCIPAL }}
```

Dockerfile
Hinton, line 1: Can we move this dockerfile into `crates/bws`?

Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+FROM ${ARCH}rust:1.73 as builder
+RUN apt-get update && \
+  apt-get install --no-install-recommends -y \
+  ca-certificates \
+  && rm -rf /var/lib/apt/lists/*
+WORKDIR /app
+COPY . .
+WORKDIR /app/crates/bws
+RUN cargo build --release
+FROM ${ARCH}debian:trixie-slim as runner
```
Hinton, line 11: This is the pre-release debian image, which seems like an odd choice.

tangowithfoxtrot: So, I tried using `bookworm-slim` before, but I ran into an SSL issue. However, after re-testing just now, I was able to auth and list secrets. I'll get this switched to `bookworm-slim`.

.github/workflows/build-cli-docker.yml
```diff
@@ -0,0 +1,163 @@
...
+
+env:
+  _AZ_REGISTRY: bitwardenprod.azurecr.io
+
+jobs:
+  build-docker:
+    name: Build Docker image
+    runs-on: ubuntu-22.04
+    steps:
+
```
vgrassia, line 27:
```suggestion
```

.github/workflows/build-cli-docker.yml
```diff
@@ -0,0 +1,163 @@
...
+      - name: Checkout Repository
+        uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac # v4.0.0
+      - name: Check Branch to Publish
```
vgrassia, line 30:
```suggestion

      - name: Check Branch to Publish
```

.github/workflows/build-cli-docker.yml
```diff
@@ -0,0 +1,163 @@
...
+        uses: docker/setup-buildx-action@4b4e9c3e2d4531116a6f8ba8e71fc6e2cb6e6c8c # v2.5.0
+
+      ########## Login to Docker registries ##########
+      - name: Login to Azure - Prod Subscription
+        uses: Azure/login@92a5484dfaf04ca78a94597f4f19fea633851fa2 # v1.4.7
+        with:
+          creds: ${{ secrets.AZURE_PROD_KV_CREDENTIALS }}
+
+      - name: Login to Azure ACR
+        run: az acr login -n bitwardenprod
```
vgrassia, line 59:
```suggestion
        run: az acr login -n ${_AZ_REGISTRY%.azurecr.io}
```

vgrassia (changes requested):

.github/workflows/build-cli-docker.yml
tangowithfoxtrot: @vgrassia, I finally saw this suggestion and applied it. This is ready for review again.

crates/bws/.dockerignore
Hinton, line 1: Realistically we should be able to be much more restrictive in this. And only include the root cargo.toml + `crates` dir. Everything else is unnecessary for building the SDK in the dockerfile.

crates/bws/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+  apt-get install --no-install-recommends -y \
+  ca-certificates \
+  && rm -rf /var/lib/apt/lists/*
+COPY ../../ /app
+WORKDIR /app/crates/bws
+RUN cargo build --release
+FROM ${ARCH}debian:bookworm-slim as runner
+WORKDIR /usr/local/bin
+COPY --from=builder /app/target/release/bws .
+COPY --from=builder /etc/ssl/certs /etc/ssl/certs
```
Hinton, line 13: I think it's clearer which stage is active if there is a newline between the stages.

Is it possible to remove the WORKDIR instruction? I think it's more obvious what's going on if the copy includes the full path. I'm also curious if `/usr/local/bin` is the correct path for the binary, I've put it in the docker root in other projects I've worked it. Admittedly they haven't been CLI's though.

```suggestion
RUN cargo build --release

FROM ${ARCH}debian:bookworm-slim as runner
COPY --from=builder /app/target/release/bws /usr/local/bin/bws
COPY --from=builder /etc/ssl/certs /etc/ssl/certs
```

tangowithfoxtrot: Anywhere in `$PATH` is fine: `/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`. I usually install stuff to `/usr/local/bin` because it's where binaries that aren't included with the distribution would typically go (aside from `/opt`, which isn't in `$PATH` by default).

Copying it to `/` would mean that if you overrode the entrypoint for an interactive session, you'd have to specify the full path to the binary since it wouldn't be in `$PATH`.

crates/bws/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+ENTRYPOINT ["bws"]
```
vgrassia, line 15:
```suggestion
###############################################
#                 Build stage                 #
###############################################
FROM --platform=$BUILDPLATFORM rust:1.73 AS build

# Docker buildx supplies the value for this arg
ARG TARGETPLATFORM

RUN apt-get update && apt-get install -y --no-install-recommends \
  ca-certificates \
  && rm -rf /var/lib/apt/lists/*

# Copy required project files
COPY . /app

# Build project
WORKDIR /app/crates/bws
RUN cargo build --release

###############################################
#                  App stage                  #
###############################################
FROM debian:bookworm-slim

ARG TARGETPLATFORM
LABEL com.bitwarden.product="bitwarden"

# Copy built project from the build stage
WORKDIR /usr/local/bin
COPY --from=build /app/target/release/bws .
COPY --from=build /etc/ssl/certs /etc/ssl/certs

ENTRYPOINT ["bws"]

```

vgrassia (changes requested):

.github/workflows/build-cli-docker.yml
```diff
@@ -0,0 +1,163 @@
...
+          fi
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@3b5e8027fcad23fda98b2e3ac259d8d67585f671 # v3.2.0
+        with:
+          context: .
+          file: crates/bws/Dockerfile
+          platforms: |
+            linux/amd64,
+            linux/arm64/v8
```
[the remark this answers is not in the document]
tangowithfoxtrot: The only architectures I've gotten to succeed are `arm64/v8` and `amd64`. Other arches fail: https://github.com/bitwarden/sdk/pull/305/commits/b5d1dc14e3ad00903604e06cb4610133c040e0c2

Failed run with `arm64/v7`: https://github.com/bitwarden/sdk/actions/runs/6709219540/job/18231777228

crates/bws/.dockerignore
tangowithfoxtrot: I've applied Vince's newer, more restrictive, dockerignore suggestion.

michalchecinski (dismissed):
LGTM. @vgrassia since you added few suggestions please also check if everything is good for you 😃

vgrassia (approved):

vgrassia (approved):

PR #314, opened 2023-10-30
Remove encrypt from EncryptionSettings

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
After #297, there were some places in the secrets manager parts of the code that used EncryptionSettings to encrypt directly, instead of dealing with the Encryptable/KeyEncryptable trait. This PR removes those uses so that all encryption opearations have to go through Encryptable/KeyEncryptable.

### DIFF
--- a/crates/bitwarden/src/secrets_manager/projects/create.rs
+++ b/crates/bitwarden/src/secrets_manager/projects/create.rs
@@ -29,7 +29,7 @@ pub(crate) async fn create_project(
         .ok_or(Error::VaultLocked)?;
 
     let project = Some(ProjectCreateRequestModel {
-        name: input.name.clone().encrypt_with_key(&key)?.to_string(),
+        name: input.name.clone().encrypt_with_key(key)?.to_string(),
     });
 
     let config = client.get_api_configurations().await;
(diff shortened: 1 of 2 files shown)


crates/bitwarden/src/secrets_manager/secrets/update.rs
```diff
@@ -24,14 +28,15 @@ pub(crate) async fn update_secret(
...
+        .get_key(&Some(input.organization_id))
+        .ok_or(Error::VaultLocked)?;
 
     let secret = Some(SecretUpdateRequestModel {
-        key: enc.encrypt(input.key.as_bytes(), &org_id)?.to_string(),
-        value: enc.encrypt(input.value.as_bytes(), &org_id)?.to_string(),
-        note: enc.encrypt(input.note.as_bytes(), &org_id)?.to_string(),
+        key: input.key.clone().encrypt_with_key(key)?.to_string(),
+        value: input.value.clone().encrypt_with_key(key)?.to_string(),
+        note: input.note.clone().encrypt_with_key(key)?.to_string(),
```
Hinton, line 39: This makes me doubt that having encryptable interface consuming the string was correct 😩. @MGibson1 

Would `to_owned()` be more appropriate than `clone`?

dani-garcia: There is `to_owned()` and `to_string()`, either would work too, yeah.

Ultimately if we want the `Encryptable` interface to consume the decrypted items, we need to change the APIs to take the requests by value and not by reference, which is a breaking change, otherwise at some point we're going to need to copy the values to have ownership of them.

Hinton (approved):

PR #316, opened 2023-10-31
PHP SDK implementation

milost77:
## Type of change

Implemented PHP library that wraps native C library and exposed its commands through BitwardenClient class.

```
- [ ] Bug fix
- [ x] New feature development
- [ ] Other
```

## Objective

To provide PHP interface for bitwarden c library files by which you can use PHP code to work with Bitwarden API. It implements CRUD requests for projects and secrets. 

## Code changes

TODO: Updating package repository - will hosted on Packagist.


.github/workflows/publish-php.yml
```diff
@@ -0,0 +1,73 @@
...
+          mkdir -p languages/php/src/lib/macos-x64
+          mkdir -p languages/php/src/lib/windows-x64
+
+          platforms=("macos-arm64" "ubuntu-x64" "macos-x64" "windows-x64")
+          files=("libbitwarden_c.dylib" "libbitwarden_c.so" "libbitwarden_c.dylib" "bitwarden_c.dll")
+
+          for ((i=0; i<${#platforms[@]}; i++)); do
+            cp "temp/${platforms[$i]}/${files[$i]}" "languages/php/src/lib/${platforms[$i]}/${files[$i]}"
+          done
+        shell: bash
```
mimartin12, line 70: The default shell for actions is bash, there shouldn't be a need to specify it explicitly here.

mimartin12 (changes requested):

.github/workflows/publish-php.yml
milost77: removed it

languages/php/src/BitwardenSDK.php
```diff
@@ -0,0 +1,67 @@
...
+
+        $this->commandRunner = new CommandRunner($this->bitwarden_lib, $this->handle);
+        $this->projectsClient = new ProjectsClient($this->commandRunner);
+        $this->secretsClient = new SecretsClient($this->commandRunner);
+    }
+
+    /**
+     * @throws \Exception
+     */
+    public function authorize(string $access_token)
```
dani-garcia, line 45: Can we rename this to `accessTokenLogin` to keep the API as close to the other bindings?

languages/php/src/BitwardenSDK.php
```diff
@@ -0,0 +1,67 @@
...
+
+class BitwardenSDK
+{
+    private BitwardenLib $bitwarden_lib;
+
+    private ClientSettings $clientSettings;
+
+    public ProjectsClient $projectsClient;
+
+    public SecretsClient $secretsClient;
```
dani-garcia, line 20: Can we create functions to return these to keep the API similar to the others?
```
$bitwarden_client->secrets()->list(...)
```

languages/php/src/BitwardenSDK.php
```diff
@@ -0,0 +1,67 @@
...
+namespace Bitwarden\Sdk;
+
+use Bitwarden\Sdk\Schemas\AccessTokenLoginRequest;
+use Bitwarden\Sdk\schemas\ClientSettings;
+use Bitwarden\Sdk\Schemas\Command;
+use FFI;
+use Swaggest\JsonDiff\Exception;
+
+
+class BitwardenSDK
```
dani-garcia, line 12: Can we rename this to `BitwardenClient` to keep inline with the other bindings?

languages/php/example.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+require_once 'vendor/autoload.php';
+
+$access_token = '<you access token here>';
+$bitwarden_sdk = new \Bitwarden\Sdk\BitwardenSDK();
+$res = $bitwarden_sdk->authorize($access_token);
+
+// get project
+$res = $bitwarden_sdk->projectsClient->get("00056058-cc70-4cd2-baea-b0810134a729");
```
dani-garcia, line 10: It would be nice if the example could be run without modifying the code too much, by that I mean:
Make the user set the organization id just like we do with the access_token:
```
$access_token = '<your access token here>';
$organization_id = '<your organization id here>';
```

And then organize the requests in such a way that we don't require anything to exist. By that I mean:
- First create a project with name 'php test project' or whatever, and store it's id from the response
- Then list all projects
- Then create a secret based on the previous project id
- Then list all secrets
- Then delete the secret and project

Not a huge priority though.

languages/php/src/BitwardenSDK.php
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        if ($result->authenticated == False) {
+            throw new \Exception("Unauthorized");
+        }
+
+        return $result;
+    }
+
+    public function free_mem()
```
dani-garcia, line 63: Would this be better using a destructor?
https://www.php.net/manual/en/language.oop5.decon.php#language.oop5.decon.destructor

languages/php/src/BitwardenLib.php
```diff
@@ -0,0 +1,79 @@
...
+            }
+        } elseif (PHP_OS === 'Darwin') {
+            $architecture = trim(exec('uname -m'));
+            if ($architecture === 'x86_64' || $architecture === 'amd64') {
+                $lib_file = __DIR__.'/lib/macos-x64/libbitwarden_c.dylib';
+            } elseif ($architecture === 'arm64') {
+                $lib_file = __DIR__.'/lib/macos-arm64/libbitwarden_c.dylib';
+            }
+            if (file_exists($lib_file) == false) {
+                $lib_file = __DIR__.'../../../../target/debug/libbitwarden_c.dylib';
```
dani-garcia, line 42: I think these directories are going one directory too high and should instead be:
```suggestion
                $lib_file = __DIR__.'/../../../target/debug/libbitwarden_c.dylib';
```

Otherwise we're looking for the file in a directory outside the sdk repo.

The same applies to the other operating systems.

languages/php/src/BitwardenSDK.php
milost77: renamed

languages/php/src/BitwardenSDK.php
milost77: renamed properties, ie, secrets and projects used.

languages/php/src/BitwardenSDK.php
milost77: renamed

languages/php/example.php
milost77: I inserted the organization as var as well. Also made similar kind of order so that nothing will be created or delete that exists outside of the example file.

languages/php/src/BitwardenSDK.php
milost77: good idea, implemented.

languages/php/src/CommandRunner.php
```diff
@@ -0,0 +1,38 @@
...
+        $result = $this->bitwardenLib->run_command($command);
+        if (isset($result->data)) {
+            if ($result->success == true) {
+                return $result->data;
+            }
+            if (isset($result->error))
+            {
+                throw new \Exception($result->error);
+            }
+        }
```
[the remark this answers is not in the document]
milost77: thanks, corrected.

languages/php/src/BitwardenLib.php
milost77: thank you, this was a mistake on my part.

languages/php/src/BitwardenSDK.php
```diff
@@ -0,0 +1,67 @@
...
+
+    public ProjectsClient $projectsClient;
+
+    public SecretsClient $secretsClient;
+
+    private CommandRunner $commandRunner;
+
+    private FFI\CData $handle;
+
+    public function __construct()
```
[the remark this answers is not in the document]
milost77: Changed so we can pass ClientSettings to client.

languages/php/src/BitwardenSDK.php
dani-garcia: Sorry, maybe I wasn't completely clear, I meant the function name, so something like this (had to check that PHP uses underscores for function names, so we should follow that pattern):
```php
    public function access_token_login(string $access_token)
```

languages/php/src/BitwardenClient.php
```diff
@@ -0,0 +1,67 @@
...
+
+    private FFI\CData $handle;
+
+    public function __construct(ClientSettings $clientSettings)
+    {
+        $this->clientSettings = $clientSettings;
+        $this->clientSettings->apiUrl = getenv('API_URL') ?: 'https://api.bitwarden.com';
+        $this->clientSettings->identityUrl = getenv('IDENTITY_URL') ?: 'https://identity.bitwarden.com';
+        $this->clientSettings->userAgent = getenv('USER_AGENT') ?: 'SDK';
+        $this->clientSettings->deviceType = getenv('DEVICE_TYPE') ?: 'SDK';
```
dani-garcia, line 32: Now that we're passing the clientSettings as a parameter we don't need to set these here, so these lines can be removed.

The user should be the one in charge of setting these values, like the code in the readme shows.

milost77: It's removed now. Sorry, I was swayed by implementations in other languages.

.github/workflows/publish-php.yml
```diff
@@ -0,0 +1,75 @@
...
+
+          platforms=("macos-arm64" "ubuntu-x64" "macos-x64" "windows-x64")
+          files=("libbitwarden_c.dylib" "libbitwarden_c.so" "libbitwarden_c.dylib" "bitwarden_c.dll")
+
+          for ((i=0; i<${#platforms[@]}; i++)); do
+            cp "temp/${platforms[$i]}/${files[$i]}" "languages/php/src/lib/${platforms[$i]}/${files[$i]}"
+          done
+
+      - name: Publish version
+        run: curl -XPOST -H'content-type:application/json' 'https://packagist.org/api/update-package?username=malirobot&apiToken=${{secrets.PACKAGIST_KEY}}' -d'{"repository":{"url":"https://packagist.org/packages/bitwarden/sdk"}}'
```
dani-garcia, line 74: @mimartin12 Should this be updated with some Bitwarden managed credentials, or do we leave that for after the merge?

dani-garcia (approved):
Code changes LGTM

languages/php/README.md
```diff
@@ -0,0 +1,66 @@
...
+// get secret
+$res = $bitwarden_sdk->secrets->get("75d3a7ff-30ed-433a-91aa-b099016e4833");
+// list secrets
+$res = $bitwarden_sdk->secrets->list("5688da1f-cc25-41d7-bb9f-b0740144ef1d");
+// create secret
+$res = $bitwarden_sdk->secrets->create("New Key", "hello world", "5688da1f-cc25-41d7-bb9f-b0740144ef1d", ["b23818dd-827b-4a22-b97a-b07e010ae9d4"], "123");
+// update secret
+$res = $bitwarden_sdk->secrets->update("901d102d-af7d-46a1-99f5-b0a6017e2f07", "hello world 2", "hello", "5688da1f-cc25-41d7-bb9f-b0740144ef1d", ["b23818dd-827b-4a22-b97a-b07e010ae9d4"], "123");
+// delete secret
+$res = $bitwarden_sdk->secrets->delete(["380b5c30-d8fc-472d-a514-b0ac00f17071"]);
```
dani-garcia, line 62: Can we replace this example code in the readme that hardcodes uuids by the code we have in `example.php`?

.github/workflows/publish-php.yml
milost77: What we can do is that I can add someone from Bitwarden as a collaborator on packagist, then the access to the key won't be a problem (it is used in action secrets). The other scenario is to make a completely new package, the only potential problem, one that I haven't thought about before, is that the name of the package is reserved to "bitwarden/sdk", so it would need to be named differently.

languages/php/README.md
milost77: I changed it to be similar as in example.php

languages/php/src/schemas/Kdf.php
Hinton, line 1: We can delete this file since we currently don't support this request.

languages/php/src/schemas/PasswordLoginRequest.php
Hinton, line 1: We can delete this file since we currently don't support this request.

languages/php/src/schemas/SyncRequest.php
Hinton, line 1: We can delete this file since we currently don't support this request.

languages/php/src/schemas/TwoFactorRequest.php
Hinton, line 1: We can delete this file since we currently don't support this request.

Hinton (commented):
Some nits.

.gitignore
```diff
@@ -35,6 +35,9 @@ crates/bitwarden-napi/sdk-napi.*.node
 # Complied TypeScript client
 crates/bitwarden-napi/dist
 
+# PHP
+languages/php/vendor
```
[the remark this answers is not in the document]
milost77: removed

languages/php/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+To interact with the client first you need to obtain the access token from Bitwarden.
+You can initialize ClientSettings and its setting before passing it to the BitwardenClient.
+
+```php
+$client_settings = new \Bitwarden\Sdk\Schemas\ClientSettings()
+$client_settings->apiUrl = getenv('API_URL') ?: 'https://api.bitwarden.com';
+$client_settings->identityUrl = getenv('IDENTITY_URL') ?: 'https://identity.bitwarden.com';
+$client_settings->userAgent = getenv('USER_AGENT') ?: 'SDK';
+$client_settings->deviceType = getenv('DEVICE_TYPE') ?: 'SDK';
```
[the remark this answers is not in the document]
milost77: I made BitwardenSettings class for use instead of client settings. It hanldes default values like provided code sample and handles blank device type.

languages/php/src/schemas/FingerprintRequest.php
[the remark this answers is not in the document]
milost77: deleted

languages/php/src/schemas/Kdf.php
milost77: deleted

languages/php/src/schemas/KdfPBKDF2.php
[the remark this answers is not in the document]
milost77: deleted

languages/php/src/schemas/PasswordLoginRequest.php
milost77: deleted

languages/php/src/schemas/TwoFactorRequest.php
milost77: deleted

languages/php/src/BitwardenSettings.php
```diff
@@ -0,0 +1,54 @@
...
+        } else {
+            $this->api_url = $api_url;
+        }
+
+        if (is_null($identity_url))
+        {
+            $this->identity_url = getenv('IDENTITY_URL') ?: 'https://identity.bitwarden.com';
+        } else {
+            $this->identity_url = $identity_url;
+        }
```
Hinton, line 30: In the C# project we decided against falling back to env variables, I think we should keep that pattern. Especially since `API_URL` and `IDENTITY_URL` are generic.

I believe you can just have them default to null, the SDK should handle it appropriately since they are nullable.

languages/php/README.md
```diff
@@ -16,22 +16,23 @@ If you are not using the standalone version of this library, file will be placed
...
 
 ```php
-$access_token = '<you access token here>';
+$api_url = "";
+$identity_url = "";
+$client_settings = new \Bitwarden\Sdk\BitwardenSettings();
+
 $bitwarden_client = new \Bitwarden\Sdk\BitwardenClient($client_settings);
-$result = $bitwarden_client->access_token_login($access_token);
+$res = $bitwarden_client->access_token_login($access_token);
```
Hinton, line 35: `access_token_login` should swallow the result similar to the c# implementation.

languages/php/src/BitwardenSettings.php
```diff
@@ -0,0 +1,54 @@
...
+
+class BitwardenSettings
+{
+    private string $api_url;
+
+    private string $identity_url;
+
+    private string $user_agent = "Bitwarden PHP SDK";
+
+    private string $device_type;
```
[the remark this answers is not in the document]
milost77: I removed them, leaving just api_url and identity_url attributes.

languages/php/src/BitwardenSettings.php
milost77: Yes, it works even when they are null, so I don't use any default values now.

languages/php/src/BitwardenClient.php
```diff
@@ -23,9 +23,11 @@ class BitwardenClient
 
     private FFI\CData $handle;
 
-    public function __construct(ClientSettings $clientSettings)
+    public function __construct(BitwardenSettings $bitwardenSettings)
     {
-        $this->clientSettings = $clientSettings;
+        $this->clientSettings = new ClientSettings();
+        $this->clientSettings->apiUrl = $bitwardenSettings->get_api_url();
+        $this->clientSettings->identityUrl = $bitwardenSettings->get_identity_url();
```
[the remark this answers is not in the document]
milost77: changed to use it

languages/php/README.md
milost77: changes so that access_token_login doesn't return anything

languages/php/src/BitwardenSettings.php
```diff
@@ -0,0 +1,54 @@
...
+
+    public function __construct($api_url = null, $identity_url = null, $device_type = null)
```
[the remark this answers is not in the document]
milost77: removed

dani-garcia (dismissed):

Hinton (dismissed):
LGTM

.github/workflows/publish-php.yml
mimartin12: What @Hinton is suggesting sounds like a good way to go about it, especially since you will want to preserve the package name. I am not familiar with Packagist, but it looks like their [API tokens](https://packagist.org/about) are scoped to the user account. That means `${{secrets.PACKAGIST_KEY}}` will need to be updated when you do the switch.
image

mimartin12 (approved):
Workflow LGTM.

PR #317, opened 2023-11-02
Extract JWT to it's own file

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Move JWT handling from utils to auth. Also improve documentation and use `FromStr` instead of a custom function.

## Before you submit

### DIFF
--- a/crates/bitwarden/src/auth/jwt_token.rs
+++ b/crates/bitwarden/src/auth/jwt_token.rs
@@ -1,0 +1,68 @@
+use std::str::FromStr;
+
+use base64::Engine;
+
+use crate::{error::Result, util::BASE64_ENGINE};
+
+/// A Bitwarden secrets manager JWT Token.
+///
+/// References:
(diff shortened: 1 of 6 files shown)


crates/bitwarden/src/auth/jwt_token.rs
```diff
@@ -0,0 +1,68 @@
...
+
+/// A Bitwarden secrets manager JWT Token.
+///
+/// References:
+/// - <https://github.com/bitwarden/server/blob/419760623a7af5f5d0cbdfeb2a7aba8c3608d880/src/Identity/IdentityServer/ClientStore.cs#L125-L126>
+/// - <https://github.com/bitwarden/server/blob/419760623a7af5f5d0cbdfeb2a7aba8c3608d880/src/Identity/IdentityServer/ClientStore.cs#L133>
+///
+/// TODO: We need to expand this to support user based JWT tokens.
+#[derive(serde::Deserialize)]
+pub struct JWTToken {
```
dani-garcia, line 15: I think we should add a quick mention, either here or on the FromStr impl that this doesn't validate the JWT signature, and it's only meant to be used on tokens that we trust coming from the server.

dani-garcia (approved):

PR #318, opened 2023-11-02
Begin documenting the crypto module

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Begin documenting the crypto module

## Before you submit

### DIFF
--- a/crates/bitwarden/src/crypto/enc_string.rs
+++ b/crates/bitwarden/src/crypto/enc_string.rs
@@ -11,6 +11,45 @@ use crate::{
 
 use super::{KeyDecryptable, KeyEncryptable, LocateKey};
 
+/// # Encrypted string primitive
+///
+/// [EncString] is a Bitwarden specific primitive that represents an encrypted string. They are
+/// are used together with the [KeyDecryptable] and [KeyEncryptable] traits to encrypt and decrypt
+/// data using [SymmetricCryptoKey]s.
+///
+/// The flexibility of the [EncString] type allows for different encryption algorithms to be used
+/// which is represented by the different variants of the enum.
+///
+/// ## Note
+///
+/// We are currently in the progress of splitting the [EncString] into distinct AES and RSA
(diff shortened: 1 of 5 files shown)


crates/bitwarden/src/crypto/enc_string.rs
```diff
@@ -11,6 +11,45 @@ use crate::{
...
+/// [EncString] implements [Display] and [FromStr] to allow for easy serialization and uses a
+/// custom scheme to represent the different variants.
+///
+/// The scheme is one of the following schemes:
+/// - `[type].[iv]|[data]`
+/// - `[type].[iv]|[data]|[mac]`
+/// - `[type].[data]`
+///
+/// Where:
+/// - `[type]` is a single digit number representing the variant
```
dani-garcia, line 49: It's only a single digit number at the moment, if we ever add support for chacha20 or the like, we might need to expand to two digits, as we only have three values left. Maybe it's better to not specify the length?

crates/bitwarden/src/crypto/fingerprint.rs
```diff
@@ -7,6 +7,11 @@ use crate::{
     wordlist::EFF_LONG_WORD_LIST,
 };
 
+/// Computes a fingerprint of the given `fingerprint_material` using the given `public_key`.
+///
+/// This is commonly used for account fingerprints. With the following arguments:
+/// * `fingerprint_material` - users id.
```
dani-garcia, line 13:
```suggestion
/// * `fingerprint_material` - user's id.
```

crates/bitwarden/src/crypto/fingerprint.rs
```diff
@@ -7,6 +7,11 @@ use crate::{
...
+/// * `public_key` - users public key.
```
dani-garcia, line 14:
```suggestion
/// * `public_key` - user's public key.
```

dani-garcia (approved):

PR #319, opened 2023-11-02
Move login methods to auth

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Move login methods under `auth()` to clean up the core client.

## Before you submit


crates/bitwarden/src/auth/client_auth.rs
```diff
@@ -1,15 +1,27 @@
...
+    client::kdf::Kdf,
 };
-use crate::{client::kdf::Kdf, error::Result, Client};
 
 pub struct ClientAuth<'a> {
     pub(crate) client: &'a mut crate::Client,
 }
 
 impl<'a> ClientAuth<'a> {
+    #[cfg(feature = "internal")]
```
dani-garcia, line 24: I'm wondering if it would be easier for us to just have two `impl<'a> ClientAuth<'a> {` blocks and put the feature flag directly on the block instead of repeating it all over the place. Not really necessary for this PR though

crates/bitwarden/src/auth/client_auth.rs
```diff
@@ -37,14 +51,152 @@ impl<'a> ClientAuth<'a> {
...
+    pub async fn prelogin(&mut self, email: String) -> Result<Kdf> {
+        use crate::auth::login::request_prelogin;
+
+        request_prelogin(self.client, email).await?.try_into()
+    }
+
+    #[cfg(feature = "internal")]
+    pub async fn login_password(
+        &mut self,
+        input: &PasswordLoginRequest,
```
dani-garcia, line 73: Do we need to rename the structs to match as well?

Hinton: We expose the structs externally, so I would prefer to not make that change right now.

dani-garcia (approved):

PR #320, opened 2023-11-03
[BEEEP] SM-1005 - Add env output option

tangowithfoxtrot:
## Type of change

- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

Add an env output option to `bws`. This allows for easier command-line usage in scripts, particularly where a JSON or YAML parser is not desirable or available.

Basic usage examples in Bash:

- `source <(bws secret get ec9e0489-244e-4b3f-8782-b0a800fe562f -o env)`
- `bws secret list -o env > .env`

## Code changes

- **`crates/bws/src/render.rs`:** Output secrets in `key="value"` format

## Screenshots

image

## Before you submit

### DIFF
--- a/crates/bws/src/render.rs
+++ b/crates/bws/src/render.rs
@@ -62,7 +62,7 @@ pub(crate) fn serialize_response<T: Serialize + TableSerialize<N>, const N: usiz
                         format!("{}=\"{}\"", row[1], row[2])
                     } else {
                         commented_out = true;
-                        format!("# {}=\"{}\"", row[1], row[2].replace("\n", "\n# "))
+                        format!("# {}=\"{}\"", row[1], row[2].replace('\n', "\n# "))
                     }
                 })
                 .collect();
--- a/crates/bws/src/render.rs
+++ b/crates/bws/src/render.rs
@@ -51,16 +51,14 @@ pub(crate) fn serialize_response<T: Serialize + TableSerialize<N>, const N: usiz
             pretty_print("yaml", &text, color);
         }
         Output::Env => {
+            let valid_key_regex = regex::Regex::new("^[a-zA-Z_][a-zA-Z0-9_]*$").unwrap();
+
             let mut commented_out = false;
             let mut text: Vec<String> = data
                 .get_values()
                 .into_iter()
                 .map(|row| {
-                    if row[1]
-                        .chars()
-                        .all(|c| c.is_ascii_alphanumeric() || c == '_')
-                        && !row[1].chars().next().unwrap().is_digit(10)
-                    {
+                    if valid_key_regex.is_match(&row[1]) {
                         format!("{}=\"{}\"", row[1], row[2])
                     } else {
                         commented_out = true;
--- a/crates/bws/src/render.rs
+++ b/crates/bws/src/render.rs
@@ -9,6 +9,7 @@ use serde::Serialize;
 pub(crate) enum Output {
     JSON,
     YAML,
+    Env,
     Table,
     TSV,
     None,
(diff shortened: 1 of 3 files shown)


dani-garcia:
Nice! A couple of small things:
- Not sure how this is usually done by other clients but we may need to escape characters in the key and the value, at the moment you can create a key like `'key="still the key"` which is probably going to cause issues.
- Should we add the note at the end of the line as a comment? something like `key="value" # note`? Not sure if that's supported correctly in all the env parsers though, and we probably need to remove newlines so maybe it's not worth it.
- I think you can use the `pretty_print` function to get colored output, using `sh` or `bash` as a language.

tangowithfoxtrot:
> Not sure how this is usually done by other clients but we may need to escape characters in the key and the value, at the moment you can create a key like `'key="still the key"` which is probably going to cause issues.

Good point. I hadn't considered that... I'm still relatively new to Rust. What would you recommend to handle the escape characters there?

> Should we add the note at the end of the line as a comment? something like `key="value" # note`? Not sure if that's supported correctly in all the env parsers though, and we probably need to remove newlines so maybe it's not worth it.

I actually had a version of this in my own fork of the CLI, where I didn't have any issues, but I think in the scripts that I wrote that relied on it stripped-out the comments. I'd be fine excluding the notes to avoid dealing with any potential parsing issues that could occur.

> I think you can use the `pretty_print` function to get colored output, using `sh` or `bash` as a language.

Very nice! I'll switch to using that!

tangowithfoxtrot:
> Not sure how this is usually done by other clients but we may need to escape characters in the key and the value, at the moment you can create a key like 'key="still the key" which is probably going to cause issues.

Good point. I hadn't considered that... I'll try to work something out for the escape characters. 

> Should we add the note at the end of the line as a comment? something like key="value" # note? Not sure if that's supported correctly in all the env parsers though, and we probably need to remove newlines so maybe it's not worth it.

I actually had a version of this in my own fork of the CLI, where I didn't have any issues, but I think in the scripts that I wrote that relied on it stripped-out the comments. I'd be fine excluding the notes to avoid dealing with any potential parsing issues that could occur.

> I think you can use the pretty_print function to get colored output, using sh or bash as a language.

Very nice! I'll switch to using that!

tangowithfoxtrot:
@dani-garcia, regarding problematic key names: what do you think of commenting the key-value pair out if it's non-[POSIX-compliant](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_231)? Like this:
image

dani-garcia:
That seems reasonable to me, the comment explaining why they are excluded is a great touch!

crates/bws/src/render.rs
```diff
@@ -49,6 +50,33 @@ pub(crate) fn serialize_response<T: Serialize + TableSerialize<N>, const N: usiz
...
+            let mut commented_out = false;
+            let mut text: Vec<String> = data
+                .get_values()
+                .into_iter()
+                .map(|row| {
+                    if row[1]
+                        .chars()
+                        .all(|c| c.is_ascii_alphanumeric() || c == '_')
+                        && !row[1].chars().next().unwrap().is_digit(10)
+                    {
```
dani-garcia, line 63: I think this would look cleaner if we used a `regex`:

```rust
let valid_key_regex = regex::Regex::new("^[a-zA-Z_][a-zA-Z0-9_]*$").unwrap();

let mut commented_out = false;
let mut text: Vec<String> = data
    .get_values()
    .into_iter()
    .map(|row| {
        if valid_key_regex.is_match(&row[1]) {
```

Just have to add the crate in the bws/Cargo.toml:
```diff
uuid = { version = "^1.3.3", features = ["serde"] }
+regex = { version = "1.10.2", features=["std", "perf"], default-features=false }
```

tangowithfoxtrot: Agree. I applied those changes.

crates/bws/src/render.rs
```diff
@@ -49,6 +50,31 @@ pub(crate) fn serialize_response<T: Serialize + TableSerialize<N>, const N: usiz
...
+            let mut commented_out = false;
+            let mut text: Vec<String> = data
+                .get_values()
+                .into_iter()
+                .map(|row| {
+                    if valid_key_regex.is_match(&row[1]) {
+                        format!("{}=\"{}\"", row[1], row[2])
+                    } else {
+                        commented_out = true;
+                        format!("# {}=\"{}\"", row[1], row[2].replace("\n", "\n# "))
```
dani-garcia, line 65:
```suggestion
                        format!("# {}=\"{}\"", row[1], row[2].replace('\n', "\n# "))
```

Last change now, just to make CI happy 😄

dani-garcia (approved):
Nice, thank you!

PR #330, opened 2023-11-14
[PM-4695] Implement unlock with user key and update examples to use biometrics

dani-garcia:
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
Enable the SDK to export the decrypted user key and expose a way to initialize the SDK with it. The Android and iOS examples have been revamped to allow choosing how to unlock the client.

This depends on the refactor done on #329


crates/bitwarden/src/client/client.rs
```diff
@@ -222,6 +222,20 @@ impl Client {
...
+        &mut self,
+        decrypted_user_key: &str,
+        private_key: EncString,
+    ) -> Result<&EncryptionSettings> {
+        let user_key = decrypted_user_key.parse::<SymmetricCryptoKey>()?;
+        self.encryption_settings = Some(EncryptionSettings::new_decrypted_key(
+            user_key,
+            private_key,
+        )?);
+        Ok(self.encryption_settings.as_ref().unwrap())
```
Hinton, line 236: Do we need to unwrap this instead of properly handling the error?

crates/bitwarden/src/client/encryption_settings.rs
```diff
@@ -45,24 +45,29 @@ impl EncryptionSettings {
...
             }
         }
     }
 
+    #[cfg(feature = "internal")]
+    pub(crate) fn new_decrypted_key(
+        user_key: SymmetricCryptoKey,
+        private_key: EncString,
+    ) -> Result<Self> {
+        // Decrypt the private key with the user key
```
Hinton, line 58: This comment describes the code and is mostly superficial. Can we document `new_decrypted_key` instead?

languages/kotlin/app/src/main/java/com/bitwarden/myapplication/Biometrics.kt
```diff
@@ -0,0 +1,124 @@
...
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.async
+import java.security.KeyStore
+import java.util.Base64
+import javax.crypto.Cipher
+import javax.crypto.KeyGenerator
+import javax.crypto.SecretKey
+import javax.crypto.spec.GCMParameterSpec
+
+class Biometric(private var activity: FragmentActivity) {
```
Hinton, line 23: Can we add a disclaimer to these files that the code is for demonstration purposes only.

Hinton (commented):
Looks good, a few minor comments.

crates/bitwarden/src/client/client.rs
dani-garcia: The value is initialized on the previous line, so it should never happen, but at least using `.expect("Value was initialized on the previous line")` would make that more clear

Hinton (approved):

PR #332, opened 2023-11-16
Generate immutable records in uniffi bindings

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Update the uniffi code generation to output immutable structs. I've pointed the dependencies to my fork until this is included in the main uniffi repo, but it should be enough to let the mobile devs test if the solution is appropiate for them.


Hinton (changes requested):
This is merged in the main uniffi repo, so we can update our references to that instead.

dani-garcia:
That was a quick merge, updated back to main repo

Hinton (approved):

PR #333, opened 2023-11-17
Improve MAC validation

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Change the name of the function `validate_mac` to the more appropiate `generate_mac`, and change the mac comparison to be constant time with the use of the `subtle` crate.

Note that `hmac.finalize()` returns a type `CtOutput` that supports constant time comparison by using the `subtle` crate internally, but `CtOutput` doesn't allow getting a reference to the internal array which would complicate the implementation of `EncString`, so I chose to use the `subtle` constant time comparison directly.


Hinton (commented):
Do you know how much of a performance improvement this actually is?

Could you also add some unit tests to `generate_mac` while at it?

dani-garcia:
I don't think this will change much either way performance wise, if anything it's more of a tiny security improvement. 

Ultimately the goal is more to keep in line with how the `hmac` crate handles the MAC comparisons, where they use a constant time equality check by default on their output types returned from calling `hmac.finalize()`. Sadly we can't just use those directly because we require access to the raw byte values for serialization of the EncStrings.

Hinton (approved):

PR #335, opened 2023-11-17
Add support for missing AES decrypt types

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Add support for the missing AES decryption types: `AesCbc256_B64` and `AesCbc128_HmacSha256_B64`.

Note that `SymmetricCryptoKey` at the moment expects 32 byte keys, and AES128 used 16 byte keys, so the 16byte key+mac get parsed as a single 32byte key. At the moment we just split it by hand, but ultimately we would need to handle this better in a future refactor.

### DIFF
--- a/crates/bitwarden/src/crypto/aes_ops.rs
+++ b/crates/bitwarden/src/crypto/aes_ops.rs
@@ -9,8 +9,10 @@
 //! [KeyDecryptable][super::KeyDecryptable] instead.
 
 use aes::cipher::{
-    block_padding::Pkcs7, generic_array::GenericArray, typenum::U32, BlockDecryptMut,
-    BlockEncryptMut, KeyIvInit,
+    block_padding::Pkcs7,
+    generic_array::GenericArray,
+    typenum::{U16, U32},
+    BlockDecryptMut, BlockEncryptMut, KeyIvInit,
 };
 use hmac::Mac;
 use rand::RngCore;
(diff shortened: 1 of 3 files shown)


Hinton (changes requested):
Can we add some tests to confirm we're decrypting this correctly?

crates/bitwarden-crypto/src/enc_string/symmetric.rs
```diff
@@ -236,13 +236,26 @@ impl KeyEncryptable<SymmetricCryptoKey, EncString> for &[u8] {
...
         match self {
+            EncString::AesCbc256_B64 { iv, data } => {
+                let dec = crate::aes::decrypt_aes256(iv, data.clone(), &key.key)?;
+                Ok(dec)
+            }
+            EncString::AesCbc128_HmacSha256_B64 { iv, mac, data } => {
+                // TODO: SymmetricCryptoKey is designed to handle 32 byte keys only, but this
+                // variant uses a 16 byte key This means the key+mac are going to be
+                // parsed as a single 32 byte key, at the moment we split it manually
+                // When refactoring the key handling, this should be fixed.
```
Hinton, line 247: Do we have a ticket for this?

Hinton (approved):

crates/bitwarden-crypto/src/enc_string/symmetric.rs
dani-garcia: I created a new one in SDK future: PM-6164

PR #340, opened 2023-11-19
AES tests

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

An effort in making the AES operations lower level by removing the `EncString` concept from them. This makes them more easily testable and reusable.

## Code changes

- **crates/bitwarden/src/crypto/aes_ops.rs:** Removed `EncString` and made the low level interfaces return slices and vectors directly. Mocked `rand` and wrote unit tests.
- **crates/bitwarden/src/crypto/enc_string.rs**: Added `EncString::encrypt_aes256_hmac` which behaves similar to `aes_ops::encrypt_aes256_hmac` except it returns an `EncString` instead of the primitives.

## Before you submit

### DIFF
--- a/crates/bitwarden/src/crypto/enc_string.rs
+++ b/crates/bitwarden/src/crypto/enc_string.rs
@@ -1,5 +1,6 @@
 use std::{fmt::Display, str::FromStr};
 
+use aes::cipher::{generic_array::GenericArray, typenum::U32};
 use base64::Engine;
 use serde::{de::Visitor, Deserialize};
 
@@ -331,6 +332,20 @@ impl serde::Serialize for EncString {
 }
 
 impl EncString {
+    pub(crate) fn encrypt_aes256_hmac(
+        data_dec: &[u8],
+        mac_key: GenericArray<u8, U32>,
+        key: GenericArray<u8, U32>,
+    ) -> Result<EncString> {
+        let enc = super::encrypt_aes256_hmac(data_dec, mac_key, key)?;
+
+        Ok(EncString::AesCbc256_HmacSha256_B64 {
+            iv: enc.0,
+            mac: enc.1,
+            data: enc.2,
+        })
+    }
+
     /// The numerical representation of the encryption type of the [EncString].
     const fn enc_type(&self) -> u8 {
         match self {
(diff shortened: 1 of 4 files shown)


crates/bitwarden/src/crypto/enc_string.rs
```diff
@@ -331,6 +332,20 @@ impl serde::Serialize for EncString {
...
+        mac_key: GenericArray<u8, U32>,
+        key: GenericArray<u8, U32>,
+    ) -> Result<EncString> {
+        let enc = super::encrypt_aes256_hmac(data_dec, mac_key, key)?;
+
+        Ok(EncString::AesCbc256_HmacSha256_B64 {
+            iv: enc.0,
+            mac: enc.1,
+            data: enc.2,
+        })
```
dani-garcia, line 346: Small style suggestion:
```suggestion
        let (iv, mac, data) = super::encrypt_aes256_hmac(data_dec, mac_key, key)?;
        Ok(EncString::AesCbc256_HmacSha256_B64 { iv, mac, data })
```

crates/bitwarden/src/crypto/master_key.rs
```diff
@@ -61,8 +61,11 @@ fn make_user_key(
     rng.fill(&mut user_key);
 
     let stretched_key = stretch_master_key(master_key)?;
-    let protected =
-        encrypt_aes256_hmac(&user_key, stretched_key.mac_key.unwrap(), stretched_key.key)?;
```
dani-garcia, line 65: We could use `KeyEncryptable` here: 
```rust
    let protected = user_key.as_slice().encrypt_with_key(&stretched_key)?;
```

crates/bitwarden/src/crypto/rsa.rs
```diff
@@ -33,7 +33,7 @@ pub(super) fn make_key_pair(key: &SymmetricCryptoKey) -> Result<RsaKeyPair> {
         .to_pkcs8_der()
         .map_err(|_| Error::Internal("unable to create private key"))?;
 
-    let protected = encrypt_aes256_hmac(pkcs.as_bytes(), key.mac_key.unwrap(), key.key)?;
```
dani-garcia, line 36: We could also use `KeyDecryptable` here:
```rust
    let protected: EncString = pkcs.as_bytes().encrypt_with_key(key)?;
```

dani-garcia (commented):
In the cases where we use `encrypt_aes256_hmac` outside of `EncString`, we could also use `KeyEncryptable`, and that way we wouldn't need to extend the API.

This has the negative effect that we wouldn't be explicitly using AES256+HMAC, and we depend on `encrypt_with_key` doing the right thing though, which might be an issue if we ever change that default.

dani-garcia:
Yeah, that's fair. Hopefully when I get some time to refactor the SymmetricCryptoKey I can find a way so we can use KeyEncryptable while also being explicit that we need AES256+HMAC

PR #341, opened 2023-11-22
Generate better schema types

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
The `quicktype` schema generation seems to have trouble with types in schema files being referenced from other types in separate schema files, which created duplicate classes, with a lot of strange names.

Instead of generating one individual file for each response, we create a big struct with all the options. This seems to solve the issue of references between schema files, and so no duplicate classes are created.

Note that the way that the file is loaded in `quicktype` is different. By default, `quicktype` generates JSON conversion helper functions only for the root type, which would be the big struct `SchemaTypes`. By importing the file with `#/definitions/` appended, we only bring the sub schemas as root types, and ignore `SchemaTypes`, which we don't care about.

Examples of the schema changes, using C# as an example:
- `ProjectsResponse` previously contained an array of `DatumElement`, which were identical to `ProjectResponse`. Now it properly contains an array of `ProjectResponse`. This is also applicable to `SecretsResponse` which contained an array of `DatumClass`
- Previously it would generate multiple variants for `Argon2ID`, named: `FluffyArgon2Id` and `PurpleArgon2Id`, now there is only one `Argon2Id`. This is applicable to the two factor response types, which change in the same way

This is working correctly with the Java, Go, C++, PHP and Ruby bindings.

### DIFF
--- a/crates/sdk-schemas/src/main.rs
+++ b/crates/sdk-schemas/src/main.rs
@@ -1,8 +1,7 @@
 use std::{fs::File, io::Write};
 
 use anyhow::Result;
-use itertools::Itertools;
-use schemars::{schema::RootSchema, schema_for};
+use schemars::{schema::RootSchema, schema_for, JsonSchema};
 
 /// Creates a json schema file for any type passed in using Schemars. The filename and path of the generated
 /// schema file is derived from the namespace passed into the macro or supplied as the first argument.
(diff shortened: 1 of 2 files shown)


Hinton (approved):
Did you verify the TS types to?

dani-garcia:
I did now and they work correctly, I also created a PR with some of the changes to get the TS bindings up to par with the rest of them https://github.com/bitwarden/sdk/pull/398

PR #344, opened 2023-11-23
Sort cargo.toml files

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Sort `cargo.toml` files alphabetically.

## Before you submit

### DIFF
--- a/crates/bitwarden-c/Cargo.toml
+++ b/crates/bitwarden-c/Cargo.toml
@@ -4,7 +4,6 @@ version = "0.1.0"
 edition = "2021"
 rust-version = "1.57"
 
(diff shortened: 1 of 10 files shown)


dani-garcia:
Do we have a way to add a new dependency at the correct place? Usually I use `cargo add -p bitwarden whatever`, but that doesn't respect the order afaik.

Did you order them manually or use a specific formatter?

Hinton:
I used https://crates.io/crates/cargo-sort `cargo sort --grouped --workspace`

dani-garcia (approved):

PR #369, opened 2023-11-27
Sm 989 create build pipeline for the python language wrapper

tangowithfoxtrot:
## Type of change

- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Package the Python SDK so that it can be published to [PyPI.org](pypi.org).

## Code changes

- **`.github/workflows/build-python-wheels.yml`:** create wheels for Linux, macOS, and Windows (AMD64-only). A sdist is also created for other platforms.
- **`languages/python/BitwardenClient/__init__.py`:** export functions and classes for autocompletion in IDEs and the interactive Python interpreter
- **`languages/python/pyproject.toml`**: replaces `setup.py`, which is no longer officially supported. This uses Maturin to build the Rust shared-object library that will be bundled with the SDK.
- **`languages/python/setup.py`**: replaced by `pyproject.toml`

## Before you submit


.github/workflows/build-python-wheels.yml
```diff
@@ -0,0 +1,121 @@
...
+        with:
+          name: BitwardenClient-${{ env._PACKAGE_VERSION }}-${{ matrix.settings.target }}
+          path: ${{ github.workspace }}/target/wheels/BitwardenClient*.whl
+
+      - name: Upload sdists
+        if: ${{ matrix.settings.target == 'x86_64-unknown-linux-gnu' }} # we only need one sdist
+        uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
+        with:
+          name: BitwardenClient-${{ env._PACKAGE_VERSION }}-sdist
+          path: ${{ github.workspace }}/target/wheels/BitwardenClient-*.tar.gz
```
michalchecinski, line 121: New line missing.

michalchecinski (changes requested):

michalchecinski (dismissed):

michalchecinski (dismissed):

Hinton:
Can we extract the crate changes to a separate PR. I think we would want to spend some time figuring out the appropriate defaults.

dani-garcia:
For reference, the PR with the Rustls crate changes is here: https://github.com/bitwarden/sdk/pull/374

languages/python/example.py
```diff
@@ -14,10 +14,10 @@
 
 # Add some logging & set the org id
 logging.basicConfig(level=logging.DEBUG)
-organization_id = "org_id_here"
+organization_id = os.getenv("BWS_ORG_ID")
```
dani-garcia, line 17: Minor thing, but could we change these env variables to be the same we use on the other languages? That way it's easier to test them all at the same time. We could also use env variables for the api and identity URLs.

These are the names we're using with the rest of integrations:
https://github.com/bitwarden/sdk/blob/679673096254e108fa0d71f65bf8fa28ba9679f5/languages/go/example/example.go#L13-L19

Thomas-Avery (changes requested):
Looks good, one minor recommendation.

Thomas-Avery (approved):

michalchecinski (approved):

PR #370, opened 2023-11-30
Improve ClientSettings parsing

dani-garcia:
## Type of change
```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Remove null fields from the ClientSettings input, as serde deserializes the object correctly if the fields are unset, but not if they are explicitly null


Hinton:
Suggesting https://github.com/bitwarden/sdk/pull/432 instead of this.

PR #374, opened 2023-12-01
Support Rustls through cargo features

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Created two cargo features: `use_native_tls` and `use_rustls` in bitwarden and bitwarden-json to allow TLS backend selection in all the crates. By default we use native-tls to keep the same backend we are using now, with the exception of bitwarden-py, which is using rustls to avoid issues with manylinux compatibility.

Another alternative can also be to remove these features, and just force downstream users (and our internal crates) to specifically declare the reqwest dependency with their preferred backend, though this might be a problem if we ever need to change to a different HTTP crate:
```toml
reqwest = {version = "*", features = ["rustls-tls"], default-features = false }
```

<br/>

It would be good to think which backend we're going to use for all the internal crates, too. I'm going to do some testing also on the mobile apps to see how both options fare.

### DIFF
--- a/crates/bitwarden/Cargo.toml
+++ b/crates/bitwarden/Cargo.toml
@@ -13,11 +13,15 @@ edition = "2021"
 rust-version = "1.57"
 
 [features]
-default = ["secrets"]
+default = ["secrets", "use_native_tls"]
 
 secrets = []                    # Secrets manager API
 internal = []                   # Internal testing methods
 mobile = ["uniffi", "internal"] # Mobile-specific features
+
+# TLS backend selection
+use_native_tls = ["reqwest/native-tls"]
+use_rustls = ["reqwest/rustls-tls"]
 
 [dependencies]
 aes = ">=0.8.2, <0.9"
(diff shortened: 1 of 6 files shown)


crates/bitwarden/Cargo.toml
```diff
@@ -13,12 +13,16 @@ edition = "2021"
...
-default = ["secrets"]
+default = ["secrets", "use_native_tls"]
 
 secrets = []                    # Secrets manager API
 internal = []                   # Internal testing methods
 mobile = ["uniffi", "internal"] # Mobile-specific features
 
+# TLS backend selection
+use_native_tls = ["reqwest/native-tls"]
+use_rustls = ["reqwest/rustls-tls"]
```
Hinton, line 24: We should look into using https://github.com/rustls/rustls-platform-verifier for rustls

dani-garcia: Implemented support for it everywhere but android, where I chose rustls-webpki instead. We can change it later once the Android setup is easier.

Hinton (dismissed):
LGTM, let's hold of on merging this until next week.

Hinton (approved):

Hinton (approved):

PR #377, opened 2023-12-01
Improve SDK bindings examples

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Updated all the samples to load access token, organization id and urls from env variables. This made it much easier to quickly test all the bindings. Also added one Java example.


dani-garcia:
Sounds good, I'm going to check that all the examples work correctly when not setting the URLs or setting them to null, we might need #370 for them all to work though.

Hinton (approved):

PR #392, opened 2023-12-04
Initial TOTP implementation

Hinton:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Convert the `TotpService` from `clients` repository to rust.

## Before you submit

### DIFF
--- a/crates/bitwarden/src/vault/totp.rs
+++ b/crates/bitwarden/src/vault/totp.rs
@@ -121,7 +121,7 @@ impl Totp {
 impl FromStr for Totp {
     type Err = Error;
 
-    /// Parses the provided key and returns the corresponding `TotpParams`.
+    /// Parses the provided key and returns the corresponding `Totp`.
     ///
     /// Key can be either:
     /// - A base32 encoded string
--- a/crates/bitwarden/src/vault/totp.rs
+++ b/crates/bitwarden/src/vault/totp.rs
@@ -38,12 +38,15 @@ pub struct TotpResponse {
 /// Arguments:
 /// - `key` - The key to generate the TOTP code from
 /// - `time` - The time in UTC to generate the TOTP code for, defaults to current system time
-pub async fn generate_totp(key: String, time: Option<DateTime<Utc>>) -> Result<TotpResponse> {
+pub(crate) async fn generate_totp(
+    key: String,
+    time: Option<DateTime<Utc>>,
+) -> Result<TotpResponse> {
     let params: Totp = key.parse()?;
 
     let time = time.unwrap_or_else(Utc::now);
 
-    let otp = params.derive_otp(time.timestamp())?;
+    let otp = params.derive_otp(time.timestamp());
 
     Ok(TotpResponse {
         code: otp,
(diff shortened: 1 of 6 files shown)


crates/bitwarden/src/vault/totp.rs
```diff
@@ -22,9 +34,244 @@ pub struct TotpResponse {
...
+impl Algorithm {
+    // Derive the HMAC hash for the given algorithm
+    fn derive_hash(&self, key: &[u8], time: &[u8]) -> Result<Vec<u8>> {
+        fn compute_digest<D: Mac>(mut digest: D, time: &[u8]) -> Vec<u8> {
+            digest.update(time);
+            digest.finalize().into_bytes().to_vec()
+        }
+
+        Ok(match self {
+            Algorithm::Sha1 => compute_digest(HmacSha1::new_from_slice(key)?, time),
```
dani-garcia, line 71: `Hmac::new_from_slice` can't fail, as it accepts keys of any size, so we can just use `.expect()` and make this function infallible

crates/bitwarden/src/vault/totp.rs
```diff
@@ -22,9 +34,244 @@ pub struct TotpResponse {
...
+            secret: "".to_string(),
+        }
+    }
+}
+
+impl Totp {
+    fn derive_otp(&self, time: i64) -> Result<String> {
+        let time = time / self.period as i64;
+
+        let secret = BASE32
```
dani-garcia, line 102: I think the secret should be base32 decoded in the from_str implementation, with that and the change suggested in `derive_hash`, we can make this function infallible

dani-garcia (commented):
Left some comments, mostly style things, and some ways to make the `derive_otp` function infallible

crates/bitwarden/src/vault/totp.rs
```diff
@@ -22,9 +34,248 @@ pub struct TotpResponse {
...
+                        .get("secret")
+                        .map(|v| v.to_string())
+                        .ok_or(Error::Internal("Missing secret in otpauth URI"))?,
+                )?,
+            }
+        } else if key.starts_with("steam://") {
+            Totp {
+                algorithm: Algorithm::Steam,
+                digits: 5,
+                secret: decode_secret(key.strip_prefix("steam://").expect("Prefix is defined"))?,
```
dani-garcia, line 180: We can avoid doing the same thing twice by just destructuring the strip_prefix in the else if:
```suggestion
        } else if let Some(secret) = key.strip_prefix("steam://") {
            Totp {
                algorithm: Algorithm::Steam,
                digits: 5,
                secret: decode_secret(secret)?,
```

crates/bitwarden/src/vault/totp.rs
```diff
@@ -22,9 +34,248 @@ pub struct TotpResponse {
...
+    period: u32,
+    secret: Vec<u8>,
+}
+
+impl Default for Totp {
+    fn default() -> Self {
+        Self {
+            algorithm: Algorithm::Sha1,
+            digits: 6,
+            period: 30,
```
dani-garcia, line 103: I think this default impl would be better off as three constant values: DEFAULT_ALGORITHM, DEFAULT_DIGITS, DEFAULT_PERIOD. As it is, we only ever use the Default impl to extract the values and assign them one by one anyway

Hinton: True, I agree.

PR #388, opened 2023-12-03
SM-866: Add state to the SDK & BWS CLI

coltonhurst:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Tech debt (refactoring, code cleanup, dependency upgrades, etc)
- [ ] Build/deploy pipeline (DevOps)
- [ ] Other
```

## Objective

This PR adds the ability to manage basic state via the SDK. It also updates the `bws` crate to use this SDK change, implementing basic state management for auth tokens. The core benefit is that reauthentication per command will no longer be needed, so that server auth requests will be mitigated.

## Code changes

_(to be added)_

- **file.ext:** Description of what was changed and why

## Before you submit

- Please add **unit tests** where it makes sense to do so (encouraged but not required)

PR #397, opened 2023-12-04
SM-1023: Remove Instant from Client in SDK

coltonhurst:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Refactor out the `Instant` type in Client and use unix timestamps. This is to support this PR: https://github.com/bitwarden/sdk/pull/388 where we need to serialize token expiry times.

It also brings benefits to other areas like WASM, where `Instant` doesn't exist.

## Code changes

- **crates/bitwarden/src/auth/renew.rs:** Update the `renew_token` function to use timestamps instead of the `Instant` type
- **crates/bitwarden/src/client/client.rs:** Update `Client` to store the unix timestamp for token expiration

## Before you submit

### DIFF
--- a/crates/bitwarden/src/auth/renew.rs
+++ b/crates/bitwarden/src/auth/renew.rs
@@ -1,4 +1,4 @@
-use std::time::{Duration, Instant};
+use chrono::Utc;
 
 #[cfg(feature = "internal")]
 use crate::{auth::api::request::ApiTokenRequest, client::UserLoginMethod};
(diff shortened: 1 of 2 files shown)


Hinton (approved):

Hinton:
I think we could refactor the logic around expiration date to. The JWT token contains the expiration date, not sure why we map it back and forth.

PR #398, opened 2023-12-04
[SM-1086] Improve TS bindings

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Implement support for projects, accesstokenlogin and response unwrapping. Also added a small example


languages/js/sdk-client/src/client.ts
```diff
@@ -1,26 +1,48 @@
...
+
 export class BitwardenClient {
   client: BitwardenSDKClient;
 
   constructor(client: BitwardenSDKClient) {
     this.client = client;
   }
 
-  async login(email: string, password: string, pbkdf_iter: number): Promise<ResponseForPasswordLoginResponse> {
+  async accessTokenLogin(accessToken: string): Promise<AccessTokenLoginResponse> {
```
Hinton, line 33: I think we want to return void here?

languages/js/sdk-client/src/client.ts
```diff
@@ -1,26 +1,48 @@
...
+        accessTokenLogin: {
+          accessToken,
+        },
+      }),
+    );
+
+    return handleResponse(Convert.toResponseForAccessTokenLoginResponse(response));
+  }
+
+  async login(email: string, password: string, pbkdf_iter: number): Promise<PasswordLoginResponse> {
```
Hinton, line 45: We should remove the PM features.

Hinton:
This should go through SM review and QA process before merging. From a quick glance it looks to follow the other language bindings. @coltonhurst do we want to look into deprecating the napi bindings and swap to the wasm for gh actions?

coltonhurst:
> This should probably go through SM review and QA process before merging. From a quick glance it looks to follow the other language bindings. @coltonhurst do we want to look into deprecating the napi bindings and swap to the wasm for gh actions?

Sorry for the delayed response. Yes, I think that would be a good thing to take care of, I can create an internal ticket for this. The work deprecating the napi bindings and swapping to wasm for gh actions shouldn't impact this PR though, correct?

coltonhurst (dismissed):

coltonhurst (approved):
Looks great, only other thing I can think of is to add a readme to the languages/js directory that helps explain how to build and run the example project, kind of like we have in the other language directories. Still approved though!

(The ruby one also has a changelog file, which might be beneficial, but that could also wait until we do a release.)

PR #400, opened 2023-12-05
Disable publish ruby

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Code changes

## Before you submit


vgrassia:
@Hinton I just disabled the workflow in the GitHub UI, is that sufficient?

PR #401, opened 2023-12-05
[PM-4930] Implement password validation

Hinton:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Client will call `hash_password` with `Purpose::Local`, persist it and revalidate by using `validate_password(password, persisted_hash)`.

Long term the SDK will persist the hash itself and not require it as input.

## Before you submit

### DIFF
--- a/crates/bitwarden-uniffi/src/auth/mod.rs
+++ b/crates/bitwarden-uniffi/src/auth/mod.rs
@@ -3,6 +3,7 @@ use std::sync::Arc;
 use bitwarden::{
     auth::{password::MasterPasswordPolicyOptions, RegisterKeyResponse},
     client::kdf::Kdf,
+    crypto::HashPurpose,
 };
 
 use crate::{error::Result, Client};
@@ -50,6 +51,7 @@ impl ClientAuth {
         email: String,
         password: String,
         kdf_params: Kdf,
+        purpose: HashPurpose,
     ) -> Result<String> {
         Ok(self
             .0
@@ -57,7 +59,7 @@ impl ClientAuth {
             .read()
             .await
             .kdf()
-            .hash_password(email, password, kdf_params)
+            .hash_password(email, password, kdf_params, purpose)
             .await?)
     }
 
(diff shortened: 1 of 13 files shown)


crates/bitwarden/src/auth/password.rs
```diff
@@ -16,6 +24,33 @@ pub(super) fn satisfies_policy(
...
+            | UserLoginMethod::ApiKey { email, kdf, .. } => {
+                let hash =
+                    determine_password_hash(email, kdf, &password, HashPurpose::LocalAuthorization)
+                        .await?;
+
+                Ok(hash == password_hash)
+            }
+        }
+    } else {
+        Ok(false)
```
Hinton, line 50: Should we maybe throw an error here?

dani-garcia: I think throwing an error makes sense, we're already throwing NotAuthenticated if the login_method is None, seems reasonable to also throw if it's invalid.

crates/bitwarden/src/auth/client_auth.rs
```diff
@@ -90,6 +92,10 @@ impl<'a> ClientAuth<'a> {
     pub async fn send_two_factor_email(&mut self, tf: &TwoFactorEmailRequest) -> Result<()> {
         send_two_factor_email(self.client, tf).await
     }
+
+    pub async fn validate_password(&self, password: String, user_key: String) -> Result<bool> {
+        validate_password(self.client, password, user_key).await
```
dani-garcia, line 97: If we are checking the password against the password hash, `user_key` should be renamed or it's going to cause confusion

crates/bitwarden-uniffi/src/auth/mod.rs
```diff
@@ -76,4 +78,16 @@ impl ClientAuth {
...
+
+    /// Validate the user password
+    pub async fn validate_password(&self, password: String) -> Result<bool> {
+        Ok(self
+            .0
+             .0
+            .write()
+            .await
+            .auth()
+            .validate_password(password, "".to_string())
```
dani-garcia, line 90: This function should also take the password hash as a parameter, as it is, it will always fail because the hash is never going to be the empty string

Hinton: Correct, seems I forgot to fix it 🤦

dani-garcia (approved):

PR #402, opened 2023-12-05
[PM-5691] Extract crypto to separate crate

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

We should make an effort to break up the large `bitwarden` crate into smaller more isolated crates. This extracts the cryptographic primitives into a crate, which better isolates them and may improve compile time slightly. We should continue to extract further areas.

## Code changes

- **crates/bitwarden/src/crypto:** Move everything to `crates/bitwarden-crypto`.

## Before you submit


crates/bitwarden/src/client/kdf.rs
```diff
@@ -22,6 +22,42 @@ pub enum Kdf {
...
+                parallelism,
+            } => bitwarden_crypto::Kdf::Argon2id {
+                iterations: *iterations,
+                memory: *memory,
+                parallelism: *parallelism,
+            },
+        }
+    }
+}
+
```
Hinton, line 60: This is pretty ugly, I think we may want to just not have a `TryFrom` for `PreloginResponseModel`, and instead use a normal function. That would allow the crypto crate to own the Kdf completely and we won't need to map between them.

dani-garcia: Is this because of the orphan rule? Yeah I think it makes sense to have our own function to do the conversion, we already follow that pattern with other types that have a `process_response` function.

crates/bitwarden-crypto/src/error.rs
```diff
@@ -0,0 +1,58 @@
...
+#[derive(Debug, Error)]
+pub enum CryptoError {
+    #[error("The provided key is not the expected type")]
+    InvalidKey,
+    #[error("The cipher's MAC doesn't match the expected value")]
+    InvalidMac,
+    #[error("Error while decrypting EncString")]
+    KeyDecrypt,
+    #[error("The cipher key has an invalid length")]
+    InvalidKeyLen,
```
dani-garcia, line 16: Just noticed now that we have `InvalidKeyLen` and `InvalidLen` variants, and they both seem to be used with `hkdf.expand()`, maybe they should be combined?

crates/bitwarden/Cargo.toml
```diff
@@ -15,20 +15,24 @@ rust-version = "1.57"
...
+    "bitwarden-crypto/mobile",
+] # Mobile-specific features
 
 wasm-bindgen = ["chrono/wasmbind"]
 
 [dependencies]
-aes = ">=0.8.2, <0.9"
 argon2 = { version = ">=0.5.0, <0.6", features = [
     "alloc",
 ], default-features = false }
```
dani-garcia, line 31:
```suggestion
```

crates/bitwarden/Cargo.toml
```diff
@@ -15,20 +15,24 @@ rust-version = "1.57"
...
 base64 = ">=0.21.2, <0.22"
 bitwarden-api-api = { path = "../bitwarden-api-api", version = "=0.2.3" }
 bitwarden-api-identity = { path = "../bitwarden-api-identity", version = "=0.2.3" }
+bitwarden-crypto = { path = "../bitwarden-crypto", version = "=0.1.0" }
 cbc = { version = ">=0.1.2, <0.2", features = ["alloc"] }
```
dani-garcia, line 36:
```suggestion
```

crates/bitwarden/Cargo.toml
```diff
@@ -38,13 +42,11 @@ chrono = { version = ">=0.4.26, <0.5", features = [
 data-encoding = ">=2.5.0, <3.0"
 # We don't use this directly (it's used by rand), but we need it here to enable WASM support
 getrandom = { version = ">=0.2.9, <0.3", features = ["js"] }
-hkdf = ">=0.12.3, <0.13"
 hmac = ">=0.12.1, <0.13"
 lazy_static = ">=1.4.0, <2.0"
 log = ">=0.4.18, <0.5"
 num-bigint = ">=0.4, <0.5"
 num-traits = ">=0.2.15, <0.3"
```
dani-garcia, line 49:
```suggestion
```

dani-garcia (commented):
Overall looks good, some minor things and duplicate dependencies

crates/bitwarden/src/client/kdf.rs
Hinton: We own neither of the structs which means we can't implement traits on them.

dani-garcia (approved):

dani-garcia (dismissed):

michalchecinski (changes requested):
There's a need to add `bitwarden-crypto` here: https://github.com/bitwarden/sdk/blob/1028efde6557e30d241c03fd4754d32fe38db17a/.github/workflows/version-bump.yml#L18

And also to publish workflow: https://github.com/bitwarden/sdk/blob/main/.github/workflows/publish-rust-crates.yml

michalchecinski (approved):

PR #421, opened 2023-12-11
[PM-4694] Implement PIN unlock

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
Implement support for deriving the PIN encrypted user key and initializing crypto with it. Updated the mobile app demos to use this new functionality.

### DIFF
--- a/crates/bitwarden/src/client/client.rs
+++ b/crates/bitwarden/src/client/client.rs
@@ -224,10 +224,9 @@ impl Client {
     #[cfg(feature = "mobile")]
     pub(crate) fn initialize_user_crypto_decrypted_key(
         &mut self,
-        decrypted_user_key: &str,
+        user_key: SymmetricCryptoKey,
         private_key: EncString,
     ) -> Result<&EncryptionSettings> {
-        let user_key = decrypted_user_key.parse::<SymmetricCryptoKey>()?;
         self.encryption_settings = Some(EncryptionSettings::new_decrypted_key(
             user_key,
             private_key,
@@ -236,6 +235,27 @@ impl Client {
             .encryption_settings
             .as_ref()
             .expect("It was initialized on the previous line"))
+    }
+
+    #[cfg(feature = "mobile")]
+    pub(crate) fn initialize_user_crypto_pin(
+        &mut self,
+        pin: &str,
+        pin_protected_user_key: EncString,
+        private_key: EncString,
+    ) -> Result<&EncryptionSettings> {
+        use crate::crypto::MasterKey;
+
+        let pin_key = match &self.login_method {
+            Some(LoginMethod::User(
+                UserLoginMethod::Username { email, kdf, .. }
+                | UserLoginMethod::ApiKey { email, kdf, .. },
+            )) => MasterKey::derive(pin.as_bytes(), email.as_bytes(), kdf)?,
+            _ => return Err(Error::NotAuthenticated),
+        };
+
+        let decrypted_user_key = pin_key.decrypt_user_key(pin_protected_user_key)?;
+        self.initialize_user_crypto_decrypted_key(decrypted_user_key, private_key)
     }
 
     pub(crate) fn initialize_crypto_single_key(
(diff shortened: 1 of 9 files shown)


crates/bitwarden/src/mobile/crypto.rs
```diff
@@ -91,3 +106,50 @@ pub async fn get_user_encryption_key(client: &mut Client) -> Result<String> {
...
 }
+
+#[cfg(feature = "internal")]
+#[derive(Serialize, Deserialize, Debug, JsonSchema)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+#[cfg_attr(feature = "mobile", derive(uniffi::Record))]
+pub struct DerivePinKeyRequest {
+    /// The user's PIN
+    pub pin: String,
+}
```
Hinton, line 117: Do we need a request for this vs just accepting a string?

crates/bitwarden/src/crypto/master_key.rs
```diff
@@ -52,6 +52,15 @@ impl MasterKey {
...
     }
+
+    #[cfg(feature = "mobile")]
+    pub(crate) fn encrypt_user_key(&self, user_key: &SymmetricCryptoKey) -> Result<EncString> {
+        use super::KeyEncryptable;
+
+        let stretched_key = stretch_master_key(self)?;
+
+        (user_key.to_vec().as_slice()).encrypt_with_key(&stretched_key)
+    }
```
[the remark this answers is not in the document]
dani-garcia: Good point, changed

crates/bitwarden/src/client/client.rs
```diff
@@ -238,6 +237,27 @@ impl Client {
             .expect("It was initialized on the previous line"))
     }
 
+    #[cfg(feature = "mobile")]
+    pub(crate) fn initialize_user_crypto_pin(
+        &mut self,
+        pin: &str,
+        pin_protected_user_key: EncString,
+        private_key: EncString,
+    ) -> Result<&EncryptionSettings> {
```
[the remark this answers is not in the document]
dani-garcia: Yeah, I think it would be a good idea to merge them into one with an enum, just like we're doing in the uniffi crate.

Hinton (commented):
Can we add some test cases.

crates/bitwarden/src/crypto/master_key.rs
```diff
@@ -52,6 +52,14 @@ impl MasterKey {
         let dec: Vec<u8> = user_key.decrypt_with_key(&stretched_key)?;
         SymmetricCryptoKey::try_from(dec.as_slice())
     }
+
+    pub(crate) fn encrypt_user_key(&self, user_key: &SymmetricCryptoKey) -> Result<EncString> {
+        use super::KeyEncryptable;
+
+        let stretched_key = stretch_master_key(self)?;
+
+        (user_key.to_vec().as_slice()).encrypt_with_key(&stretched_key)
```
Hinton, line 61: I previously used the `EncString::encrypt_aes256_hmac` since we want to be certain it's encrypted that way.

crates/bitwarden/src/mobile/crypto.rs
```diff
@@ -141,3 +141,54 @@ pub fn derive_pin_key(client: &mut Client, pin: String) -> Result<DerivePinKeyRe
...
+                email: "test@bitwarden.com".into(),
+                private_key: priv_key.to_owned(),
+                method: InitUserCryptoMethod::Pin {
+                    pin: "1234".into(),
+                    pin_protected_user_key: pin_key.pin_protected_user_key,
+                },
+            },
+        )
+        .await
+        .unwrap();
```
Hinton, line 192: Do we want to verify we get the same user key out?

dani-garcia: They are decrypting the same private key so they should be, but it's a good idea to add a check just in case anyway.

Hinton (dismissed):

crates/bitwarden/src/mobile/crypto.rs
Hinton: Ah true.

Hinton (approved):

PR #427, opened 2023-12-12
Change login_method to have the access token directly

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Now that set_login_method is split from set_tokens, we can just change this to directly contain the token without adding the Clone bound to it. This way we avoid duplicating 2/3rds of the AccessToken.

This should help with #388

### DIFF
--- a/crates/bitwarden/src/client/access_token.rs
+++ b/crates/bitwarden/src/client/access_token.rs
@@ -1,4 +1,4 @@
-use std::str::FromStr;
+use std::{fmt::Debug, str::FromStr};
 
 use base64::Engine;
 use uuid::Uuid;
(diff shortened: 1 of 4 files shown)


crates/bitwarden/src/client/access_token.rs
```diff
@@ -15,6 +15,15 @@ pub struct AccessToken {
     pub encryption_key: SymmetricCryptoKey,
 }
 
+// We don't want to log the more sensitive fields from an AccessToken
+impl Debug for AccessToken {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("AccessToken")
+            .field("access_token_id", &self.access_token_id)
+            .finish()
```
Hinton, line 23: This only logs the id?

Hinton (approved):

crates/bitwarden/src/client/access_token.rs
dani-garcia: Yeah, I didn't want to log the secret or the encryption key to avoid leaking potentially sensitive information in the logs. I think just logging the ID is a good compromise between debug-ability and security in this case.

PR #431, opened 2023-12-14
Remove token from client

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
As mentioned in #388, the `token` and `refresh_token` fields in `Client` are never really required directly by `Client`. The `token` is already directly set within the API clients, and the `refresh_token` is only used for the renew function with the user login method.

Knowing this, we can remove `token` entirely and move `refresh_token` into `UserLoginMethod::Username`.

### DIFF
--- a/crates/bitwarden/src/auth/password.rs
+++ b/crates/bitwarden/src/auth/password.rs
@@ -87,6 +87,7 @@ mod tests {
                 iterations: NonZeroU32::new(100_000).unwrap(),
             },
             client_id: "1".to_string(),
+            refresh_token: None,
         }));
 
         let password = "password123".to_string();
--- a/crates/bitwarden/src/auth/login/access_token.rs
+++ b/crates/bitwarden/src/auth/login/access_token.rs
@@ -54,11 +54,7 @@ pub(crate) async fn login_access_token(
             .parse()
             .map_err(|_| Error::InvalidResponse)?;
 
-        client.set_tokens(
-            r.access_token.clone(),
-            r.refresh_token.clone(),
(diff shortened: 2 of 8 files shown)


crates/bitwarden/src/auth/password.rs
```diff
@@ -87,6 +87,7 @@ mod tests {
                 iterations: NonZeroU32::new(100_000).unwrap(),
             },
             client_id: "1".to_string(),
+            refresh_token: None,
```
Hinton, line 90: I'm a bit hesitant to couple refresh_token to login method. While it's true certain login methods get a refresh token while others don't, there is no direct relation to users.

dani-garcia: It's not so much which methods return a refresh token and which don't, but wether we actually use the refresh_token for that method when renewing tokens.

At the moment we have the following renewal patterns:
- User with username: We use client_id and refresh_token
- User with apikey: We use client_id and client_secret
- Service account with access token: We use the access token's id and client_secret

So even if the other methods return a refresh_token, it would only be used in the case of user with username.

Hinton: In that case what do you think about changing the flow to always use refresh_token if set and only falling back to the generic renew if it doesn't exist?

dani-garcia: Hmm, that can work, but we would also need to pull the client_id from each login_method and into the client, as we need it for the token refresh request.

PR #433, opened 2023-12-14
Wire up code coverage

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Code changes

## Before you submit

### DIFF
--- a/.github/workflows/rust-test.yml
+++ b/.github/workflows/rust-test.yml
@@ -49,6 +49,34 @@ jobs:
       - name: Test
         run: cargo test --all-features
 
+  coverage:
+    name: Coverage
+    runs-on: "ubuntu-latest"
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+      - name: Install rust
+        uses: dtolnay/rust-toolchain@be73d7920c329f220ce78e0234b8f96b7ae60248 # stable
+        with:
+          toolchain: stable
+          components: llvm-tools
+
+      - name: Cache cargo registry
+        uses: Swatinem/rust-cache@3cf7f8cc28d1b4e7d01e3783be10a97d55d483c8 # v2.7.1
+
+      - name: Install cargo-llvm-cov
+        run: cargo install cargo-llvm-cov --version 0.5.38
+
+      - name: Generate coverage
+        run: cargo llvm-cov --all-features --lcov --output-path lcov.info
+
+      - name: Upload to codecov.io
+        uses: codecov/codecov-action@eaaf4bedf32dbdc6b720b63067d99c4d77d6047d # v3.1.4
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+
   wasm:
     name: WASM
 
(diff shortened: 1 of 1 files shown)


.github/workflows/rust-test.yml
```diff
@@ -49,6 +49,34 @@ jobs:
       - name: Test
         run: cargo test --all-features
 
+  coverage:
```
withinfocus, line 52: ❓ Can you save a run and just use the above, then append what's needed for Codecov?

.github/workflows/rust-test.yml
```diff
@@ -49,6 +49,34 @@ jobs:
...
+      - name: Install cargo-llvm-cov
+        run: cargo install cargo-llvm-cov --version 0.5.38
+
+      - name: Generate coverage
+        run: cargo llvm-cov --all-features --lcov --output-path lcov.info --ignore-filename-regex "crates/bitwarden-api-"
+
+      - name: Upload to codecov.io
+        uses: codecov/codecov-action@eaaf4bedf32dbdc6b720b63067d99c4d77d6047d # v3.1.4
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
```
withinfocus, line 78: ℹ️ Not needed for us as open source. For example: https://github.com/bitwarden/passwordless-dotnet/blob/a9d28bca38fb644b2dcbd51608e7768bf6610738/.github/workflows/main.yml#L129-L130

.github/workflows/rust-test.yml
```diff
@@ -49,6 +49,34 @@ jobs:
...
+        uses: Swatinem/rust-cache@3cf7f8cc28d1b4e7d01e3783be10a97d55d483c8 # v2.7.1
+
+      - name: Install cargo-llvm-cov
+        run: cargo install cargo-llvm-cov --version 0.5.38
+
+      - name: Generate coverage
+        run: cargo llvm-cov --all-features --lcov --output-path lcov.info --ignore-filename-regex "crates/bitwarden-api-"
+
+      - name: Upload to codecov.io
+        uses: codecov/codecov-action@eaaf4bedf32dbdc6b720b63067d99c4d77d6047d # v3.1.4
```
withinfocus, line 76: 🌱 There's a v4 that's a better approach but still beta. Up to us if we want to use it or not.

.github/workflows/rust-test.yml
Hinton: The above is a matrix. We could conditionally swap out half the commands but that seems somewhat annoying.

withinfocus (approved):
Removal of the token parameter is optional I suppose, and maybe we'll use it eventually.

.github/workflows/rust-test.yml
Hinton: It's a bit unclear if it's going to be required for v4. I think we can safely keep it for now if nothing else it should ensure only we can push coverage reports.

.github/workflows/rust-test.yml
```diff
@@ -49,6 +49,34 @@ jobs:
       - name: Test
         run: cargo test --all-features
 
+  coverage:
+    name: Coverage
+    runs-on: "ubuntu-latest"
```
joseph-flinn, line 54: We pin the version of the runner so that update of `ubuntu-latest` to the next version doesn't break the pipelines (every update from GitHub has broken the pipelines so far)

```suggestion
    runs-on: "ubuntu-22.04"
```

joseph-flinn (changes requested):

PR #435, opened 2023-12-15
Add support for Hmac org keys

Hinton:
## Type of change

```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Organization keys can be generated using `Rsa2048_OaepSha1_HmacSha256_B64` which previously wasn't supported. This PR adds support for decrypting org keys in that format, and changes the part of asymetric hmac to also contain a mac.

## Before you submit

### DIFF
--- a/crates/bitwarden/src/crypto/enc_string.rs
+++ b/crates/bitwarden/src/crypto/enc_string.rs
@@ -144,12 +144,12 @@ impl FromStr for EncString {
                 Ok(EncString::Rsa2048_OaepSha1_B64 { data })
             }
             #[allow(deprecated)]
-            ("5", 1) => {
+            ("5", 2) => {
                 let data = from_b64_vec(parts[0])?;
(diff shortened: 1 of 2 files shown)


crates/bitwarden/src/crypto/enc_string.rs
```diff
@@ -170,6 +172,29 @@ impl EncString {
         s.map(|s| s.parse()).transpose()
     }
 
+    /// TODO: Convert this to a trait method
```
Hinton, line 175: We probably want to rework all this logic and ensure we differentiate asymmetric and symmetric enc strings. And expand our key locator, decryption pattern to handle assymetric keys.

dani-garcia (approved):
LGTM, we're not checking the HMAC of the deprecated types but maybe that's not a big deal

Hinton:
@dani-garcia I'm replicating the behavior of the js clients which also doesn't.

PR #436, opened 2023-12-15
Improve internal error

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Allow using dynamic strings on the Internal error variant, tested on the passphrase length error.

The Internal variant is using a `Cow<'static, str>`, to avoid unnecessary allocations on all the cases where a `&'static str` would suffice.

I've implemented `From<&'static str>` and `From<String>` for Error. It feels a bit magic, but it avoids repeating `Error::Internal` all over the place.

### DIFF
--- a/crates/bitwarden/src/auth/jwt_token.rs
+++ b/crates/bitwarden/src/auth/jwt_token.rs
@@ -29,9 +29,7 @@ impl FromStr for JWTToken {
     fn from_str(s: &str) -> Result<Self> {
         let split = s.split('.').collect::<Vec<_>>();
         if split.len() != 3 {
-            return Err(crate::error::Error::Internal(
-                "JWT token has an invalid number of parts",
-            ));
+            return Err("JWT token has an invalid number of parts".into());
         }
         let decoded = BASE64_ENGINE.decode(split[1])?;
         Ok(serde_json::from_slice(&decoded)?)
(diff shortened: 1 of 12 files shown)


Hinton (commented):
Hmm, I wonder if removing the explicit `Error::Internal` will make it more difficult to search for the dynamic errors in the future. Ideally we should try and replace most of them with explicit types.

dani-garcia:
The `?` operator already does make the error conversion a bit implicit as it automatically calls `.into()` to convert the errors, so I don't think this is much worse than the current situation. We could always remove the `Into` impl and just have a function `Error::new` or `Error::internal` to make it more explicit. 

Ultimately I think we need to rethink the error handling, at the moment we have a lot of error variants that aren't that actionable by the library consumers, which just bloats the error type. I think some of that would get improved as we split the code base into crates, like when we extract the crypto features, which would also mean split Error enums. But in a lot of cases the explicitness won't really help (I'm thinking variants like `Serde`, `InvalidBase64`, `InvalidEncString`, `AccessTokenInvalid`, these all just mean "you gave me the wrong input" or "the server gave me unexpected data", and the course of action that any client can follow won't change much based on them)

Hinton (approved):

PR #447, opened 2023-12-18
Implement KeyEncryptable support for Asymmetric keys

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Implement KeyEncryptable support for Asymmetric keys.


crates/bitwarden/src/crypto/asymmetric_crypto_key.rs
```diff
@@ -0,0 +1,99 @@
...
+use crate::{
+    crypto::CryptoKey,
+    error::{CryptoError, Error, Result},
+    util::BASE64_ENGINE,
+};
+
+/// An asymmetric encryption key. Used to encrypt and decrypt [`EncString`](crate::crypto::EncString)
+pub struct AsymmetricCryptoKey {
+    pub(in crate::crypto) key: RsaPrivateKey,
+}
```
Hinton, line 15: I think we need to add public key here to. Or derive it from the private key.

Usually encrypting is done with the public key, and decrypting using the private.

dani-garcia: The docs say that RsaPrivateKey implements the PublicKey interface so it should be handled internally by the crate. That said, I haven't implemented RSA encrypt on this PR so haven't had the time to test it.
https://docs.rs/rsa/latest/rsa/struct.RsaPrivateKey.html#method.to_public_key

crates/bitwarden/src/crypto/asymmetric_crypto_key.rs
```diff
@@ -0,0 +1,99 @@
...
+
+// We manually implement these to make sure we don't print any sensitive data
+impl std::fmt::Debug for AsymmetricCryptoKey {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("AsymmetricCryptoKey").finish()
+    }
+}
+
+/*
```
Hinton, line 80: Disabled test?

crates/bitwarden/src/crypto/enc_string.rs
```diff
@@ -395,33 +369,26 @@ fn invalid_len_error(expected: usize) -> impl Fn(Vec<u8>) -> EncStringParseError
...
-            EncString::AesCbc256_HmacSha256_B64 { iv, mac, data } => {
-                let mac_key = key.mac_key.ok_or(CryptoError::InvalidMac)?;
-                let dec = decrypt_aes256_hmac(iv, mac, data.clone(), mac_key, key.key)?;
-                Ok(dec)
-            }
-            _ => Err(CryptoError::InvalidKey.into()),
-        }
+impl<Key: CryptoKey> KeyDecryptable<Key, Vec<u8>> for EncString {
+    fn decrypt_with_key(&self, key: &Key) -> Result<Vec<u8>> {
+        key.decrypt(self)
```
Hinton, line 380: Does these interfaces still make sense? Why not simply call `key.decrypt` in the consumer.

dani-garcia: This is mostly a stopgap to get this working until I finish up another PR to get the EncString split into symm/asymm variants, then we can just have `impl KeyDecryptable<SymmCryptoKey> for EncString` and `impl KeyDecryptable<AsymmCryptoKey> for AsymmEncString`

crates/bitwarden/src/crypto/asymmetric_crypto_key.rs
```diff
@@ -0,0 +1,99 @@
...
+impl TryFrom<&[u8]> for AsymmetricCryptoKey {
+    type Error = Error;
+
+    fn try_from(value: &[u8]) -> Result<Self, Self::Error> {
+        use rsa::pkcs8::DecodePrivateKey;
+        Ok(Self {
+            key: rsa::RsaPrivateKey::from_pkcs8_der(value).map_err(|_| CryptoError::InvalidKey)?,
+        })
+    }
+}
```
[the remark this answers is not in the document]
dani-garcia: Initially I had a FromStr for the pem format as well, but then was worried it might be confused for the FromStr on SymmetricCryptoKey which is just base64 encoded bytes, so decided to go with an explicit alternative.

It's probably a good idea to be explicit in both cases though, yeah.

crates/bitwarden/src/crypto/enc_string/asymmetric.rs
```diff
@@ -178,6 +177,44 @@ impl AsymmEncString {
     }
 }
 
+impl KeyEncryptable<AsymmetricCryptoKey, AsymmEncString> for &[u8] {
+    fn encrypt_with_key(self, _key: &AsymmetricCryptoKey) -> Result<AsymmEncString> {
+        todo!("implement encrypt")
+    }
+}
```
Hinton, line 184: Any benefit to implementing this vs excluding it? It should cause a compile error without it which is preferable over runtime.

crates/bitwarden/src/crypto/enc_string/asymmetric.rs
Hinton, line 1: We seem to be missing tests in many of these?

crates/bitwarden/src/crypto/enc_string/asymmetric.rs
dani-garcia: That's a good point yeah, removed

crates/bitwarden/src/crypto/enc_string/asymmetric.rs
dani-garcia: Should be better now, I removed the previous `decrypt` function which is replaced by KeyDecryptable and added some tests for the other AsymmEncString types.

Hinton (dismissed):

Hinton (approved):

PR #448, opened 2023-12-18
Split EncString for asymmetric support

dani-garcia:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Split `EncString` into an asymmetric variant. I've left the symmetric variant with the name `EncString` as it's the most common variant to avoid major changes.

This is somewhat related to #447 and should be merged first so that PR can take advantage of the symm/asymm separation.

I've moved the `EncString` files into their own folder to organize the crypto directory a bit better too.

### DIFF
--- a/crates/bitwarden/src/crypto/enc_string/asymmetric.rs
+++ b/crates/bitwarden/src/crypto/enc_string/asymmetric.rs
@@ -1,0 +1,272 @@
+use std::{fmt::Display, str::FromStr};
+
+use base64::Engine;
+#[cfg(feature = "internal")]
+use rsa::{Oaep, RsaPrivateKey};
+use serde::{de::Visitor, Deserialize};
+
+use crate::{
+    error::{EncStringParseError, Error, Result},
+    util::BASE64_ENGINE,
+};
+
+#[cfg(feature = "internal")]
+use crate::error::CryptoError;
+
+use super::{from_b64_vec, split_enc_string};
+
+/// # Encrypted string primitive
+///
+/// [AsymmEncString] is a Bitwarden specific primitive that represents an asymmetrically encrypted string. They are
+/// are used together with the KeyDecryptable and KeyEncryptable traits to encrypt and decrypt
+/// data using AsymmetricCryptoKeys.
+///
+/// The flexibility of the [AsymmEncString] type allows for different encryption algorithms to be used
+/// which is represented by the different variants of the enum.
+///
+/// ## Note
+///
+/// For backwards compatibility we will rarely if ever be able to remove support for decrypting old
+/// variants, but we should be opinionated in which variants are used for encrypting.
+///
+/// ## Variants
+/// - [Rsa2048_OaepSha256_B64](AsymmEncString::Rsa2048_OaepSha256_B64)
+/// - [Rsa2048_OaepSha1_B64](AsymmEncString::Rsa2048_OaepSha1_B64)
+///
+/// ## Serialization
+///
+/// [AsymmEncString] implements [Display] and [FromStr] to allow for easy serialization and uses a
+/// custom scheme to represent the different variants.
+///
+/// The scheme is one of the following schemes:
+/// - `[type].[data]`
+///
+/// Where:
+/// - `[type]`: is a digit number representing the variant.
+/// - `[data]`: is the encrypted data.
+#[derive(Clone)]
+#[allow(unused, non_camel_case_types)]
+pub enum AsymmEncString {
+    /// 3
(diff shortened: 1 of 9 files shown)


crates/bitwarden/src/crypto/enc_string/asymmetric.rs
```diff
@@ -0,0 +1,272 @@
...
+            ("6", 2) => {
+                let data = from_b64_vec(parts[0])?;
+                let mac: Vec<u8> = from_b64_vec(parts[1])?;
+                Ok(AsymmEncString::Rsa2048_OaepSha1_HmacSha256_B64 { data, mac })
+            }
+
+            (enc_type, parts) => Err(EncStringParseError::InvalidType {
+                enc_type: enc_type.to_string(),
+                parts,
+            }
```
Hinton, line 100: Do you think it could be confusing to return the same error type for both symmetric and asymmetic? I'm primarily thinking about the cases where we send a symmetric key to asymmetric since it's technically a valid encstring.

crates/bitwarden/src/crypto/enc_string/asymmetric.rs
```diff
@@ -0,0 +1,272 @@
...
+            .into()),
+        }
+    }
+}
+
+#[allow(unused)]
+impl AsymmEncString {
+    /// TODO: Convert this to a trait method
+    #[cfg(feature = "internal")]
+    pub(crate) fn decrypt_with_private_key(&self, key: &RsaPrivateKey) -> Result<Vec<u8>> {
```
Hinton, line 110: Since it's a different type we could just call this `decrypt` or `decrypt_with_key`.

```suggestion
    pub(crate) fn decrypt(&self, key: &RsaPrivateKey) -> Result<Vec<u8>> {
```

crates/bitwarden/src/crypto/enc_string/asymmetric.rs
```diff
@@ -0,0 +1,272 @@
...
+            where
+                E: serde::de::Error,
+            {
+                AsymmEncString::from_str(v).map_err(|e| E::custom(format!("{:?}", e)))
+            }
+        }
+
+        deserializer.deserialize_str(CSVisitor)
+    }
+}
```
Hinton, line 172: This looks mostly identical to Symm, is there a way to avoid duplicating as much code? Maybe if we have the generic higher EncString enum and delgate it to that?

Hinton (commented):
Overall looks good, some thoughts.

I think we should rename EncString to `SymcEncString`. We should also standardize on our abbrevation, should `SymmetricCryptoKey` be `SymCryptokey`.

I also wonder if we should keep `EncString` as an Enum around Sym and Assym.

crates/bitwarden/src/crypto/enc_string/asymmetric.rs
dani-garcia: I've extracted the visitor implementation out, which should help a fair bit

crates/bitwarden/src/crypto/enc_string/asymmetric.rs
dani-garcia: This will be removed when KeyEncryptable supports asymm keys, but for now we can rename it yes.

crates/bitwarden/src/crypto/enc_string/asymmetric.rs
dani-garcia: Yeah fair point, I've split the error into symm/asymm variants.

dani-garcia:
I chose not to rename `EncString` to `SymmEncString` mostly because it's used everywhere, and because it's 90% of our crypto operations, so in a way it's the default, and you have to go out of your way if you want to use asymmetric encryption.

About having an enum encompassing both EncString types, should be easy enough to do in the future but at the moment it doesn't look like we have a need for it. Also I'd be worried that it would give people the idea that it's the main one they would need to use, while we should prefer the use of the compile time explicit Symm/AsymmEncString as much as possible.

The naming convention in my head is:
`EncString`            /   `SymmCryptoKey`
`AsymmEncString` / `AsymmCryptoKey`

`AnyEncString` <- For the enum containing both variants, makes it more explicit that it's a runtime selection

`EncString` would break the convention of the other names, but I think it's going the far more common type, so it makes sense to me for it to be shorter. Plus we're going to enforce a to compile time that you can only use a symmetric key with a symmetric EncString, so it shouldn't cause accidental misuse.

For the rest, they would see less use in most common use cases, so I'm fine with going with the `Symm` abreviation or the full `Symmetric` even.

Hinton (approved):

PR #450, opened 2023-12-19
Change jwt parsing to use URL_SAFE_NO_PAD

Hinton:
## Type of change

```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

JWT should be url encoded and never use padding.

## Before you submit


MGibson1 (approved):
This is the default, right?

Hinton:
This follows the JWT spec, and how we encode our tokens.

PR #454, opened 2023-12-20
Expand testing to include sync and models

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Include some more complex tests of client login/sync/encrypt/decrypt

Some model fields were previously private, so I've had to change them to pub, for now it's not a problem as no one is using the rust models directly.

### DIFF
--- a/crates/bitwarden/src/vault/collection.rs
+++ b/crates/bitwarden/src/vault/collection.rs
@@ -13,28 +13,28 @@ use crate::{
 #[serde(rename_all = "camelCase", deny_unknown_fields)]
 #[cfg_attr(feature = "mobile", derive(uniffi::Record))]
 pub struct Collection {
-    id: Option<Uuid>,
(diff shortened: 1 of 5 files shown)


crates/bitwarden/src/vault/collection.rs
```diff
@@ -13,28 +13,28 @@ use crate::{
...
 
-    name: EncString,
+    pub name: EncString,
 
-    external_id: Option<String>,
-    hide_passwords: bool,
-    read_only: bool,
+    pub external_id: Option<String>,
+    pub hide_passwords: bool,
+    pub read_only: bool,
```
Hinton, line 23: The long term aspiration is to not expose this model externally. So I'm a bit hesitant to just make all of these public with that knowledge.

Would it not be sufficient if we verify that the decrypted view matches our expectation?

PR #476, opened 2024-01-05
Codecov ignore tools

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [x] Other
```

## Objective

Ignore tools since they don't need to be tested.

## Before you submit

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,0 +1,5 @@
+coverage:
+  status:
+    project:
+      default:
+        informational: true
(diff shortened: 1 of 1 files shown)


withinfocus (approved):

withinfocus (changes requested):
Lemme test something.

withinfocus (approved):
I set a global config, but it came with this message:

> bitwarden YAML configuration has been successfully saved. New coverage reports will reflect these changes. Repositories with a codecov.yaml file extend and override this account level Codecov config.

so we'll want to still merge this given the exclusions.

PR #484, opened 2024-01-08
Fix SendView.key behaviour and fix types for SendFile and SendFileView

Hinton:
## Type of change

```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Fix SendView.key behaviour and fix types for SendFile and SendFileView

## Before you submit

### DIFF
--- a/crates/bitwarden/src/vault/send.rs
+++ b/crates/bitwarden/src/vault/send.rs
@@ -1,3 +1,4 @@
+use base64::{engine::general_purpose::URL_SAFE_NO_PAD, Engine};
 use bitwarden_api_api::models::{SendFileModel, SendResponseModel, SendTextModel};
 use chrono::{DateTime, Utc};
 use schemars::JsonSchema;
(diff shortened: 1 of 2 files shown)


crates/bitwarden/src/vault/send.rs
```diff
@@ -134,8 +136,12 @@ impl Send {
         enc_key: &SymmetricCryptoKey,
     ) -> Result<SymmetricCryptoKey> {
         let key: Vec<u8> = send_key.decrypt_with_key(enc_key)?;
-        let key = derive_shareable_key(key.try_into().unwrap(), "send", Some("send"));
-        Ok(key)
+        Self::derive_shareable_key(key)
+    }
+
+    fn derive_shareable_key(key: Vec<u8>) -> Result<SymmetricCryptoKey, Error> {
```
dani-garcia, line 142: We can change the definition on this function to take a slice, that we we can avoid some clones when using this function
```suggestion
    fn derive_shareable_key(key: &[u8]) -> Result<SymmetricCryptoKey, Error> {
```

dani-garcia (commented):
I'm thinking some of this code might be simplified if we had a `struct Base64String(String)` that implemented `KeyEncryptable` and just did the correct thing, but maybe that's something we can look into later.

PR #487, opened 2024-01-08
Fix firefox forwarder when website isn't set

dani-garcia:
## Type of change
```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
When the website parameter is None, the request would include a `generated_for: null` field, while the API expects the field to not be present.


Hinton (approved):

crates/bitwarden/src/tool/generators/username_forwarders/firefox.rs
```diff
@@ -77,6 +78,19 @@ mod tests {
...
+                .and(matchers::header("Content-Type", "application/json"))
+                .and(matchers::header("Authorization", "Token MY_OTHER_TOKEN"))
+                .and(matchers::body_json(json!({
+                    "enabled": true,
+                    "description": "Generated by Bitwarden."
+                })))
+                .respond_with(ResponseTemplate::new(201).set_body_json(json!({
+                    "full_address": "856f7765@mozmail.com"
+                })))
+                .expect(1),
```
Hinton, line 93: It's generally easier to debug if each case is it's own separate test, vs trying to figure out what belongs to what.

dani-garcia: That's a solid point, I've split them into separate test functions now.

Hinton (approved):

PR #488, opened 2024-01-08
Support generating a random send key if none is provided

Hinton:
## Type of change

```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

When creating a new send, we should generate a key for it.

## Before you submit

### DIFF
--- a/crates/bitwarden/src/vault/send.rs
+++ b/crates/bitwarden/src/vault/send.rs
@@ -245,14 +245,18 @@ impl KeyEncryptable<Send> for SendView {
     fn encrypt_with_key(self, key: &SymmetricCryptoKey) -> Result<Send> {
         // For sends, we first decrypt the send key with the user key, and stretch it to it's full size
         // For the rest of the fields, we ignore the provided SymmetricCryptoKey and the stretched key
-        let k = match self.key {
-            Some(k) => URL_SAFE_NO_PAD
+        let k = match (self.key, self.id) {
+            // Existing send, decrypt key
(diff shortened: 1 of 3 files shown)


crates/bitwarden/src/vault/send.rs
```diff
@@ -245,9 +246,15 @@ impl KeyEncryptable<Send> for SendView {
...
-        let k = URL_SAFE_NO_PAD
-            .decode(self.key.ok_or(CryptoError::MissingKey)?)
-            .map_err(|_| CryptoError::InvalidKey)?;
+        let k = match self.key {
+            Some(k) => URL_SAFE_NO_PAD
+                .decode(k)
+                .map_err(|_| CryptoError::InvalidKey)?,
+            None => {
+                let r: [u8; 16] = rand::thread_rng().gen();
+                r.to_vec()
```
Hinton, line 255: Should we provide a function in the crypto directory for this? Web uses this method https://github.com/bitwarden/clients/blame/main/libs/common/src/platform/services/web-crypto-function.service.ts#L370-L385

dani-garcia: Sounds reasonable, we're already doing something very similar in SymmetricCryptoKey https://github.com/bitwarden/sdk/blob/dfbfdbfea3108b77b4acefb9187ff90953f6250f/crates/bitwarden/src/crypto/symmetric_crypto_key.rs#L24-L25

dani-garcia (approved):

PR #489, opened 2024-01-09
[PM-5346] Implement password strength

Hinton:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Break up `password.rs` into a module for clarity. Implement password strength along test cases that are replicated in the clients codebase.

## Before you submit


crates/bitwarden/src/auth/password/strength.rs
```diff
@@ -0,0 +1,99 @@
...
+    let estimate = zxcvbn(&password, &arr);
+
+    match estimate {
+        Ok(estimate) => estimate.score(),
+        _ => 0,
+    }
+}
+
+fn email_to_user_inputs(email: &str) -> Vec<String> {
+    if email.is_empty() {
```
dani-garcia, line 29: The empty case should already be handled by the `split_once`, as it would return none, I don't think we need to do it explicitly.

crates/bitwarden/src/auth/password/strength.rs
```diff
@@ -0,0 +1,99 @@
+use std::collections::HashSet;
+
+use zxcvbn::zxcvbn;
+
+const GLOBAL_INPUTS: [&str; 3] = ["bitwarden", "BW", "Bitwarden"];
```
dani-garcia, line 5: We use different inputs on the clients, is this intentional? https://github.com/bitwarden/clients/blob/0fba5810d9476db97b11f2c72213e9acda72f1f7/libs/common/src/tools/password-strength/password-strength.service.ts#L21-L24

crates/bitwarden/src/auth/password/strength.rs
```diff
@@ -0,0 +1,99 @@
...
+    additional_inputs.extend(&GLOBAL_INPUTS);
+    additional_inputs.extend(&email_ref);
+    let arr: Vec<_> = additional_inputs.into_iter().collect();
+
+    let estimate = zxcvbn(&password, &arr);
+
+    match estimate {
+        Ok(estimate) => estimate.score(),
+        _ => 0,
+    }
```
dani-garcia, line 25:
```suggestion
    zxcvbn(&password, &arr).map_or(0, |e| e.score())
```

This is more of a personal choice, but it could be replaced by a single `map_or`

crates/bitwarden/src/auth/password/strength.rs
```diff
@@ -0,0 +1,99 @@
...
+
+pub(crate) fn password_strength(
+    password: String,
+    email: String,
+    additional_inputs: Vec<String>,
+) -> u8 {
+    let email_parts = email_to_user_inputs(&email);
+    let email_ref: Vec<_> = email_parts.iter().map(String::as_str).collect();
+
+    let mut additional_inputs: HashSet<_> = additional_inputs.iter().map(String::as_str).collect();
```
dani-garcia, line 15: zxcvbn seems to be lowercasing the additional_inputs and adding them as keys to a hashmap, so we wouldn't need to use a set here, which would simplify the code a bit.

crates/bitwarden/src/auth/password/strength.rs
Hinton: No.

crates/bitwarden/src/auth/password/strength.rs
Hinton: I'm slightly hesitant to rely on internal behaviour.

dani-garcia (approved):

PR #490, opened 2024-01-09
[PM-5348] Encrypt/decrypt attachment files

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
Add encryption/decryption support for attachment files


Hinton (commented):
I think we can better encapsulate the attachment key.

crates/bitwarden/src/vault/cipher/attachment.rs
```diff
@@ -71,3 +118,70 @@ impl TryFrom<bitwarden_api_api::models::AttachmentResponseModel> for Attachment
         })
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use data_encoding::BASE64;
```
Hinton, line 124: We should use the general b64 library instead of data_encoding.

Hinton (commented):
I think we're missing the use case of creating an attachment. Which also involves creating the key.

I think what we want is a way to create a new attachment which also creates a key. And a way to decrypt existing attachments. There should be no need to use an existing key to encrypt again.

crates/bitwarden/src/mobile/vault/client_attachments.rs
```diff
@@ -0,0 +1,91 @@
...
+    },
+    Client,
+};
+
+pub struct ClientAttachments<'a> {
+    pub(crate) client: &'a Client,
+}
+
+impl<'a> ClientAttachments<'a> {
+    pub async fn decrypt_file(
```
Hinton, line 20: Nit, but the order in uniffi is different. That starts with encrypt_buffer.

Hinton (dismissed):

Hinton (approved):

PR #493, opened 2024-01-10
Hash send passwords

Hinton:
## Type of change

```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

We should hash send passwords appropriately using pbkdf.

## Code changes

## Before you submit

### DIFF
--- a/crates/bitwarden/src/crypto/master_key.rs
+++ b/crates/bitwarden/src/crypto/master_key.rs
@@ -4,10 +4,7 @@ use rand::Rng;
 use schemars::JsonSchema;
 use sha2::Digest;
 
-use super::{
-    hkdf_expand, EncString, KeyDecryptable, PbkdfSha256Hmac, SymmetricCryptoKey, UserKey,
-    PBKDF_SHA256_HMAC_OUT_SIZE,
-};
+use super::{hkdf_expand, EncString, KeyDecryptable, SymmetricCryptoKey, UserKey};
 use crate::{client::kdf::Kdf, error::Result};
 
 #[derive(Copy, Clone, JsonSchema)]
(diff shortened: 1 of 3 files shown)


dani-garcia:
Hmm, this works but I think it would be clearer if we separated the password hashing to a different function and left the encrypt/decrypt to just do encryption.

We can even have a `HashedPassword` newtype to avoid people accidentally passing an unhashed password. What do you think?

```
let password: HashedPassword = client.kdf().hash_send_password(password);
let send = Send {
    password,
    ....
};
// Now in decrypt/encrypt we can just pass the field around without modifying it.
send.encrypt().decrypt();
```

dani-garcia:
Thinking a bit more about this, we can get away with not returning a result, as this operation should never really fail. I think this _may_ work:
```rust
        let mut key: [u8; KEY_LEN] = [0u8; KEY_LEN];
        let mut mac_key: [u8; MAC_LEN] = [0u8; MAC_LEN];
        
        rng.fill(&mut key);
        rng.fill(&mut mac_key);

        SymmetricCryptoKey { key: key.into(), mac_key: Some(mac_key.into()) }
```

Hinton:
@dani-garcia There are some more nuances to this. The server will also hash the password and send back `serverHashed(clientHashed(password))`. If we do provide `serverHashed(clientHashed(password))` and it gets sent to the server it will now be `serverHashed(serverHashed(clientHashed(password)))`.

My desire with the current approach was to completely eliminate the possibility of calling `view.encrypt()` and uploading it to the server without first removing `send.password`.

dani-garcia (approved):
Ah fair enough, I forgot about the server side hashing, maybe there's something we could do with an enum like `enum SendPassword { NoPassword, AlreadySet, New(String) }`, but maybe that won't translate very nicely to the mobile bindings.

dani-garcia (approved):

PR #495, opened 2024-01-11
Refactor make_user_key to call SymmetricCryptoKey::generate

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

The `MasterKey` previously had intimate knowledge about `SymmetricCryptKey` in order to be able to generate one. This refactor extracts the knowledge back into `SymmetricCryptoKey` by creating a `SymmetricCryptoKey::generate`.

## Before you submit

### DIFF
--- a/crates/bitwarden/src/crypto/enc_string/symmetric.rs
+++ b/crates/bitwarden/src/crypto/enc_string/symmetric.rs
@@ -274,11 +274,13 @@ impl schemars::JsonSchema for EncString {
 #[cfg(test)]
 mod tests {
     use super::EncString;
-    use crate::crypto::{KeyDecryptable, KeyEncryptable, SymmetricCryptoKey};
+    use crate::crypto::{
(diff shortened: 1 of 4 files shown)


dani-garcia (dismissed):
LGTM!

Looking at it, I think the `generate(name: &str)` function is only used in tests, so we might be able to remove it, use this new one and just have one generate function.

dani-garcia:
Thinking a bit more about this, we can get away with not returning a result, as this operation should never really fail. I think this _may_ work:
```rust
        let mut key: [u8; KEY_LEN] = [0u8; KEY_LEN];
        let mut mac_key: [u8; MAC_LEN] = [0u8; MAC_LEN];
        
        rng.fill(&mut key);
        rng.fill(&mut mac_key);

        SymmetricCryptoKey { key: key.into(), mac_key: Some(mac_key.into()) }
```

dani-garcia (approved):

PR #497, opened 2024-01-12
Trusted Device Encryption

Hinton:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Implement trusted device encryption. Verified backwards compatibility with existing device keys from desktop.

## Before you submit


crates/bitwarden-crypto/src/keys/device_key.rs
```diff
@@ -0,0 +1,126 @@
...
+
+/// Device Key
+///
+/// Encrypts the DevicePrivateKey
+/// Allows the device to decrypt the UserKey, via the DevicePrivateKey.
+#[derive(Debug)]
+pub struct DeviceKey(SymmetricCryptoKey);
+
+#[derive(Debug)]
+pub struct CreateDeviceKey {
```
jlf0dev, line 14: nit: Not sure of the naming scheme here but at a glance this tells me these are the inputs needed to create a Device Key. Maybe this should be `CreatedDeviceKey`?

jlf0dev (approved):
Looks good!

crates/bitwarden-crypto/src/keys/device_key.rs
Hinton: Good point, I've renamed it to TrustDeviceResponse which is inline with other structs. We'll want to revisit this in the future.

crates/bitwarden-crypto/src/keys/device_key.rs
```diff
@@ -0,0 +1,126 @@
...
+
+        let decrypted = device_key
+            .decrypt_user_key(protected_device_private_key, protected_user_key)
+            .unwrap();
+
+        assert_eq!(decrypted.0.key, user_key.key,);
+        assert_eq!(decrypted.0.mac_key, user_key.mac_key,);
```
dani-garcia, line 124:
```suggestion
        assert_eq!(decrypted.0.key, user_key.key);
        assert_eq!(decrypted.0.mac_key, user_key.mac_key);
```

dani-garcia (dismissed):
LGTM. Some minor formatting, not sure why rustfmt didn't catch that

dani-garcia (approved):

PR #498, opened 2024-01-12
Fix lib architectures in language SDKs

### LINKED ISSUE #491: Wrong host architectures for Java bindings (darwin-x64 -> darwin-x86-64) and others as well
### Steps To Reproduce

Use Java bindings for a real world project and deploy it to different host architectures

### Expected Result

Java bindings allow to use the Bitwarden Secrets Manager SDK on all host architectures the `bitwarden_c` library was compiled for

### Actual Result

Native library (darwin-x86-64/libbitwarden_c.dylib) not found in resource path for Apple Intel (darwin-x86-64)

### Screenshots or Videos

_No response_

### Additional Context

Thanks for providing the Bitwarden Secrets Manager SDK and including Java bindings a short while ago.

I used these to build a Java application, that is a plug-in for Cryptomator, to store vault passwords in Bitwarden Secrets Manager.

Cryptomator is available for Windows, Linux and Mac, both Apple Intel and Apple Silicon. So the plug-in, using the SDK, should work on these host architectures too.

It does on Apple Silicon, but fails to load on Apple Intel, see https://github.com/purejava/cryptomator-bitwarden/issues/4.

The reason for that is, that `libbitwarden_c.dylib` cannot be found in the resource jar file, although it is contained.
Renaming the folder the libraries are stored in within the jar file from `darwin-x64` to `darwin-x86-64` solves the issue.

A note on Linux: this is probably a problem on other architectures too, as the libraries for Linux are stored in `ubuntu-x64` instead of `linux-x64`, see [here](https://github.com/purejava/cryptomator-bitwarden/issues/4#issuecomment-1884258836).

Your Node-API bindings [for the different architectures](https://github.com/bitwarden/sdk/tree/main/crates/bitwarden-napi/npm).

### Operating System

macOS

### Operating System Version

Sonoma 14.2.1

### Build Version

0.4.0-SNAPSHOT

### Issue Tracking Info

Hinton:
## Type of change

```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

- Rename `ubuntu-x64` to `linux-x64` to better match the usage.
- Resolve java using the wrong prefixes. It should be `x86_64` for architecture, and windows should use `win32`.

Resolves #491 

## Before you submit

### DIFF
--- a/.github/workflows/build-dotnet.yml
+++ b/.github/workflows/build-dotnet.yml
@@ -50,7 +50,7 @@ jobs:
         uses: actions/download-artifact@f44cd7b40bfd40b6aa1cc1b9b5b7bf03d3c67110 # v4.1.0
         with:
           name: libbitwarden_c_files-x86_64-unknown-linux-gnu
-          path: languages/csharp/Bitwarden.Sdk/ubuntu-x64
+          path: languages/csharp/Bitwarden.Sdk/linux-x64
 
       - name: Download x86_64-pc-windows-msvc files
         uses: actions/download-artifact@f44cd7b40bfd40b6aa1cc1b9b5b7bf03d3c67110 # v4.1.0
--- a/.github/workflows/build-java.yml
+++ b/.github/workflows/build-java.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+  workflow_dispatch:
 
 jobs:
   generate_schemas:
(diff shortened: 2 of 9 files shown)


Thomas-Avery (approved):
LGTM

vgrassia (approved):

PR #508, opened 2024-01-15
Add derive_pin_user_key

Hinton:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Adds support for using pin when require master password is enabled.

## Before you submit


dani-garcia (approved):

crates/bitwarden/src/mobile/crypto.rs
```diff
@@ -112,35 +114,65 @@ pub async fn get_user_encryption_key(client: &mut Client) -> Result<String> {
 #[serde(rename_all = "camelCase", deny_unknown_fields)]
 #[cfg_attr(feature = "mobile", derive(uniffi::Record))]
 pub struct DerivePinKeyResponse {
+    /// [UserKey] protected by PIN
     pin_protected_user_key: EncString,
+    /// PIN protected [UserKey]
```
dani-garcia, line 119: What do you think about this to keep it inline with `pin_protected_user_key`
```suggestion
    /// PIN protected by [UserKey]
```

dani-garcia (approved):

PR #515, opened 2024-01-19
[PM-5728] Enable Zeroize support for keys and EncStrings

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
- Implemented Zeroize support for the encryption keys and the encStrings.
- Boxed the internals of the encryption keys to avoid Rust from moving them around on the stack.
- Changed some functions that passed around owned copies of the GenericArrays to pass references instead. GenericArray implements Copy, so passing them around by value only means more stack copies if the compiler can't optimize them out.
- Changed the from impls for SymmCryptoKey to take mut slices so we can zeroize them after load.


Hinton (commented):
I suspect we need to spend some more time thinking about vectors potentially copying on resize, this might also apply to strings?

Generally good improvements, I do wonder how we'll test this efficiently though, would be nice if we could produce memory dumps of tests somehow and assert.

dani-garcia:
I've added some tests to SymmetricCryptoKey checking if the key got zeroized correctly with the help of some unsafe. It's not perfect as some of the memory gets reused, but it shows that the zeroize works at least.

crates/bitwarden-crypto/src/enc_string/asymmetric.rs
```diff
@@ -1,6 +1,9 @@
 use std::{fmt::Display, str::FromStr};
 
 use base64::{engine::general_purpose::STANDARD, Engine};
+// This module is a workaround to avoid deprecated warnings that come from the ZeroizeOnDrop
+// macro expansion
```
Hinton, line 5: Should this be moved down to the module?

crates/bitwarden-crypto/src/keys/asymmetric_crypto_key.rs
```diff
@@ -7,6 +7,9 @@ use crate::error::{CryptoError, Result};
 
 /// An asymmetric encryption key. Used to encrypt and decrypt [`EncString`](crate::EncString)
 pub struct AsymmetricCryptoKey {
+    // Similarly to SymmetricCryptoKey, we use a pinned Box to keep the key on the heap.
+    // The RsaPrivateKey type is not Copy so the compiler shouldn't be making stack copies
+    // with each move, but for some extra precaution we box it anyway
```
Hinton, line 12: Let's keep these comments standalone, as it's currently written I would need to track down symmetric to ensure I don't miss anything.

Hinton (approved):

crates/bitwarden-crypto/src/keys/device_key.rs
```diff
@@ -63,8 +63,8 @@ impl DeviceKey {
         let device_private_key: Vec<u8> = protected_device_private_key.decrypt_with_key(&self.0)?;
         let device_private_key = AsymmetricCryptoKey::from_der(device_private_key.as_slice())?;
 
-        let dec: Vec<u8> = protected_user_key.decrypt_with_key(&device_private_key)?;
-        let user_key: SymmetricCryptoKey = dec.as_slice().try_into()?;
+        let mut dec: Vec<u8> = protected_user_key.decrypt_with_key(&device_private_key)?;
+        let user_key: SymmetricCryptoKey = dec.as_mut_slice().try_into()?;
```
Hinton, line 67: I don't believe this consumes the key. Do we need to zero `dec`?

dani-garcia: We zeroize the key inside the `TryFrom<&mut [u8]>` impl:

https://github.com/bitwarden/sdk/blob/6e844469c9e1555a5abc45248ea88abcc025679b/crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs#L90-L115

Hinton: That feels potentially confusing? Do you expect a `TryFrom` of a slice to manipulate the input? I do see the benefit of ensuring it gets consumed though.

dani-garcia: We could also just pass an owned Vec<> and consume it in the function, but that will limit the use of the API for someone trying to pass a GenericArray or a [u8; 64] for example.

I think the fact that the function receives a `&mut [u8]` should make it clear enough that something is being modified, but we could add a comment mentioning explicitly that it's being zeroed.

Hinton: Let's add a comment to make it explicit. We can revaluate if this turns out to be confusing, potentially not using try from and instead having an explicit function could make it less confusing.

Hinton (approved):

Hinton (approved):

PR #518, opened 2024-01-19
Fix android not working properly

Hinton:
## Type of change

```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

It seems for uniffi to properly parse the uniffi.toml files it has to be a direct dependency and not transient like generators previously were.

## Before you submit


languages/swift/build.sh
Hinton, line 1: This should avoid us having to touch this file in the future.

withinfocus (approved):

PR #529, opened 2024-01-22
Login with device

Hinton:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Login with device and approve device.

## Before you submit


crates/bitwarden/src/mobile/crypto.rs
```diff
@@ -51,10 +51,16 @@ pub enum InitUserCryptoMethod {
         /// this.
         pin_protected_user_key: EncString,
     },
+    Passwordless {
+        request_private_key: String,
+        protected_user_key: AsymmetricEncString,
```
dani-garcia, line 56: Could we add some comments on how to get these values? Like we do with PIN and DecryptedKey?

dani-garcia (commented):
LGTM, just missing some comments about some new fields

dani-garcia (dismissed):
LGTM, the doc.md file might need to be updated with the new comments though

dani-garcia (dismissed):

jlf0dev (approved):

dani-garcia (approved):

PR #530, opened 2024-01-22
Update README.md

maxkpower:
removed pip install method for the time being

### DIFF
--- a/languages/python/README.md
+++ b/languages/python/README.md
@@ -17,12 +17,6 @@ maturin develop
 
 You can now import `BitwardenClient` in your Python code.
 
-# Use without building locally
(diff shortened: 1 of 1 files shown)


Hinton (approved):
There is another error above

```
- `maturin` (install with `pip install maturin[patchelf]`)
```

should probably be
```
- `maturin` (install with `pip install maturin`)
```

tangowithfoxtrot:
I agree with removing `[patchelf]` from the `pip install` line. That was recommended by [maturin's docs](https://www.maturin.rs/installation) (with `pipx`, not `pip` though), but it isn't needed.

PR #534, opened 2024-01-24
[DEVOPS-1333] Add code signing to the Windows bws CLI

michalchecinski:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Digitally sign Windows CLI .exe executable to prevent warning showing up on clients' computers while running `bws` commands.

## Code changes

- **.github/workflows/build-cli.yml:** 
  - split Windows and UNIX build jobs
  - add steps to windows build job to login to Azure, get secrets from KeyVault, install azuresigntool and use azuresigntool to sign windows CLI artifact.

## Before you submit


Hinton (changes requested):
Please ensure the secrets in the pipeline are optional by only requiring them if we're publishing or have the secrets set, this will ensure external PRs don't automatically fail.

I wonder if we could put the signing as a separate stage after the build which would avoid repeating the build command.

michalchecinski:
@Hinton The build command is not repeated. I split the Unix and Windows build - the steps to sign macOS and Windows are different, so to make it clear, I made different build jobs for each system. I'll make signing steps run only on main/rc branches to ensure they will work for the external PRs.

.github/workflows/build-cli.yml
```diff
@@ -30,6 +31,103 @@ jobs:
           VERSION=$(grep -o '^version = ".*"' crates/bws/Cargo.toml | grep -Eo "[0-9]+\.[0-9]+\.[0-9]+")
           echo "package_version=$VERSION" >> $GITHUB_OUTPUT
 
+      - name: Sign if branch is main or rc
+        id: sign
+        run: |
+          if [[ $GITHUB_REF == refs/heads/main || $GITHUB_REF == refs/heads/rc ]]; then
```
vgrassia, line 37: Should we also be checking for `hotfix-rc`?

.github/workflows/build-cli.yml
```diff
@@ -30,6 +31,103 @@ jobs:
...
+            echo "sign=true" >> $GITHUB_OUTPUT
+          fi
+          echo "sign=false" >> $GITHUB_OUTPUT
+
+  build-windows:
+    name: Building CLI for - ${{ matrix.settings.os }} - ${{ matrix.settings.target }}
+    runs-on: ${{ matrix.settings.os || 'ubuntu-latest' }}
+    needs:
+      - setup
```
vgrassia, line 46:
```suggestion
    needs: setup
```

vgrassia (changes requested):

Hinton:
Thanks for excluding the secrets. I'm still slightly concern with the duplication of steps, since it's easy to miss modifying one of the places.

.github/workflows/build-cli.yml
Hinton: We might also want to check if the origin is actually this repo and not a fork.

```if: github.repository_owner == 'Upvision'```

Comment (michalchecinski) [reply .github/workflows/build-cli.yml]: @Hinton That's great idea!

File: .github/workflows/build-cli.yml [hunk 3]
```diff
@@ -30,6 +31,104 @@ jobs:
           VERSION=$(grep -o '^version = ".*"' crates/bws/Cargo.toml | grep -Eo "[0-9]+\.[0-9]+\.[0-9]+")
           echo "package_version=$VERSION" >> $GITHUB_OUTPUT
 
+      - name: Sign if branch is main or rc
```
vgrassia, line 34:
```suggestion
      - name: Sign if repo is owned by Bitwarden
```

vgrassia (changes requested):

mimartin12 (approved):

mimartin12 (dismissed):

mimartin12 (approved):

vgrassia (approved):

PR #536, opened 2024-01-25
[PM-5791] Change decrypt to return Sensitive

Hinton:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Updates the decrypt methods to return a `Sensitive` value which implements `zeroize` and `zeroize on drop` to ensure the values are zeroed on drop within the rust ecosystem.

## Before you submit


crates/bitwarden-crypto/src/decrypted.rs
```diff
@@ -0,0 +1,108 @@
...
+        self.value.clone().encrypt_with_key(key)
+    }
+}
+**/
+
+impl<T: KeyEncryptable<Key, Output> + Zeroize + Clone, Key: CryptoKey, Output>
+    KeyEncryptable<Key, Output> for Decrypted<T>
+{
+    fn encrypt_with_key(self, key: &Key) -> Result<Output, CryptoError> {
+        self.value.clone().encrypt_with_key(key)
```
Hinton, line 103: Todo: need to validate what happens with the string after encrypting it. If we consume it we should zero it.

crates/bitwarden/src/vault/send.rs
```diff
@@ -200,16 +200,16 @@ impl KeyDecryptable<SymmetricCryptoKey, SendView> for Send {
...
+        let key = Send::derive_shareable_key(k.expose())?;
 
         Ok(SendView {
             id: self.id,
             access_id: self.access_id.clone(),
 
             name: self.name.decrypt_with_key(&key).ok().unwrap_or_default(),
             notes: self.notes.decrypt_with_key(&key).ok().flatten(),
-            key: Some(URL_SAFE_NO_PAD.encode(k)),
+            key: Some(URL_SAFE_NO_PAD.encode(k.expose())),
```
dani-garcia, line 212: Should we expose some `map` function on Decrypted types to make sure that the b64 encoded value is also protected?

```rust
  pub fn map_ref<W: Zeroize>(self, func: fn(&V) -> W) -> Decrypted<W> {
      Decrypted::new(Box::new(func(&self.value)))
  }

key: Some(k.map_ref(|k| URL_SAFE_NO_PAD.encode(k))) // This returns a Decrypted<String>
```

dani-garcia (commented):
Looks pretty good to me, just some small comments

crates/bitwarden/src/vault/send.rs
Hinton: Maybe, we need to be a bit careful about the return value and ensure we zeroize it to our ability.

crates/bitwarden/src/vault/send.rs
```diff
@@ -262,28 +262,29 @@ impl KeyEncryptable<SymmetricCryptoKey, Send> for SendView {
...
+        let send_key = Send::derive_shareable_key(k.expose())?;
 
         Ok(Send {
             id: self.id,
             access_id: self.access_id,
 
             name: self.name.encrypt_with_key(&send_key)?,
             notes: self.notes.encrypt_with_key(&send_key)?,
-            key: k.encrypt_with_key(key)?,
+            key: k.expose().encrypt_with_key(key)?,
```
Hinton, line 284: We should support sensitive in encrypt operations.

crates/bitwarden/src/vault/cipher/login.rs
Hinton, line 1: LoginUri needs some work.

crates/bitwarden/src/auth/auth_request.rs
```diff
@@ -57,10 +58,13 @@ pub(crate) fn auth_request_decrypt_user_key(
     private_key: String,
     user_key: AsymmetricEncString,
 ) -> Result<SymmetricCryptoKey, Error> {
-    let key = AsymmetricCryptoKey::from_der(&STANDARD.decode(private_key)?)?;
-    let mut key: Vec<u8> = user_key.decrypt_with_key(&key)?;
+    use bitwarden_crypto::DecryptedVec;
 
-    Ok(SymmetricCryptoKey::try_from(key.as_mut_slice())?)
+    let private_key = SensitiveString::new(Box::new(private_key));
```
Hinton, line 63: Follow up: Should private key be sensitive before passed in?

crates/bitwarden/src/auth/login/access_token.rs
Hinton, line 1: TODO: Review this file and explore if we can clean it up slightly.

crates/bitwarden/src/mobile/vault/client_sends.rs
```diff
@@ -48,7 +50,8 @@ impl<'a> ClientSends<'a> {
         let key = Send::get_key(&send.key, key)?;
 
         let buf = EncString::from_buffer(encrypted_buffer)?;
-        Ok(buf.decrypt_with_key(&key)?)
+        let dec: DecryptedVec = buf.decrypt_with_key(&key)?;
+        Ok(dec.expose().to_owned())
```
Hinton, line 54: Follow up: Revisit decrypt attachments

crates/bitwarden/src/vault/cipher/cipher.rs
```diff
@@ -252,25 +256,26 @@ impl Cipher {
                 let mut sub_title = String::new();
 
                 if let Some(brand) = &card.brand {
-                    let brand: String = brand.decrypt_with_key(key)?;
-                    sub_title.push_str(&brand);
+                    let brand: DecryptedString = brand.decrypt_with_key(key)?;
+                    sub_title.push_str(brand.expose());
```
Hinton, line 260: TODO: Sub_title should be sensitive and we shouldn't expose these.

Hinton (commented):
Reminders for myself.

Hinton:
I think the remaining tasks are large enough to be separate tasks. @dani-garcia let me know if you disagree. I'll write up tickets for them.

crates/bitwarden-crypto/src/sensitive/sensitive.rs
```diff
@@ -81,6 +94,12 @@ impl From<SensitiveString> for SensitiveVec {
     }
 }
 
+impl<N: ArrayLength<u8>> From<Sensitive<GenericArray<u8, N>>> for SensitiveVec {
```
dani-garcia, line 97: `GenericArray<u8, N>` can be converted into `[u8; N]` without copies by using `.into()` or `.from()` (they just do a `mem::tranmute`), so I don't think this impl is really needed. We can update the checksum function to be like this without this impl:
```rust
    pub(crate) fn generate_checksum(&mut self) {
        if let Some(uri) = &self.uri {
            let uri_hash: SensitiveVec = Sensitive::new(Box::new({
                <[u8; 32]>::from(
                    sha2::Sha256::new()
                        .chain_update(uri.expose().as_bytes())
                        .finalize(),
                )
            }))
            .into();
            self.uri_checksum = Some(uri_hash.encode_base64(STANDARD))
        }
    }
```

crates/bitwarden/src/auth/tde.rs
```diff
@@ -15,7 +15,9 @@ pub(super) fn make_register_tde_keys(
     org_public_key: String,
     remember_device: bool,
 ) -> Result<RegisterTdeKeyResponse> {
-    let public_key = AsymmetricPublicCryptoKey::from_der(&STANDARD.decode(org_public_key)?)?;
+    let public_key = AsymmetricPublicCryptoKey::from_der(SensitiveVec::new(Box::new(
+        STANDARD.decode(org_public_key)?,
```
dani-garcia, line 19: Should we convert `org_public_key` to a sensitive value before base64 decoding it?

crates/bitwarden/src/uniffi_support.rs
```diff
@@ -18,6 +18,18 @@ uniffi::ffi_converter_forward!(
     crate::UniFfiTag
 );
 
+impl UniffiCustomTypeConverter for DecryptedString {
```
dani-garcia, line 21: I don't think this is needed here, we're already forwarding the definition of SensitiveString from bitwarden-crypto and DecryptedString is just a type alias right?

dani-garcia (commented):
Mostly LGTM, just a couple of comments. 

Let's go ahead and open tickets for the other unresolved tasks, plus the SensitiveString improvements we've mentioned.

crates/bitwarden-crypto/src/sensitive/sensitive.rs
Hinton: It depends what we want to encourage, it might be tricky to catch that `from` is safe here when other from/into make copies?

crates/bitwarden/src/vault/send.rs
Hinton: https://bitwarden.atlassian.net/browse/PM-7722

crates/bitwarden/src/auth/auth_request.rs
Hinton: https://bitwarden.atlassian.net/browse/PM-7723

crates/bitwarden/src/auth/login/access_token.rs
Hinton: https://bitwarden.atlassian.net/browse/PM-7724

crates/bitwarden/src/mobile/vault/client_sends.rs
Hinton: https://bitwarden.atlassian.net/browse/PM-7725

dani-garcia (approved):

PR #559, opened 2024-01-29
SM-1082: Add SM Go SDK Wrapper

coltonhurst:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Add the ability to easily use SM state in Go to the `bitwarden_client.go`.

## Code changes

- **bitwarden_client.go:** Add the `statePath` parameter
- **example.go:** Set state in the example

## Before you submit

### DIFF
--- a/languages/go/bitwarden_client.go
+++ b/languages/go/bitwarden_client.go
@@ -45,8 +45,8 @@ func NewBitwardenClient(apiURL *string, identityURL *string) (*BitwardenClient, 
 	}, nil
 }
 
-func (c *BitwardenClient) AccessTokenLogin(accessToken string) error {
-	req := AccessTokenLoginRequest{AccessToken: accessToken}
+func (c *BitwardenClient) AccessTokenLogin(accessToken string, state_path *string) error {
+	req := AccessTokenLoginRequest{AccessToken: accessToken, StateFile: state_path}
 	command := Command{AccessTokenLogin: &req}
 
 	responseStr, err := c.commandRunner.RunCommand(command)
--- a/languages/go/example/example.go
+++ b/languages/go/example/example.go
@@ -16,6 +16,7 @@ func main() {
 	bitwardenClient, _ := sdk.NewBitwardenClient(&apiURL, &identityURL)
 
 	accessToken := os.Getenv("ACCESS_TOKEN")
+	statePath := os.Getenv("STATE_PATH")
 	organizationIDStr := os.Getenv("ORGANIZATION_ID")
 	projectName := os.Getenv("PROJECT_NAME")
 
@@ -23,7 +24,7 @@ func main() {
 		projectName = "NewTestProject" // default value
 	}
 
-	err := bitwardenClient.AccessTokenLogin(accessToken)
+	err := bitwardenClient.AccessTokenLogin(accessToken, &statePath)
 	if err != nil {
 		panic(err)
 	}


languages/go/example/example.go
```diff
@@ -16,14 +16,15 @@ func main() {
...
+	statePath := os.Getenv("STATE_PATH")
 	organizationIDStr := os.Getenv("ORGANIZATION_ID")
 	projectName := os.Getenv("PROJECT_NAME")
 
 	if projectName == "" {
 		projectName = "NewTestProject" // default value
 	}
 
-	err := bitwardenClient.AccessTokenLogin(accessToken)
+	err := bitwardenClient.AccessTokenLogin(accessToken, &statePath)
```
Thomas-Avery, line 27: When not using state, is the idea to pass `nil` or a pointer to an empty string? 

```
err := bitwardenClient.AccessTokenLogin(accessToken, nil)
```

or

```
statePath := ""
err := bitwardenClient.AccessTokenLogin(accessToken, &statePath)
```

Does it work the same way regardless?

coltonhurst: For a developer using this function, the end result is the same either way. The correct thing to do is pass `nil`, as the underlying function in Rust takes an `Option`, and `nil` maps to the variant `None`. However, in the case an empty string is passed, the SDK won't be able to read or write to a state file, so it ends up being the same thing.

Thomas-Avery (approved):
LGTM

languages/go/bitwarden_client.go
```diff
@@ -45,8 +45,8 @@ func NewBitwardenClient(apiURL *string, identityURL *string) (*BitwardenClient,
 	}, nil
 }
 
-func (c *BitwardenClient) AccessTokenLogin(accessToken string) error {
-	req := AccessTokenLoginRequest{AccessToken: accessToken}
+func (c *BitwardenClient) AccessTokenLogin(accessToken string, statePath *string) error {
```
dani-garcia, line 48: Nit: We're calling this new variable here `statePath`. In Python we've called it `state_file_path`. In Rust it's just `state_file`.

Should we decide on a consistent naming for them all?

languages/go/example/example.go
```diff
@@ -16,14 +16,15 @@ func main() {
 	bitwardenClient, _ := sdk.NewBitwardenClient(&apiURL, &identityURL)
 
 	accessToken := os.Getenv("ACCESS_TOKEN")
+	statePath := os.Getenv("STATE_PATH")
```
dani-garcia, line 19: We should mention that this is optional, like we do with the API and identity URLs

dani-garcia (commented):
We should make sure to update the README.md in the Go folder to include the new parameter to AccessTokenLogin too.

languages/go/example/example.go
coltonhurst: Updated!

languages/go/bitwarden_client.go
coltonhurst: Yes, that's a great point. I think `statePath` is fairly clear, any thoughts on that?

I have a draft PR here for the other places in the SDK & in the Python wrapper: https://github.com/bitwarden/sdk/pull/587

dani-garcia: Yeah, that works for me 👍

languages/go/example/example.go
```diff
@@ -19,11 +19,15 @@ func main() {
 	organizationIDStr := os.Getenv("ORGANIZATION_ID")
 	projectName := os.Getenv("PROJECT_NAME")
 
+	// Configuring the statePath is optional, pass nil
+	// in AccessTokenLogni() to not use state
```
dani-garcia, line 23:
```suggestion
	// in AccessTokenLogin() to not use state
```

coltonhurst: Oh, that's terrible 🤦‍♂️ Fixed, sorry about that!

dani-garcia (commented):
We're missing a reference to the old AccessTokenLogin without statepath in the Readme: 

https://github.com/bitwarden/sdk/blob/1595306d0f80807983c374aaebc4938a835be78c/languages/go/README.md?plain=1#L42

Also it still has the old return type, while now it just returns an error, so we should change that too

coltonhurst:
> We're missing a reference to the old AccessTokenLogin without statepath in the Readme:
> 
> https://github.com/bitwarden/sdk/blob/1595306d0f80807983c374aaebc4938a835be78c/languages/go/README.md?plain=1#L42
> 
> Also it still has the old return type, while now it just returns an error, so we should change that too

Good call. I've updated the README and added a very small description regarding using `statePath` or `nil`.

dani-garcia (approved):

languages/go/README.md
```diff
@@ -36,10 +36,12 @@ bitwardenClient, _ := sdk.NewBitwardenClient(&apiURL, &identityURL)
 
 ### Login
 
-To login using an access token:
+To login using an access token. Define some `statePath` and pass it to use state, or pass `nil` instead to not use state.
```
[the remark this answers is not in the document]
differsthecat: Colton is out, I'm going to merge this PR now to get it in the hands of our vendor and and create a separate PR with these comment updates.

PR #560, opened 2024-01-29
SM-1091: Use rustfmt Nightly by Default

coltonhurst:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

This PR is a proposal to add a `rust-toolchain.toml` file to the repository, and use `rustfmt` nightly by default due to https://github.com/bitwarden/sdk/pull/512.

## Code changes

- **rust-toolchain.toml:** Adds this file

## Before you submit

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,3 @@
+[toolchain]
+channel = "nightly"
+components = [ "rustfmt" ]


dani-garcia:
I don't think there's a way to write a rust-toolchain file to force nightly rustfmt but stable compiler, which is what we should prefer to use, I think.

At the moment I changed my VSCode config to run nightly rustfmt: `"rust-analyzer.rustfmt.extraArgs": "+nightly"`, and just use shell aliases for running `cargo +nightly fmt`, not great but it works.

coltonhurst:
> I don't think there's a way to write a rust-toolchain file to force nightly rustfmt but stable compiler, which is what we should prefer to use, I think.
> 
> At the moment I changed my VSCode config to run nightly rustfmt: `"rust-analyzer.rustfmt.extraArgs": "+nightly"`, and just use shell aliases for running `cargo +nightly fmt`, not great but it works.

Bummer, was trying to experiment with this and see if there was a way to do it. Thanks for the tips, I'll use that!

PR #561, opened 2024-01-30
[PM-3438] Vault exports

Hinton:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Vault export implementation.

New crate `bitwarden-exporters` which defines it's own data structure, isolating it from the other code areas. It primarily focuses on converting a data model to csv and json. Currently all data has to be decrypted independently of what is required for the export which is an area of improvement for the future.

## Code changes

## Before you submit


crates/bitwarden-exporters/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+    pub passport_number: Option<String>,
+    pub license_number: Option<String>,
+}
+
+pub fn export(folders: Vec<Folder>, ciphers: Vec<Cipher>, format: Format) -> String {
+    match format {
+        Format::Csv => export_csv(folders, ciphers).unwrap(),
+        Format::Json => export_json(folders, ciphers).unwrap(),
+        // Format::EncryptedJson { password } => export_encrypted_json(folders, ciphers, password),
+        _ => todo!(),
```
dani-garcia, line 120: I think we're better off commenting the `EncryptedJson` definition in the enum and removing the `_` branch here, we also should just return a result instead of unwrapping.

crates/bitwarden/src/vault/cipher/field.rs
```diff
@@ -34,11 +34,11 @@ pub struct Field {
...
 pub struct FieldView {
-    name: Option<String>,
-    value: Option<String>,
-    r#type: FieldType,
+    pub name: Option<String>,
+    pub value: Option<String>,
+    pub r#type: FieldType,
 
-    linked_id: Option<LinkedIdType>,
+    pub linked_id: Option<LinkedIdType>,
```
dani-garcia, line 41: I think these can be `pub(crate)`? Same in some other structs as well

crates/bitwarden/src/vault/mod.rs
```diff
@@ -6,12 +6,7 @@ mod send;
 #[cfg(feature = "mobile")]
 mod totp;
 
-pub use cipher::{
-    attachment::{
-        Attachment, AttachmentEncryptResult, AttachmentFile, AttachmentFileView, AttachmentView,
-    },
-    Cipher, CipherListView, CipherView,
-};
+pub use cipher::*;
```
dani-garcia, line 9: So far we've shied away from glob imports, specially for public API, should we change this back?

crates/bitwarden/src/tool/exporters/mod.rs
```diff
@@ -13,21 +19,26 @@ pub use client_exporter::ClientExporters;
...
-        ExportFormat::Csv => "Csv".to_owned(),
-        ExportFormat::Json => "Json".to_owned(),
-        ExportFormat::AccountEncryptedJson => "AccountEncryptedJson".to_owned(),
-        ExportFormat::EncryptedJson { .. } => "EncryptedJson".to_owned(),
-    })
+    let enc = client.get_encryption_settings()?;
+
+    let folders: Vec<FolderView> = folders.decrypt(enc, &None)?;
+    let folders: Vec<bitwarden_exporters::Folder> =
+        folders.into_iter().flat_map(|f| f.try_into()).collect();
```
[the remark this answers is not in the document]
Hinton: I think so, it shouldn't fail unless the data is malformed since each folder should have an ID.

crates/bitwarden/src/vault/mod.rs
Hinton: I'm split. It's a pain to re-export everything in every layer, the `cipher` is already a mod file.

crates/bitwarden-exporters/src/csv.rs
```diff
@@ -0,0 +1,266 @@
...
+pub enum CsvError {
+    #[error("CSV error")]
+    Csv,
+}
+
+pub(crate) fn export_csv(folders: Vec<Folder>, ciphers: Vec<Cipher>) -> Result<String, CsvError> {
+    let folders: HashMap<Uuid, String> = folders.iter().map(|f| (f.id, f.name.clone())).collect();
+
+    let rows = ciphers
+        .iter()
```
dani-garcia, line 20: If we change this to `.into_iter()`, we can remove some clones below, in name, notes and fields at least

crates/bitwarden-exporters/src/json.rs
```diff
@@ -0,0 +1,755 @@
...
+            name: cipher.name,
+            notes: cipher.notes,
+            r#type,
+            login,
+            identity,
+            card,
+            secure_note,
+            favorite: cipher.favorite,
+            reprompt: cipher.reprompt,
+            fields: cipher.fields.iter().map(|f| f.clone().into()).collect(),
```
dani-garcia, line 252: By using `.into_iter()` here we can remove the clone

crates/bitwarden-exporters/src/lib.rs
```diff
@@ -0,0 +1,129 @@
...
+    pub company: Option<String>,
+    pub email: Option<String>,
+    pub phone: Option<String>,
+    pub ssn: Option<String>,
+    pub username: Option<String>,
+    pub passport_number: Option<String>,
+    pub license_number: Option<String>,
+}
+
+pub fn export(folders: Vec<Folder>, ciphers: Vec<Cipher>, format: Format) -> String {
```
dani-garcia, line 123: I think this should return result if there is a chance of it failing

dani-garcia (approved):

PR #566, opened 2024-01-31
Implement password update support

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
Support re-encrypting the user key with the new password.

### DIFF
--- a/crates/bitwarden/src/mobile/crypto.rs
+++ b/crates/bitwarden/src/mobile/crypto.rs
@@ -129,6 +129,53 @@ pub async fn get_user_encryption_key(client: &mut Client) -> Result<String> {
 #[derive(Serialize, Deserialize, Debug, JsonSchema)]
 #[serde(rename_all = "camelCase", deny_unknown_fields)]
 #[cfg_attr(feature = "mobile", derive(uniffi::Record))]
+pub struct UpdatePasswordResponse {
+    /// Hash of the new password
+    password_hash: String,
+    /// User key, encrypted with the new password
+    new_key: EncString,
+}
+
+pub fn update_password(
+    client: &mut Client,
+    new_password: String,
+) -> Result<UpdatePasswordResponse> {
+    let user_key = client
+        .get_encryption_settings()?
+        .get_key(&None)
+        .ok_or(Error::VaultLocked)?;
+
+    let login_method = client
+        .login_method
+        .as_ref()
+        .ok_or(Error::NotAuthenticated)?;
+
+    // Derive a new master key from password
+    let new_master_key = match login_method {
+        LoginMethod::User(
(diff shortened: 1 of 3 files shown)


crates/bitwarden/src/mobile/crypto.rs
```diff
@@ -125,6 +125,53 @@ pub async fn get_user_encryption_key(client: &mut Client) -> Result<String> {
...
+        .as_ref()
+        .ok_or(Error::NotAuthenticated)?;
+
+    // Derive a new master key from password
+    let new_master_key = match login_method {
+        LoginMethod::User(
+            UserLoginMethod::Username { email, kdf, .. }
+            | UserLoginMethod::ApiKey { email, kdf, .. },
+        ) => MasterKey::derive(new_password.as_bytes(), email.as_bytes(), kdf)?,
+        _ => return Err(Error::NotAuthenticated),
```
Hinton, line 159: We should probably have another error type for these scenarios. You technically are authenticated it's just not compatible with the current operation.

crates/bitwarden/src/mobile/crypto.rs
```diff
@@ -194,6 +241,66 @@ mod tests {
...
+                method: InitUserCryptoMethod::Password {
+                    password: "asdfasdfasdf".into(),
+                    user_key: "2.u2HDQ/nH2J7f5tYHctZx6Q==|NnUKODz8TPycWJA5svexe1wJIz2VexvLbZh2RDfhj5VI3wP8ZkR0Vicvdv7oJRyLI1GyaZDBCf9CTBunRTYUk39DbZl42Rb+Xmzds02EQhc=|rwuo5wgqvTJf3rgwOUfabUyzqhguMYb3sGBjOYqjevc=".into(),
+                },
+            },
+        )
+        .await
+        .unwrap();
+
+        let new_password_response = update_password(&mut client, "123412341234".into()).unwrap();
```
Hinton, line 267: Do we want to validate the new password hash too? We have a `validate_password` that could be used.

dani-garcia: Hmm, `validate_password` is validating a `LocalAuthorization` hash, while we would be returning a `ServerAuthorization` hash from the `update_password` function, right?

crates/bitwarden/src/mobile/crypto.rs
dani-garcia: Yeah I went with this one because it was used in other places too, but this should probably be `Error::InvalidAuthentication` or something. We can change it but it's going to affect other places too, so I can leave it for another PR right after this one.

Hinton (approved):

PR #568, opened 2024-01-31
Implement alternative login with device flow

Hinton:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

When approving from other devices we need to support the master key flow. Also implements login with device for the CLI as a testing tool.

## Before you submit


crates/bitwarden/src/client/client.rs
```diff
@@ -86,7 +86,11 @@ impl Client {
     pub fn new(settings_input: Option<ClientSettings>) -> Self {
         let settings = settings_input.unwrap_or_default();
 
-        let headers = header::HeaderMap::new();
+        let mut headers = header::HeaderMap::new();
+        headers.append(
+            "Device-Type",
+            HeaderValue::from_str(&(settings.device_type as u8).to_string()).unwrap(),
+        );
```
Hinton, line 93: @dani-garcia I can't remember if we re-used the configs for email aliases? I needed this for auth calls to behave correctly. If so we should probably expose a 3rd client that can be used for external services.

dani-garcia: Yeah at the moment we're using the same HTTP client for api, identity and the forwarders. It would be nice if the swagger generated code handled adding that header when needed, but yeah exposing a separate client without any headers set sounds good to me.

dani-garcia (approved):

PR #571, opened 2024-02-01
Use non-root user in Docker

tangowithfoxtrot:
## Type of change

- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

Use a non-root user in Docker. There doesn't seem to be an agreed upon standard for what to name the non-root user, so I named it _app_, as I followed the general guidance from [this MS blog post](https://devblogs.microsoft.com/dotnet/securing-containers-with-rootless/).

A `BWS_CONFIG_FILE` environment variable was added to `bws` as well to make it easier to avoid guessing where the correct config file dir is if you are unaware of the _app_ user's name (and therefor, their home directory).

## Code changes

- **crates/bws/Dockerfile:** Use a non-root user.
- **crates/bws/README.md:** Document using the CLI with Docker.
- **crates/bws/src/main.rs:** Add `BWS_CONFIG_FILE` environment variable.
- **crates/bws/CHANGELOG.md:** Update changelog.

## Before you submit


crates/bws/README.md
```diff
@@ -44,3 +44,45 @@ echo 'source <(/path/to/bws completions bash)' >> ~/.bashrc
...
+```bash
+# Set the BWS_CONFIG_FILE environment variable on your host
+export BWS_CONFIG_FILE="/path/to/config/file"
+
+# Pass the BWS_CONFIG_FILE environment variable to the container
+docker run --rm -it -e BWS_CONFIG_FILE="$BWS_CONFIG_FILE" -v "$BWS_CONFIG_FILE":"$BWS_CONFIG_FILE" bitwarden/bws --help
+```
+
+Note that if you want to use identical config file paths on your host and in the container, the
+parent directory must exist on both.
```
Hinton, line 88: I think we can shorten this down a fair bit. The readme is only really visible on https://crates.io/crates/bws. We can link to the docker concepts instead of having to explain it ourselves. And/or link to our help pages where we explain this in more details.

## Docker

We also provide a docker image preloaded with the `bws` cli.

```bash
# From the root of the repository
docker build -f crates/bws/Dockerfile -t bitwarden/bws .

docker run --rm -it bitwarden/bws --help
```

To use a configuration file, utilize docker [bind mounting](https://docs.docker.com/storage/bind-mounts/) to expose it to the container.

# You may also provide a config file
docker run --rm -it -v "$HOME"/.bws:/home/app/.bws bitwarden/bws --help
```

dani-garcia (approved):
LGTM now

Hinton (approved):

coltonhurst (approved):
Thanks for adding the `BWS_CONFIG_FILE` option 🎉

PR #572, opened 2024-02-02
Support validating password by decrypting user key

Hinton:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Enables password validation without knowing the password hash by decrypting the user key.

## Before you submit

### DIFF
--- a/crates/bitwarden/src/auth/password/validate.rs
+++ b/crates/bitwarden/src/auth/password/validate.rs
@@ -8,7 +8,7 @@ use crate::{
 };
 
 /// Validate if the provided password matches the password hash stored in the client.
-pub(crate) async fn validate_password(
+pub(crate) fn validate_password(
     client: &Client,
     password: String,
     password_hash: String,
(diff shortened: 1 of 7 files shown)


crates/bitwarden/src/auth/password/validate.rs
```diff
@@ -34,8 +34,49 @@ pub(crate) async fn validate_password(
...
+                    .map_err(|_| "wrong password")?;
+
+                let enc = client
+                    .get_encryption_settings()
+                    .map_err(|_| Error::VaultLocked)?;
+
+                let existing_key = enc.get_key(&None).ok_or(Error::VaultLocked)?;
+
+                if user_key.to_vec() != existing_key.to_vec() {
+                    return Err("wrong password".into());
```
dani-garcia, line 64: I think this error would be better as "wrong user key" or something like that, if the password was wrong we'd fail when decrypting already.

crates/bitwarden/src/auth/password/validate.rs
```diff
@@ -59,4 +100,63 @@ mod tests {
...
+
+        client
+            .initialize_user_crypto("asdfasdfasdf", user_key.parse().unwrap(), private_key)
+            .unwrap();
+
+        let result =
+            validate_password_user_key(&client, "asdfasdfasdf".to_string(), user_key.to_string())
+                .unwrap();
+
+        assert_eq!(result, "aOvkBXFhSdgrBWR3hZCMRoML9+h5yRblU3lFphCdkeA=");
```
dani-garcia, line 131: We could also call `validate_password` here to validate it against the hash and make sure it returns true

dani-garcia (approved):

PR #588, opened 2024-02-06
[PM-3438] Encrypted export

Hinton:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Implement encrypted export

## Before you submit

### DIFF
--- a/crates/bitwarden-crypto/src/keys/master_key.rs
+++ b/crates/bitwarden-crypto/src/keys/master_key.rs
@@ -1,16 +1,11 @@
-use std::{num::NonZeroU32, pin::Pin};
+use std::num::NonZeroU32;
 
-use aes::cipher::typenum::U32;
 use base64::{engine::general_purpose::STANDARD, Engine};
-use generic_array::GenericArray;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
-use sha2::Digest;
 
-use crate::{
-    util::{self, hkdf_expand},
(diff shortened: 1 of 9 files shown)


crates/bitwarden-crypto/src/keys/master_key.rs
```diff
@@ -77,6 +77,16 @@ impl MasterKey {
...
+
+    pub fn encrypt(&self, data: &[u8]) -> Result<EncString> {
+        let stretched_key = stretch_master_key(self)?;
+
+        EncString::encrypt_aes256_hmac(
+            data,
+            stretched_key.mac_key.as_ref().unwrap(),
+            &stretched_key.key,
+        )
+    }
```
Hinton, line 89: We should probably swap this out for key encryptable.

crates/bitwarden-exporters/src/encrypted_json.rs
```diff
@@ -0,0 +1,246 @@
...
+            1,
+            iterations.get(),
+            Some(memory.get()),
+            Some(parallelism.get()),
+        ),
+    };
+
+    let salt: [u8; 16] = generate_random_bytes();
+    let salt = STANDARD.encode(salt);
+    let key = MasterKey::derive(password.as_bytes(), salt.as_bytes(), &kdf)?;
```
Hinton, line 48: We used to call this PIN, so open ended question is if we should use master key here, or define a new primitive.

dani-garcia (dismissed):

dani-garcia (approved):

PR #589, opened 2024-02-06
Update MSRV to 1.71 and add CI test

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Update MSRV in Cargo.toml and Readme to be up to date with the dependencies that we're using, plus add a test to check that we don't deviate from it.

The dependency causing the bump is `env_logger` with a MSRV of 1.71 (released July 2023) https://github.com/rust-cli/env_logger/blob/9303b0c0393c33046a791b0a6497b0f03ef1f434/Cargo.toml#L8

The issue seems to also be affecting other crates in the workspace that don't depend on that crate because the MSRV bump is due to some cargo changes regarding `dep:` features and so it seems to affect the version resolution of the entire workspace.

Note that this required a small change to exporters, apparently older rust versions have trouble disambiguating the `csv` crate with the `csv` module.


Hinton (dismissed):

.github/workflows/minimum-rust-version.yml
```diff
@@ -0,0 +1,46 @@
...
+          toolchain: 1.71.0
+          targets: ${{ matrix.settings.target }}
+
+      - name: Cache cargo registry
+        uses: Swatinem/rust-cache@23bce251a8cd2ffc3c1075eaa2367cf899916d84 # v2.7.3
+        with:
+          key: msrv-${{ matrix.settings.target }}-cargo-${{ matrix.settings.os }}
+
+      - name: cargo check MSRV
+        run: cargo check --all-features
```
Hinton, line 46: We probably only want to check the `bitwarden` crate and it's dependencies since we will build the CLIs and other bindings on latest.

Hinton (approved):

vgrassia (approved):

PR #590, opened 2024-02-07
Update swagger generated API crates

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Update swagger generator to version 7.2.0, pull in the new templates and generate the new API crates.

I've also updated the build instructions to use .NET 8, as is required in the latest server code.

The version of the server code used to generate this is https://github.com/bitwarden/server/commit/fc1d7c70592cfbc65e61c047bcc138850d91a55b
I've had to make some small modifications to the server code to get it working, ~~and I'll link the PR here shortly.~~ https://github.com/bitwarden/server/pull/3760

To make this easier to review, I've split this into three commits:
- b42f189924a770aa127b88478876548e8beb3879: Update the swagger and dotnet versions, and pull in the new templates as-is. Note that this won't generate working code.
- 9727d462924f9e533c0e3608ff0ba8aaeb0a2124: Patched the provided templates to make it work with our code. This is in a separate commit to make it easier to see what we've changed and hopefully make it easier in the future to update these templates. The changes are:
    - Ignore warnings by default 
    - Use Uuids instead of strings
    - Use serde_repr to support int enums
    - Disabled default features for reqwest (to remove default tls)
    - Removing double optionals (This could be reverted once we're better at marking everything as required in the server)
- ~~f9b63606477d0f0f987b8534ef6e6755c76003a2: New autogenerated swagger code, no manually edited files here.~~ Moved to a separate PR (#593)


support/openapi-template/hyper/api.mustache
```diff
@@ -38,7 +38,7 @@ impl<C: hyper::client::connect::Connect>{{{classname}}} for {{{classname}}}Clien
...
+    fn {{{operationId}}}(&self, {{#allParams}}{{{paramName}}}: {{^required}}Option<{{/required}}{{#required}}{{#isNullable}}Option<{{/isNullable}}{{/required}}{{#isString}}{{^isUuid}}&str{{/isUuid}}{{/isString}}{{#isUuid}}&str{{/isUuid}}{{^isString}}{{^isUuid}}{{^isPrimitiveType}}{{^isContainer}}crate::models::{{/isContainer}}{{/isPrimitiveType}}{{{dataType}}}{{/isUuid}}{{/isString}}{{^required}}>{{/required}}{{#required}}{{#isNullable}}>{{/isNullable}}{{/required}}{{^-last}}, {{/-last}}{{/allParams}}) -> Pin<Box<dyn Future<Output = Result<{{^returnType}}(){{/returnType}}{{#returnType}}{{{.}}}{{/returnType}}, Error>>>> {
```
Hinton, line 41: Are we missing `uuid` here? Or is this no longer needed.

Hinton (commented):
Looks like we might be removing some of our custom logic around uuids?

support/openapi-template/hyper/api.mustache
dani-garcia: We don't use the hyper variant, which I think is for server code generation. So I left that part untouched, we could remove the whole folder as well, but I left it exactly the same as the openapitools repo.

Hinton (approved):

PR #591, opened 2024-02-07
[PM-6104] Add locking support to bitwarden-json to improve bindings thread safety

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
I've added the lock into bitwarden-json because we already do locking in bitwarden-wasm and this way we can remove it from there.

### DIFF
--- a/crates/bitwarden-json/src/client.rs
+++ b/crates/bitwarden-json/src/client.rs
@@ -1,3 +1,4 @@
+use async_lock::Mutex;
 use bitwarden::client::client_settings::ClientSettings;
 
 #[cfg(feature = "secrets")]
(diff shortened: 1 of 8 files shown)


crates/bitwarden-json/src/client.rs
```diff
@@ -44,41 +45,43 @@ impl Client {
             }
         };
 
+        let mut client = self.0.lock().await;
```
Hinton, line 48: Bit un-intuitive that we never have to release the lock and it's handled on drop.

Hinton (approved):

PR #621, opened 2024-02-21
[SM-1129] Run command with secrets

tangowithfoxtrot:
## Type of change

- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

Add a `run` command to allow running processes with secrets injected.

Example:
`bws run 'docker compose up -d'`, `bws run -- docker compose up -d`, or from stdin: `echo 'docker compose up -d' | bws run`

Where the compose file is expecting secrets:
```yaml
services:
  echo:
    image: hashicorp/http-echo
    container_name: echo
    ports:
      - "127.0.0.1:5678:5678"
    command: [
        "-text",
        "Local DB user: ${LOCAL_DB_USER}\nLocal DB pass: ${LOCAL_DB_PASS}", # secrets from Secrets Manager
      ]

```

Other examples: `bws run -- npm run start`, `bws run -- 'echo $SECRET_BY_NAME_FROM_SM'`, etc.

A `--shell` option is provided to override the default shell (`sh` on UNIX-like OSes, and `powershell` on Windows) where the process is executed.

A `--no-inherit-env` option is provided for additional safety in cases where you want to pass the minimum amount of values into a process. `$PATH` is always passed though, as omitting it would cause nearly every command to fail.

If duplicate keynames are detected, the `run` command will error-out and suggest using the `--uuids-as-keynames` argument. This argument (and equivalent environment variable `BWS_UUIDS_AS_KEYNAMES`) will use the secret UUID (in POSIX-compliant form; ex `_36527bf9_ed6c_41ad_ba49_b11d00b371f4`).

## Code changes

- **crates/bws/src/command/run.rs:** add the `run` command and associated args
- **crates/bws/src/cli.rs:** add args for `--shell`, `--no-inherit-env`, and `--uuids-as-keynames`; add environment variable for `BWS_UUIDS_AS_KEYNAMES`
- **crates/bws/src/util.rs:** add `is_valid_posix_name` and `uuid_to_posix` functions
- **crates/bws/src/render.rs:** use `is_valid_posix_name` from `util` crate
- **crates/bws/Cargo.toml:** use `which` to detect presence of a shell

## Before you submit


crates/bws/src/main.rs
```diff
@@ -603,6 +628,108 @@ async fn process_commands() -> Result<()> {
...
+            } else {
+                command.envs(&environment);
+            }
+
+            let mut child = command.spawn().expect("failed to execute process");
+
+            let exit_status = child.wait().expect("process failed to execute");
+            let exit_code = exit_status.code().unwrap_or(1);
+
+            let _ = child.wait();
```
dani-garcia, line 726: I don't think this is needed, we're already calling child.wait() above, any other calls will just return immediately, I think

crates/bws/src/main.rs
```diff
@@ -603,6 +628,108 @@ async fn process_commands() -> Result<()> {
...
+                _ => unreachable!(),
+            };
+
+            if which(&shell).is_err() {
+                eprintln!("Error: shell '{}' not found", shell);
+                std::process::exit(1);
+            }
+
+            let user_command = if command.is_empty() {
+                if atty::is(Stream::Stdin) {
```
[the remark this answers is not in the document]
tangowithfoxtrot: I intended for this to only work as a non-interactive command, so something like `echo 'output from some other script or input' | bws run` _should_ work, but it should print the help if you try to run it interactively, in the examples that you provided.

I did this because it felt like letting someone pop an interactive shell with secrets in it could encourage risky behavior, like using `bws` as a shell itself, outside of the intended use-cases.

crates/bws/src/main.rs
```diff
@@ -222,6 +238,15 @@ enum DeleteCommand {
     Secret { secret_ids: Vec<Uuid> },
 }
 
+#[derive(Subcommand, Debug)]
+enum RunCommand {
```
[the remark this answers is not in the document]
tangowithfoxtrot: Whoops. You're right. Removed it.

crates/bws/src/main.rs
```diff
@@ -404,6 +409,102 @@ async fn process_commands() -> Result<()> {
...
+                .await?
+                .data;
+
+            let environment = secrets
+                .into_iter()
+                .map(|s| (s.key, s.value))
+                .collect::<std::collections::HashMap<String, String>>();
+
+            for key in environment.keys() {
+                if !is_valid_posix_name(key) {
```
mzieniukbw, line 472: will shell start with invalid environment variables ? don't we need to filter those out ?

crates/bws/src/main.rs
```diff
@@ -404,6 +409,102 @@ async fn process_commands() -> Result<()> {
...
+                .stdout(process::Stdio::inherit())
+                .stderr(process::Stdio::inherit());
+
+            if no_inherit_env {
+                let path = std::env::var("PATH").unwrap_or_else(|_| "/bin:/usr/bin".to_string());
+                command.env_clear();
+                command.env("PATH", path); // PATH is always necessary
+                command.envs(&environment);
+            } else {
+                command.envs(&environment);
```
mzieniukbw, line 493: for additional security, do we want to remove "BWS_ACCESS_TOKEN" from inherited env variables ?

mzieniukbw (commented):
Nice! 

Few questions: 
1. Will this work with quoted args and multiline ?
Examples:
```
bws run docker run --platform "linux/amd64" --rm --name "ubuntu-1" ubuntu echo 'hello'

bws run if [ "$ENV1" = "1234" ]; then
echo "hello";
fi
```

2. Is creating the environment variables secrets with `key` good idea, since it is not guaranteed to be unique within one and across multiple projects and does not follow posix env format ?
Maybe it should be uuid by default (with replaced `-` characters to underscores `_`) with cli option like `--use-secret-key` to additionally expand to env variable by key (to be more user friendly, maybe even prefixed by project name or id ?)

crates/bws/src/main.rs
tangowithfoxtrot: > 1. Will this work with quoted args and multiline ?
> ```
> bws run docker run --platform "linux/amd64" --rm --name "ubuntu-1" ubuntu echo 'hello'
> 
> bws run if [ "$ENV1" = "1234" ]; then
> echo "hello";
> fi
> ```

Great questions!

In the first case, I'd use an "end of arguments" (`--`) operator to construct a command like that to ensure that the shell is passing all of the arguments to our command: `bws run -- docker run --platform "linux/amd64" --rm --name "ubuntu-1" ubuntu echo 'hello'`.

In the case of the multi-line `if` statement, I'm not sure if there's a way for `bws` to handle that, as the first new-line passed to the shell would immediately attempt to execute `bws run if [ "$ENV1" = "1234" ]; then` on its own. I'm unsure if anything other than a shell builtin could trigger the interactive multiline input.

I wouldn't generally use the `run` command to execute ad-hoc shell commands like that, but if you wanted to, you could wrap it in single-quotes to avoid having the shell executing or interpolating anything before it's passed to our command. Something like the following would work:
```bash
bws run -- 'if [ "$ENV1" = "1234" ]; then
echo "hello";
fi'
```

> 2. Is creating the environment variables secrets with `key` good idea, since it is not guaranteed to be unique within one and across multiple projects and does not follow posix env format ?

I would prefer keeping the default behavior to use the key names, as it greatly improves usability and is in-line with CLIs that have similar functionality. Making UUIDs the default would require more effort for users to use the `run` command to replace things like local env files, where the environment variables wouldn't have a Bitwarden-generated secret UUID to reference.

>  Maybe it should be uuid by default (with replaced - characters to underscores _) with cli option like --use-secret-key to additionally expand to env variable by key (to be more user friendly, maybe even prefixed by project name or id ?)

What would you think of a `BWS_USE_UNIQUE_NAMES=1` environment variable that could be set that would ensure that all secrets that are set get the secret UUID (in snake_case) appended to them? So, if you're running a script where you have secrets with identical names, you can ensure uniqueness. We could possibly add an equivalent argument (`--use-unique-names`?), but I feel like you would be much more likely to need that functionality on a per-script/session basis, rather than as a one-off. Either way, I think having something like that would help people that can't guarantee the uniqueness of their secret names.

crates/bws/src/main.rs
tangowithfoxtrot: I have also wondered this. I added the `--no-inherit-env` argument for a similar reason, but I suppose I could remove that variable by default.

crates/bws/src/main.rs
tangowithfoxtrot: > will shell start with invalid environment variables ? don't we need to filter those out ?

It will set the environment variables, even if they're not POSIX-compliant, yes. You'll get warnings printed to stderr stating that you have some non-POSIX-compliant key names, but they are technically set correctly by the `run` command. I figured that if someone were to executing a binary/script that had less restrictive naming conventions for environment variables, we could leave them set. For example:

```bash
bws run -- 'python -c "import os; print(os.getenv("1 non-compliant keyname"))"'
```

would work, since `getenv` doesn't seem to require POSIX compliance. I imagine the same is possible for equivalent libs in other languages.

I'm open to un-setting them, but I'm not sure how people might feel about that if they're used to not being limited by POSIX-compliance.

mzieniukbw: > --use-unique-name

I like this, being explicit in the cli options is better. 
And like you mentioned, makes the solution more future proof, when users wants to do things by Id.

crates/bws/src/main.rs
mzieniukbw: @bwdil thoughts ?

crates/bws/src/main.rs
mzieniukbw: > I'm open to un-setting them, but I'm not sure how people might feel about that if they're used to not being limited by POSIX-compliance.

I am fine either way, just want to avoid `bws run` not working because of some special characters in the secret key.

crates/bws/src/main.rs
bwdil: @mzieniukbw please unpack what you mean by 'for additional security' what is your concern? About having `BWS_ACCESS_TOKEN` inherited and how it gets subsequently used thereafter? 

What are the security gains you're trying to achieve and what's the risk?

mzieniukbw: @bwdil My though is that whatever get's executed as part of `bws run` is probably something that needs to run one off and just need read access to some of the secrets.
With `BWS_ACCESS_TOKEN` we expose much more than intended:
- Access to all secrets that the token have access to - `bws run` can specify a project, but with access token we have access to every single of them
- With access token you can do write operations - which is really bad in case it get's to wrong hands.

bwdil: @mzieniukbw if `BWS_ACCESS_TOKEN` is not required for subsequent commands that get spawned then it can be cleared from env to limit exposure. I think your concern is something that `bws run` invokes (another process) getting hijacked and then stealing the access token.

tangowithfoxtrot: `BWS_ACCESS_TOKEN` is now unset by default.

mzieniukbw (dismissed):
LGTM 👍 
I would appreciate more UTs though 😅

mzieniukbw (approved):

crates/bws/src/command/run.rs
```diff
@@ -0,0 +1,135 @@
...
+                    "Error: multiple secrets with name '{}' found. Use --uuids-as-keynames or use unique names for secrets.",
+                    key
+                );
+                std::process::exit(1);
+            }
+            false => {
+                environment.insert(key, s.value);
+            }
+        }
+    });
```
Hinton, line 106: **suggestion:** It's generally more idiomatic to use map or flatmap and collect.

This could look something like

```rust
let evironment = secres
    .into_iter()
    .map(|s| {
        if uuids_as_keynames {
            uuid_to_posix(&s.id)
        } else {
            s.key
        }
    })
    .inspect(|s| {
        if !is_valid_posix_name(&key) {
            eprintln!(
                "Warning: secret '{}' does not have a POSIX-compliant name",
                key
            );
        }
    });

// Using itertools https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.all_unique
if !environment.all_unique() {
    return Err("Not unique)
}
```

crates/bws/src/command/run.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    let mut command = process::Command::new(shell);
+    command
+        .arg("-c")
+        .arg(&user_command)
+        .stdout(process::Stdio::inherit())
+        .stderr(process::Stdio::inherit());
+
+    if no_inherit_env {
+        let path = std::env::var("PATH").unwrap_or_else(|_| "/bin:/usr/bin".to_string());
```
Hinton, line 116: **question:** Do we need to handle windows separately?

crates/bws/src/command/run.rs
```diff
@@ -0,0 +1,135 @@
...
+        command.env("PATH", path); // PATH is always necessary
+        command.envs(&environment);
+    } else {
+        command.env_remove(ACCESS_TOKEN_KEY_VAR_NAME);
+        command.envs(&environment);
+    }
+
+    let mut child = command.spawn().expect("failed to execute process");
+
+    let exit_status = child.wait().expect("process failed to execute");
```
Hinton, line 127: **suggestion:** We should return an Error instead of panicing here.

Hinton (changes requested):
Looks like a good start.

This review follows [Conventional Comments](https://conventionalcomments.org/)

crates/bws/Cargo.toml
```diff
@@ -51,6 +53,8 @@ clap_complete = "4.5.2"
 clap_mangen = "0.2.20"
 uuid = { version = "1.7.0" }
 
+which = "6.0.0"
```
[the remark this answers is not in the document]
tangowithfoxtrot: It wasn't. Good catch. Fixed in https://github.com/bitwarden/sdk/pull/621/commits/14a9e50c602f2ee1d048e7460342109e16126644.

crates/bws/Cargo.toml
```diff
@@ -15,6 +15,7 @@ repository.workspace = true
 license-file.workspace = true
 
 [dependencies]
+atty = "0.2.14"
```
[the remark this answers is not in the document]
tangowithfoxtrot: Fixed in a836b65dbdcfce092c49e20d5326bf7e1a0f2e02.

crates/bws/src/cli.rs
```diff
@@ -10,6 +10,8 @@ pub(crate) const CONFIG_FILE_KEY_VAR_NAME: &str = "BWS_CONFIG_FILE";
 pub(crate) const PROFILE_KEY_VAR_NAME: &str = "BWS_PROFILE";
 pub(crate) const SERVER_URL_KEY_VAR_NAME: &str = "BWS_SERVER_URL";
 
+pub(crate) const UUIDS_AS_KEYNAMES: &str = "BWS_UUIDS_AS_KEYNAMES";
```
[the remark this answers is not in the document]
tangowithfoxtrot: Fixed in c93bf8fb198132c34ae6a52d3d96735f19f9b1f6.

crates/bws/src/util.rs
```diff
@@ -0,0 +1,52 @@
...
+pub(crate) fn is_valid_posix_name(input_text: &str) -> bool {
+    match Regex::new(VALID_POSIX_NAME_REGEX) {
+        Ok(r) => r.is_match(input_text),
+        Err(_) => false,
+    }
+}
+
+pub(crate) fn uuid_to_posix(uuid: &Uuid) -> String {
+    // POSIX environment variable names must start with a letter or an underscore
+    // and can only contain letters, numbers, and underscores.
```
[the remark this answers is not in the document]
tangowithfoxtrot: Fixed in 5be08ed13f91ecf6dd011d5280f4d2a6ff06521a.

crates/bws/src/cli.rs
```diff
@@ -10,6 +10,8 @@ pub(crate) const CONFIG_FILE_KEY_VAR_NAME: &str = "BWS_CONFIG_FILE";
 pub(crate) const PROFILE_KEY_VAR_NAME: &str = "BWS_PROFILE";
 pub(crate) const SERVER_URL_KEY_VAR_NAME: &str = "BWS_SERVER_URL";
 
+pub(crate) const UUIDS_AS_KEYNAMES_VAR_NAME: &str = "BWS_UUIDS_AS_KEYNAMES";
```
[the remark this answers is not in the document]
tangowithfoxtrot: Fixed in 3df1c6d.

crates/bws/Cargo.toml
```diff
@@ -43,6 +43,7 @@ thiserror = "1.0.57"
 tokio = { version = "1.36.0", features = ["rt-multi-thread", "macros"] }
 toml = "0.8.10"
 uuid = { version = "1.7.0", features = ["serde"] }
+which = "6.0.0"
```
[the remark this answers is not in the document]
tangowithfoxtrot: Fixed in 68973d0.

crates/bws/src/command/run.rs
```diff
@@ -0,0 +1,135 @@
...
+    command: Vec<String>,
+) -> Result<()> {
+    let shell = match std::env::consts::OS {
+        "windows" => shell.unwrap_or_else(|| "powershell".to_string()),
+        _ => shell.unwrap_or_else(|| "sh".to_string()),
+    };
+
+    if which(&shell).is_err() {
+        eprintln!("Error: shell '{}' not found", shell);
+        std::process::exit(1);
```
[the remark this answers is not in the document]
tangowithfoxtrot: Fixed in 3080dac6f133c70a154709adf8cd260b15b6721f.

crates/bws/src/command/run.rs
```diff
@@ -0,0 +1,135 @@
...
+        std::io::stdin().read_to_string(&mut buffer)?;
+        buffer
+    } else {
+        command.join(" ")
+    };
+
+    if user_command.is_empty() {
+        let mut cmd = Cli::command();
+        eprintln!("{}", cmd.render_help().ansi());
+        std::process::exit(1);
```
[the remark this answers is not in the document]
tangowithfoxtrot: Fixed in 3080dac6f133c70a154709adf8cd260b15b6721f.

crates/bws/src/command/run.rs
tangowithfoxtrot: Fixed in 23ab8f4b8551e53fb46a2e15b7a38e5a0dfdee95.

crates/bws/src/command/run.rs
tangowithfoxtrot: An attempt at making this portion more idiomatic was made in 3fcd32ce1d7ba50da19d4e751dd2921a013912c1, with a follow-up in eda00aa19bf8ec06498932224c976f855abae526 (in response to a clippy warning).

I moved things around a little bit so we could exit earlier if we detect duplicates. Let me know if that makes sense.

crates/bws/src/command/run.rs
tangowithfoxtrot: Fixed in f9629b8.

crates/bws/src/command/run.rs
```diff
@@ -0,0 +1,133 @@
...
+    no_inherit_env: bool,
+    shell: Option<String>,
+    command: Vec<String>,
+) -> Result<i32> {
+    let is_windows = std::env::consts::OS == "windows";
+
+    let shell = match is_windows {
+        true => shell.unwrap_or_else(|| "powershell".to_string()),
+        false => shell.unwrap_or_else(|| "sh".to_string()),
+    };
```
Hinton, line 37: I would generally write this as

```suggestion
    let shell = shell.unwrap_or_else(|| {
        if is_windows {
            "powershell".to_string()
        } else {
            "sh".to_string()
        }
    });
```

crates/bws/src/command/run.rs
```diff
@@ -0,0 +1,133 @@
...
+        .secrets()
+        .get_by_ids(SecretsGetRequest { ids: secret_ids })
+        .await?
+        .data;
+
+    if !uuids_as_keynames {
+        let mut seen = HashSet::new();
+        if let Some(s) = secrets.iter().find(|s| !seen.insert(&s.key)) {
+            bail!("Multiple secrets with name: '{}'. Use --uuids-as-keynames or use unique names for secrets", s.key);
+        }
```
Hinton, line 78: suggestion: We should either extract a helper function here or use itertools. Itertools is already a transient dependency for all our crates, and an explicit in some.

```suggestion
        if let Some(duplicate) = secrets.iter().map(|s| &s.key).duplicates().next() {
            bail!("Multiple secrets with name: '{}'. Use --uuids-as-keynames or use unique names for secrets", duplicate);
        }
```

crates/bws/src/main.rs
```diff
@@ -120,6 +121,30 @@ async fn process_commands() -> Result<()> {
...
+                no_inherit_env,
+                shell,
+                command,
+            )
+            .await
+            {
+                // exit with the exit code from the child process
+                Ok(exit_code) => std::process::exit(exit_code),
+                Err(e) => Err(e),
+            }
```
Hinton, line 145: You can rewrite this as:

```suggestion
            let exit_code = command::run::run(
                client,
                organization_id,
                project_id,
                uuids_as_keynames,
                no_inherit_env,
                shell,
                command,
            )
            .await?;

            // exit with the exit code from the child process
            std::process::exit(exit_code);
```

which removes the match and is easier to follow.

crates/bws/src/command/run.rs
tangowithfoxtrot: Applied in 75e3d8b918e5c7adc9f5e0d99975b3866dc5753d.

crates/bws/src/main.rs
tangowithfoxtrot: Thanks! Applied in e86aa2e06503900c31ade3e01649ad73c658ed82.

crates/bws/src/command/run.rs
tangowithfoxtrot: Applied in b86d0265ecfef374e2cb6042472f13b0688342f5 and added the dependency in f9456b8b0879564350b81e2d3c6147fc921a40d6.

crates/bws/src/util.rs
```diff
@@ -0,0 +1,55 @@
+use regex::Regex;
+use uuid::Uuid;
+
+const VALID_POSIX_NAME_REGEX: &str = "^[a-zA-Z_][a-zA-Z0-9_]*$";
+
+pub(crate) fn is_valid_posix_name(input_text: &str) -> bool {
+    match Regex::new(VALID_POSIX_NAME_REGEX) {
+        Ok(r) => r.is_match(input_text),
+        Err(_) => false,
+    }
```
Hinton, line 10: I wonder if we would want to panic here instead of handling invalid regex? Unless we modify the regex we know it will be a valid regex, and the below tests executes the code which should catch regressions.

Even with a regression silently marking all values as invalid posix might be confusing and using `expect` would provide an accurate error message.

```suggestion
    Regex::new(VALID_POSIX_NAME_REGEX)
        .expect("invalid VALID_POSIX_NAME_REGEX")
        .is_match(input_text)
```

tangowithfoxtrot: Updated in 5bdf99c.

Hinton (approved):
LGTM!

crates/bws/src/cli.rs
```diff
@@ -144,3 +166,12 @@ pub(crate) enum ProjectCommand {
     },
     List,
 }
+
+#[derive(Subcommand, Debug)]
+pub(crate) enum RunCommand {
```
mzieniukbw, line 171: I don't think we use this. Should we ?

tangowithfoxtrot: Good catch. I think I missed this when refectoring the command at some point. Removed in f9c9946.

mzieniukbw (approved):
LGTM

mzieniukbw (approved):

dani-garcia (approved):

mzieniukbw (approved):

dani-garcia (approved):

coltonhurst (approved):
Great work!

PR #626, opened 2024-02-22
add documentation and build script for Go wrapper

tangowithfoxtrot:
## Type of change

- [ ] Bug fix
- [ ] New feature development
- [x] Other

## Objective

Add a build script for building the libs needed for the Go wrapper.

## Code changes

- **languages/go/README.md:** add a reference to the build script.
- **languages/go/build.sh:** build the necessary dependencies for the Go wrapper. 

## Before you submit

### DIFF
--- a/languages/go/build.sh
+++ b/languages/go/build.sh
@@ -1,0 +1,44 @@
+#!/usr/bin/env bash
+
+check_command() {
+  if ! command -v "$1" &>/dev/null; then
+    printf '%s\n' "$1 is required to build locally. Please install $2."
+    exit 1
+  fi
+}
+
+check_command cargo Rust
+check_command go Go
+check_command npm Node.js
+
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+GO_LIB_DIR="$REPO_ROOT/languages/go/internal/cinterface/lib"
+
+pushd "$REPO_ROOT" || exit 1
+printf '%s\n\n' "Cleaning old builds..."
+rm -f languages/go/example/example && rm -rf "$GO_LIB_DIR"
+
+printf '%s\n\n' "Building binaries..."
+cargo build
+npm run schemas
+
+printf '%s\n\n' "Copying Go bindings to $GO_LIB_DIR..."
+mkdir -p "$GO_LIB_DIR"
+find target/debug -maxdepth 1 -type f -name "libbitwarden_c.*" -exec cp {} "$GO_LIB_DIR"/ \;
+
+printf '%s\n\n' "Build complete!"
+printf '%s\n' "To run the Go example, set the following environment variables:
+  export API_URL=\"http://localhost:4000\" # your Bitwarden API URL
+  export IDENTITY_URL=\"http://localhost:33656\" # your Bitwarden Identity URL
+  export ACCESS_TOKEN=\"your-access-token\" # your Bitwarden access token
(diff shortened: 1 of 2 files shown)


languages/go/build.sh
```diff
@@ -0,0 +1,44 @@
...
+
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+GO_LIB_DIR="$REPO_ROOT/languages/go/internal/cinterface/lib"
+
+pushd "$REPO_ROOT" || exit 1
+printf '%s\n\n' "Cleaning old builds..."
+rm -f languages/go/example/example && rm -rf "$GO_LIB_DIR"
+
+printf '%s\n\n' "Building binaries..."
+cargo build
```
dani-garcia, line 22: Should we expose a way of building in release mode?

tangowithfoxtrot: Fair enough. I can get that added pretty easily.

languages/go/build.sh
```diff
@@ -0,0 +1,52 @@
...
+  cargo build --release
+else
+  cargo build
+fi
+
+npm i && npm run schemas
+
+printf '%s\n\n' "Copying Go bindings to $GO_LIB_DIR..."
+mkdir -p "$GO_LIB_DIR"
+find target/debug -maxdepth 1 -type f -name "libbitwarden_c.*" -exec cp {} "$GO_LIB_DIR"/ \;
```
dani-garcia, line 35: There's a reference to the `target/debug` dir here, which should be `target/release` when building in release mode.

Also, the `libbitwarden_c.*` wildcard copies four files for me (`libbitwarden_c.{a,d,dylib,rlib}`), while only the `dylib` seems to be needed on Mac.

I don't think the extra copies are necessarily a concern, but only `*.dylib` and `*.a` are in the gitignore file, so at the very least we should add the two other types to not clutter the working directory (`*.d` and `*.rlib`).

languages/go/build.sh
```diff
@@ -0,0 +1,52 @@
...
+
+printf '%s\n\n' "Building binaries..."
+
+if [ "$1" = "--release" ]; then
+  cargo build --release
+else
+  cargo build
+fi
+
+npm i && npm run schemas
```
Hinton, line 31: We never want to run `npm i`, and `ci` would re-install every dependency. I suggest presuming the environment is. already bootstrapped like our other build scripts.

languages/go/build.sh
```diff
@@ -0,0 +1,52 @@
...
+check_command() {
+  if ! command -v "$1" &>/dev/null; then
+    printf '%s\n' "$1 is required to build locally. Please install $2."
+    exit 1
+  fi
+}
+
+check_command cargo Rust
+check_command go Go
+check_command npm Node.js
```
Hinton, line 14: We generally presume the environment is correctly bootstraped. These checks can provide a false confidence since we don't validate the version of the scripts.

mzieniukbw (commented):
I am a bit lost with the changes :)

languages/go/build.sh
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
coltonhurst: I would agree, @tangowithfoxtrot maybe we can discuss what the heart of the PR is? I have an idea that I think you'll like 😄

tangowithfoxtrot:
Closing this in favor of updating the documentation and possibly implementing it as an `npm` script.

PR #638, opened 2024-02-28
[PM-6262] Add basic feature flags support to enable cipher key encryption

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
Implement support for some basic feature flagging for the mobile clients, the only flag supported is `enableCipherKeyEncryption`


crates/bitwarden/src/client/flags.rs
```diff
@@ -0,0 +1,43 @@
...
+    pub enable_cipher_key_encryption: bool,
+}
+
+impl Flags {
+    pub fn load_from_map(map: std::collections::HashMap<String, bool>) -> Self {
+        let map = map
+            .into_iter()
+            .map(|(k, v)| (k, serde_json::Value::Bool(v)))
+            .collect();
+        serde_json::from_value(serde_json::Value::Object(map)).expect("Valid map")
```
Hinton, line 13: Serde 😢, we could write a proc macro for this but unsure if it's worth the hassle.

crates/bitwarden/src/mobile/vault/client_ciphers.rs
```diff
@@ -12,9 +12,16 @@ pub struct ClientCiphers<'a> {
 }
 
 impl<'a> ClientCiphers<'a> {
-    pub async fn encrypt(&self, cipher_view: CipherView) -> Result<Cipher> {
+    pub async fn encrypt(&self, mut cipher_view: CipherView) -> Result<Cipher> {
         let enc = self.client.get_encryption_settings()?;
 
+        if cipher_view.key.is_none() && self.client.get_flags().enable_cipher_key_encryption {
```
Hinton, line 18: Can we add a comment reminding us to add this logic to the regular encryption logic when we remove this flag?

crates/bitwarden/src/client/flags.rs
dani-garcia: We could even do a small declarative macro that just does:
```
this_flag = map.get("this_flag").unwrap_or(false);
that_flag = map.get("that_flag").unwrap_or(false);
another_flag = map.get("another_flag").unwrap_or(false);
```

That said I think we can deal with that when we have more flags to handle.

Hinton (approved):

PR #639, opened 2024-02-28
[PM-5635] Support rustls-platform-verifier on Android

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Support rustls-platform-verifier on Android, I still have some errors with some self signed certificates left to investigate, but it works already with valid certificates.


Hinton (commented):
Do we need to document what settings are required on the app side?

crates/bitwarden-core/src/android_support.rs
Hinton, line 1: Can we move these support files to bitwarden-uniffi?

dani-garcia:
> Do we need to document what settings are required on the app side?

The only setting needed is the network security config xml, and only if you want to use self signed certificates. The android app is using them already, so they shouldn't require any changes: https://github.com/bitwarden/android/blob/main/app/src/main/res/xml/network_security_config.xml

(They are not disabling self signed certs for bitwarden.eu, though)

I'll mention in in the readme for bitwarden-uniffi and in the Kotlin code example.

Hinton (approved):

PR #641, opened 2024-02-29
[PM-6100] Test for memory leaks of secrets

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
 Added a small test framework to test for secret leaks in memory.

This consists of a few parts:
- Binary crate `memory-testing`, this compiles to a binary that uses `bitwarden_crypto` to create some keys in memory and then frees them. The execution of this program goes like this:
    - Starts, keys get defined in memory
    - Waits for input (This is where we create an initial core dump)
    - Program frees the keys
    - Waits for input (This is where we create a final core dump)
    - Program exits normally
- A `capture_dumps.py` Python script, it's purpose is starting the program and orchestrating the core dumps and sending inputs to the program to continue.
- A `Dockerfile` that will compile the program and run the  `capture_dumps.py` script, this is needed because the dumps only work on a Linux environment.
- A `test.py` script that analyzes the memory dumps for secrets in memory
- A `run_tests.sh` script that builds and runs the docker container and the test script in one invocation

I've tried other tools to run it natively on other operating systems like osxpmem on mac and they either don't work on ARM Macs or they require running as root and disabling System Integrity Protection.

I've also added a small workflow to run these tests, as that runs on a linux environment, it's run directly without docker.

The results are printed to a table now: 
image


crates/memory-testing/src/main.rs
```diff
@@ -0,0 +1,41 @@
...
+
+    let mut test_string = String::new();
+    test_string.push_str("THIS IS USED TO CHECK THAT ");
+    test_string.push_str("THE MEMORY IS DUMPED CORRECTLY");
+
+    // In HEX:
+    // KEY: 15f8 5554 ff1f 9852 1963 55a6 46cc cf99 1995 0b15 cd59 5709 7df3 eb6e 4cb0 4cfb
+    // MAC: 4136 481f 8581 93f8 3f6c 5468 b361 7acf 7dfb a3db 2a32 5aa3 3017 d885 e5a3 1085
+    let symm_key = SymmetricCryptoKey::from_str(
+        "FfhVVP8fmFIZY1WmRszPmRmVCxXNWVcJffPrbkywTPtBNkgfhYGT+D9sVGizYXrPffuj2yoyWqMwF9iF5aMQhQ==",
```
Hinton, line 16: Do you think we should search the dump for the base64 encoded string to? To ensure we clean up `from_str`. I think that means moving away from FromStr to From<String> to ensure we consume the strings. We also have to use some direct memory calls to replace the existing string memory. Probably better left for the future.

Hinton, line 16: Thoughts on proving the key as input from stdin? In the future I envision we would expand the tool with more scenarios and it might be beneficial to not burn in all variants of enc strings and crypto keys.

.github/workflows/memory-testing.yml
Hinton, line 1: Should we limit this workflow to only run when you modify the crypto files and/or memory testing? I think we generally have to many workflows running all the time.

Hinton (commented):
Looks like a great start. I'm somewhat pondering if we could write the python logic in rust and avoid having multiple programming languages.

Hinton:
We should also exclude the memory tool from codecov.

crates/memory-testing/src/main.rs
dani-garcia: Yeah sounds reasonable, we can add the test for it now, and we can just ignore the failing result until we have dealt with it later

dani-garcia: Yeah reading from stdin or a json file sounds interesting, though I'd be worried about the std library or serde making unnecessary copies and affecting the results. I'll give it a try and see what happens.

.github/workflows/memory-testing.yml
dani-garcia: Yeah that makes sense, done

.github/workflows/memory-testing.yml
```diff
@@ -0,0 +1,54 @@
...
+jobs:
+  msrv:
+    name: Memory Testing for - ${{ matrix.settings.os }} - ${{ matrix.settings.target }}
+    runs-on: ${{ matrix.settings.os || 'ubuntu-latest' }}
+    strategy:
+      fail-fast: false
+      matrix:
+        settings:
+          - os: ubuntu-22.04
+            target: x86_64-unknown-linux-gnu
```
Hinton, line 31: Do you think we'll expand this to other architectures? If not lets remove the matrix and just run it on ubuntu.

crates/memory-testing/src/main.rs
Hinton: Should we add a comment so we don't get confused why it's disabled and being triggered.

.github/workflows/memory-testing.yml
dani-garcia: At the moment this is linux only, but we could also run it on arm linux. That said none of the code we're handling deals with anything os or architecture specific so there might be no reason to.

.gitignore
```diff
@@ -12,7 +12,6 @@ x64/
 x86/
 build/
 bld/
-[Bb]in/
```
dani-garcia, line 15: This was conflicting with the rust crates bin directory, I've removed it because I don't think it's applicable anywhere else.

crates/memory-testing/src/main.rs
dani-garcia: Added the comment in the analyze program

dani-garcia: I pushed some new changes that migrate the scripts to rust binaries and loads the test cases through JSON, seems to be fine as long as you zeroize the string after parsing it to JSON.

Hinton (approved):

crates/memory-testing/run_test.sh
```diff
@@ -0,0 +1,20 @@
+# Move to the root of the repository
+cd "$(dirname "$0")"
+cd ../../
+
+BASE_DIR="./crates/memory-testing"
+
+mkdir -p $BASE_DIR/output
+rm $BASE_DIR/output/*
+
+cargo build -p memory-testing
```
Hinton, line 10: I think we only need this in the no-docker condition?

dani-garcia: At the moment this script is running only the `capture-dumps` binary in Docker, as that's the only one that requires special permissions for the memory dump. The `analyze-dumps` is run on the host system.

I think we could move the analyze binary to run inside docker too to save one compilation, but assuming this is being run inside a developer's repo, they probably already have the binaries built anyway.

Hinton: Got it, never mind in that case.

.github/workflows/memory-testing.yml
```diff
@@ -0,0 +1,47 @@
+---
+name: Memory Testing
```
withinfocus, line 2:
```suggestion
name: Test for memory leaks
```

.github/workflows/memory-testing.yml
```diff
@@ -0,0 +1,47 @@
...
+
+on:
+  pull_request:
+    paths:
+      - "crates/bitwarden-crypto/**"
+      - "crates/memory-testing/**"
+  push:
```
withinfocus, line 9: ❓ Is this needed if the intention is to catch these during code review?

.github/workflows/memory-testing.yml
```diff
@@ -0,0 +1,47 @@
...
+      - "crates/bitwarden-crypto/**"
+      - "crates/memory-testing/**"
+    branches:
+      - "main"
+      - "rc"
+      - "hotfix-rc"
+
+defaults:
+  run:
+    shell: bash
```
withinfocus, line 20: ❓ Is this necessary?

.github/workflows/memory-testing.yml
```diff
@@ -0,0 +1,47 @@
...
+
+jobs:
+  memory-test:
+    name: Memory Testing
```
withinfocus, line 24:
```suggestion
    name: Testing
```

.github/workflows/memory-testing.yml
```diff
@@ -0,0 +1,47 @@
...
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: Checkout
```
withinfocus, line 28:
```suggestion
      - name: Check out repo
```

.github/workflows/memory-testing.yml
```diff
@@ -0,0 +1,47 @@
...
+        uses: dtolnay/rust-toolchain@be73d7920c329f220ce78e0234b8f96b7ae60248 # stable
+        with:
+          toolchain: stable
+
+      - name: Cache cargo registry
+        uses: Swatinem/rust-cache@23bce251a8cd2ffc3c1075eaa2367cf899916d84 # v2.7.3
+        with:
+          key: memtest-cargo
+
+      - name: Memory Testing
```
withinfocus, line 46:
```suggestion
      - name: Test
```

crates/memory-testing/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+WORKDIR /app
+RUN cargo build -p memory-testing
+
+###############################################
+#                  App stage                  #
+###############################################
+FROM debian:bookworm-slim
+
+# This specifically needs to run as root to be able to capture core dumps
+USER root
```
withinfocus, line 19: ❌ There are a few Checkmarx findings that can be addressed for this testing container I'd think.

withinfocus (commented):
Few things based on how we are writing other Actions these days. Also room to fix a few others.

Eeebru (dismissed):

.github/workflows/memory-testing.yml
dani-garcia: We're running a shell script so I just set it by default, but the script isn't doing anything that specifically requires bash, I've removed it and it seems to work.

.github/workflows/memory-testing.yml
dani-garcia: Not really, I based myself in the `rust-test.yml` which does run on push too, so I made it the same for consistency. But we could remove it and make it run only on PR

withinfocus: Could catch non-PR scenarios too. I don't know why that'd happen but doesn't hurt.

crates/memory-testing/Dockerfile
dani-garcia: Good point, fixed a few of them, now there's only two left:
- Root user: This is required to take the memory dump
- Docker healthcheck: The container only runs for a couple of seconds, so I don't think this would be very useful in this situation

Should I close these two findings on the Checkmarx website?

withinfocus: Fair points. Yes, it would be good to propose those as not exploitable with comments and then we'll be OK. Thanks!

Hinton (approved):

Eeebru (approved):

PR #645, opened 2024-03-01
[SM-1130] - add install scripts

tangowithfoxtrot:
## Type of change

- [ ] Bug fix
- [ ] New feature development
- [x] Other

## Objective

Allow easy installation of `bws` on most systems with:

`curl <script-url> | sh` on Linux and macOS
or
`iwr <script-url> | iex` on Windows

## Code changes

- **crates/bws/scripts/install.(sh|ps1):** added install scripts that will download `bws` from GitHub Releases, validate the checksums, and install it. I am unsure if this is the _best_ place to house these scripts and would welcome suggestions for other places these might belong.
- **.github/workflows/version-bump.yml:** added version bumps for the install scripts

## Before you submit


coltonhurst:
Not a bad location, but I agree we should think about where we want this. (If we do keep it in the `bws` crate I think it should definitely be in some type of subdirectory rather than the crate root.)

Some location ideas:
- Separate directory, like `crates/bws/scripts/<here>`
- GitHub Gist
- Somewhere else?

.github/workflows/version-bump.yml
```diff
@@ -103,16 +103,19 @@ jobs:
...
       ### bws
-
       - name: Bump bws Version
         if: ${{ inputs.project == 'bws' }}
-        run: cargo set-version -p bws ${{ inputs.version_number }}
+        run: |
+          cargo set-version -p bws ${{ inputs.version_number }}
+          # bump the version in install.sh
+          sed -i 's/BWS_VERSION="${BWS_VERSION:-[0-9]\+\.[0-9]\+\.[0-9]\+}"/BWS_VERSION="${BWS_VERSION:-${{ inputs.version_number }}}"/' ./crates/bws/install.sh
+          # bump the version in install.ps1
```
tangowithfoxtrot, line 117: These sed expressions work locally, but there doesn't appear to be a way to test this in the workflow, so I would appreciate some scrutiny here.

tangowithfoxtrot:
@coltonhurst, yeah, I'm beginning to think that the first option you provided is probably the way to go to avoid cluttering things too much. I don't think a GitHub gist will do though because we'll need to be able to automatically bump the version numbers in the install scripts.

I'll wait to see if any other ideas come up before moving them there, just to give other people the chance to chime-in.

.github/workflows/version-bump.yml
```diff
@@ -103,16 +103,19 @@ jobs:
...
+          sed -i 's/{ \$env:bwsVersion } else { "[0-9]\+\.[0-9]\+\.[0-9]\+" }/\{ \$env:bwsVersion } else { "${{ inputs.version_number }}" }/' ./crates/bws/install.ps1
```
tangowithfoxtrot, line 118: These sed expressions worked when I tested them locally, but there doesn't appear to be a way to test this in the workflow, so I would appreciate some scrutiny here.

crates/bws/install.sh
```diff
@@ -0,0 +1,139 @@
...
+  install_bws
+}
+
+error() {
+  echo "$1" >&2
+  echo "Exiting..." >&2
+  exit 1
+}
+
+check_required() {
```
coltonhurst, line 25: I'm thinking let's potentially check to make sure some of the other programs required are present... Like `sha256sum`, etc.

crates/bws/install.sh
```diff
@@ -0,0 +1,139 @@
...
+}
+
+validate_checksum() {
+  checksum_url="https://github.com/bitwarden/sdk/releases/download/bws-v${BWS_VERSION}/bws-sha256-checksums-${BWS_VERSION}.txt"
+  echo "Downloading checksum file from: $checksum_url"
+  checksum_file="$tmp_dir/bws-checksums.txt"
+  downloader "$checksum_url" "$checksum_file"
+
+  expected_checksum="$(grep "bws-${ARCH}-${PLATFORM}-${BWS_VERSION}.zip" "$checksum_file" | awk '{print $1}')"
+  actual_checksum="$(sha256sum "$tmp_dir/bws.zip" | awk '{print $1}')"
```
coltonhurst, line 100: Doesn't this get the expected and actual checksum from the same downloaded zip?

coltonhurst (changes requested):
Some brief thoughts, I can do a deeper dive once it's fully ready again?

We might want to consider adding a way to uninstall as well, and make sure nothing is left on the user's system.

crates/bws/install.sh
coltonhurst: I also did not recognize some commands in the script, they also do not seem to be on my system, at least by default. Examples are `error`, `extract`, etc unless I am missing something 🙂 (Which is very possible, I am not an expert in bash scripts!)

tangowithfoxtrot: Ah... It looks like `sha256sum` is a part of `coreutils` on macOS, which has to be installed. I was trying to use only pre-installed executables where possible. I'll see if macOS has a preinstalled way to do this and go with that if I can.

tangowithfoxtrot: `error` and `extract` are defined in [here](https://github.com/bitwarden/sdk/pull/645/files#diff-24589d5888f18206c9e01fa697e2ec59167c759cbced32664241bc456166f373R19-R23) and [here](https://github.com/bitwarden/sdk/pull/645/files#diff-24589d5888f18206c9e01fa697e2ec59167c759cbced32664241bc456166f373R82-R84).

crates/bws/install.sh
tangowithfoxtrot: This greps for the string `"bws-${ARCH}-${PLATFORM}-${BWS_VERSION}.zip"` in the [checksum file](https://github.com/bitwarden/sdk/pull/645/files#diff-24589d5888f18206c9e01fa697e2ec59167c759cbced32664241bc456166f373R96-R97) and uses awk to return _just_ the checksum value. 

Like this:
```bash
bash-5.2$ grep bws-aarch64-apple-darwin-0.5.0.zip ~/Downloads/bws-sha256-checksums-0.5.0.txt
aeac01edbd7cdfb8e75e9143d13d69221e6e0bff0fd9f8de69b85c3108a4e986 bws-aarch64-apple-darwin-0.5.0.zip
bash-5.2$ grep bws-aarch64-apple-darwin-0.5.0.zip ~/Downloads/bws-sha256-checksums-0.5.0.txt | awk '{print $1}'
aeac01edbd7cdfb8e75e9143d13d69221e6e0bff0fd9f8de69b85c3108a4e986
```

crates/bws/install.sh
tangowithfoxtrot: It looks like `shasum -a 256` is the way to go for macOS. `shasum` is included in default macOS installations, so I just need to add some special handling there for macOS.

tangowithfoxtrot: I added a fallback for checksum validation on macOS installations that haven't installed `coreutils` [here](https://github.com/bitwarden/sdk/pull/645/files#diff-24589d5888f18206c9e01fa697e2ec59167c759cbced32664241bc456166f373R74-R80).

coltonhurst: Awesome, thanks!

mzieniukbw (changes requested):
Nice scripts!
I have concern with the checksums on powershell though - did you had a chance to test it out on Windows? :)

.github/workflows/version-bump.yml
```diff
@@ -103,16 +103,19 @@ jobs:
...
-
       - name: Bump bws Version
         if: ${{ inputs.project == 'bws' }}
-        run: cargo set-version -p bws ${{ inputs.version_number }}
+        run: |
+          cargo set-version -p bws ${{ inputs.version_number }}
+          # bump the version in install.sh
+          sed -i 's/BWS_VERSION="${BWS_VERSION:-[0-9]\+\.[0-9]\+\.[0-9]\+}"/BWS_VERSION="${BWS_VERSION:-${{ inputs.version_number }}}"/' ./crates/bws/scripts/install.sh
+          # bump the version in install.ps1
+          sed -i 's/{ \$env:bwsVersion } else { "[0-9]\+\.[0-9]\+\.[0-9]\+" }/\{ \$env:bwsVersion } else { "${{ inputs.version_number }}" }/' ./crates/bws/scripts/install.ps1
```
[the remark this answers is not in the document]
tangowithfoxtrot: Great idea!

crates/bws/scripts/install.ps1
```diff
@@ -0,0 +1,80 @@
...
+  Write-Host "Validating checksum..."
+
+  $checksumUrl = "https://github.com/bitwarden/sdk/releases/download/bws-v$bwsVersion/bws-sha256-checksums-$bwsVersion.txt"
+  $checksumFile = Join-Path $env:TEMP "bws-checksums.txt"
+  Invoke-WebRequest -Uri $checksumUrl -OutFile $checksumFile
+
+  $expectedChecksum = (Get-Content $checksumFile | Where-Object { $_ -match "bws-$arch-pc-windows-msvc-$bwsVersion.zip" }).Split(" ")[0]
+  $actualChecksum = (Get-FileHash -Algorithm SHA256 -Path $zipPath).Hash
+
+  if ($actualChecksum -ne $expectedChecksum) {
```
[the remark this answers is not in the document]
tangowithfoxtrot: From what I can tell, PowerShell is case-insensitive, where possible. I found [this](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_comparison_operators?view=powershell-7.4#common-features), which says that this is also the case for comparison operators, so the comparison should return `true`, despite being in different cases.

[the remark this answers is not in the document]
tangowithfoxtrot: Regarding testing on Windows: I have tested it sparingly in a VM a few times throughout updating the script using `iwr <url-for-raw-file> | iex`, and it seems to be working.

I haven't been able to test it much on x86_64 though, as the only Windows machine in my house is a VM on my M2 Mac.

.github/workflows/version-bump.yml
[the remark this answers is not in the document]
tangowithfoxtrot: Fixed in https://github.com/bitwarden/sdk/pull/645/commits/2dba059c1ee17366debb9535cff34fda767c3ff9

.github/workflows/version-bump.yml
tangowithfoxtrot: Updated with simpler regex matches in https://github.com/bitwarden/sdk/pull/645/commits/2dba059c1ee17366debb9535cff34fda767c3ff9.

tangowithfoxtrot:
@coltonhurst, I added an uninstallation arg in https://github.com/bitwarden/sdk/pull/645/commits/24e0e98281f30200fec93980ce4b40d803487975. 

It's pretty easy to invoke in Linux/macOS with:

```sh
curl https://raw.githubusercontent.com/bitwarden/sdk/2dba059c1ee17366debb9535cff34fda767c3ff9/crates/bws/scripts/install.sh | sh -s -- --uninstall
```

However, I can't figure out a straightforward way to pass the `-Uninstall` arg through `Invoke-Expression` in PowerShell, so the best way to perform an uninstallation on Windows will require downloading the install script locally and passing the arg directly, like this:

```powershell
.\install.ps1 -Uninstall
```

If you have greater PowerShell finesse than me, you can probably pass the `-Uninstall` arg through the 

```powershell
iex https://raw.githubusercontent.com/bitwarden/sdk/2dba059c1ee17366debb9535cff34fda767c3ff9/crates/bws/scripts/install.ps1 | iex # not sure how to pass args to the script through this???
```

one-liner, but I haven't been successful in doing so.

crates/bws/scripts/install.ps1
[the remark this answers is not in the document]
mzieniukbw: Windows as always have to be unique. Indeed it seems that string comparison is case insensitive, which is mind blowing to me 🤯 
Thank you for the insights into testing, i wondered how do you test out Windows on Mac. 👍

mzieniukbw (approved):
Approved, but would like to know other's feedback @coltonhurst on whether uninstalling should also remove `.bws` config directory - i am personally fine with it.

tangowithfoxtrot:
@coltonhurst, Added mention of the install scripts in the README in https://github.com/bitwarden/sdk/pull/645/commits/035ca9340bfaa706370e083e95dd79f9cc364b39

mzieniukbw (approved):

michalchecinski (approved):

PR #647, opened 2024-03-01
[SM-1223] restructure Go wrapper for distribution

tangowithfoxtrot:
## Type of change

- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Restructure the Go wrapper so that it works when consumed as a package. The current structure uses "internal" Go modules, which will throw errors when you try to use them:
```
/go/pkg/mod/github.com/bitwarden/sm-sdk-go@v0.1.0/bitwarden_client.go:6:2: use of internal package github.com/bitwarden/sdk/languages/go/internal/cinterface not allowed
```

## Code changes

## Before you submit


languages/go/bitwarden_client.go
```diff
@@ -3,7 +3,7 @@ package sdk
 import (
 	"encoding/json"
 
-	"github.com/bitwarden/sdk/languages/go/internal/cinterface"
+	"github.com/bitwarden/sm-sdk-go/cinterface"
```
Hinton, line 6: The internal package was intentional, to limit the external APIs and reduce the surface for major versions. The `cinterface` should not be exposed externally. The reason this doesn't work is because you are trying to import it from the sm-sdk-go repository and it's loading two different packages.

Hinton (changes requested):

Hinton:
We also don't want to distribute the lib files in a repository unless we force puch every release. Since with the binaries occupying 30mb+ we can expect the repository size to quickly become unreasonable.

tangowithfoxtrot:
> We also don't want to distribute the lib files in a repository unless we force puch every release. Since with the binaries occupying 30mb+ we can expect the repository size to quickly become unreasonable.

That makes sense, but what's the alternative? Would we require people that want to import our Go wrapper to build the libs from source?

sylwit:
I'm working on adding Bitwarden secrets manager as a provider to external secret operator for k8s and not being able to use this sdk as a package is a real deal breaker.

Do you have any recommendations on how this should be used in a multi architecture environment?

And the go example doesn't work, the Command struct is not imported

Thank you

PR #650, opened 2024-03-04
Exclude unecessary files from published bws

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Prevent unecessary files from being included in the crates distributed build.

## Before you submit

### DIFF
--- a/crates/bws/Cargo.toml
+++ b/crates/bws/Cargo.toml
@@ -11,6 +11,7 @@ description = """
 Bitwarden Secrets Manager CLI
 """
 keywords = ["bitwarden", "secrets-manager", "cli"]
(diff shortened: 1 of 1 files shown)


dani-garcia (commented):
As far as I can see those files are already not included in the package:

image

We can probably exclude Cross.toml though, and maybe the changelog too

Hinton:
@dani-garcia 0.4.0 was released a while ago when we had cross.toml in the repository. https://github.com/bitwarden/sdk/tree/bws-v0.4.0/crates/bws

dani-garcia (approved):
Somehow I though the last release was pretty recent 🤦

PR #652, opened 2024-03-07
Explore using sqlite

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Code changes

## Before you submit


crates/bitwarden/src/vault/cipher/repository.rs
```diff
@@ -0,0 +1,222 @@
...
+        // TODO: Move into a separate util
+        guard
+            .execute(
+                "CREATE TABLE IF NOT EXISTS ciphers (
+                id TEXT PRIMARY KEY,
+                value TEXT NOT NULL
+         )",
+                (),
+            )
+            .unwrap();
```
Hinton, line 48: I'm thinking of moving this into a migration service instead.

crates/bitwarden/src/vault/client_vault.rs
```diff
@@ -1,8 +1,12 @@
...
-use crate::{error::Result, Client};
+use super::{
+    repository::CipherSqliteRepository,
+    sync::{sync, SyncRequest, SyncResponse},
+};
+use crate::{error::Result, vault::cipher::repository::CipherRepository, Client};
 
 pub struct ClientVault<'a> {
     pub(crate) client: &'a mut crate::Client,
+    pub cipher_repository: Box<dyn CipherRepository + Send>,
```
Hinton, line 9: I think repositories should be internal to the module and we probably want to expose some service instead. Since we will want to hook into some event system, and probably expose nice methods for saving decrypted views.

PR #653, opened 2024-03-11
fix: linux arm64 cross-compilation (hopefully)

tangowithfoxtrot:
## Type of change

- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Code changes

## Before you submit

### DIFF
--- a/.github/workflows/build-rust-cross-platform.yml
+++ b/.github/workflows/build-rust-cross-platform.yml
@@ -57,7 +57,9 @@ jobs:
         env:
           RUSTFLAGS: "-D warnings"
         if: ${{ matrix.settings.target == 'aarch64-unknown-linux-gnu' }}
-        run: cross build --target ${{ matrix.settings.target }} --release
+        run: |
+          cargo clean
+          cross build --target ${{ matrix.settings.target }} --release
 
       - name: Upload Artifact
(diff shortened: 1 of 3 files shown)


.github/workflows/build-rust-cross-platform.yml
```diff
@@ -44,8 +46,18 @@ jobs:
...
 
+      - name: Build Rust
+        env:
+          RUSTFLAGS: "-D warnings"
+        if: ${{ matrix.settings.target }} == aarch64-unknown-linux-gnu
+        run: |
+          sudo apt update
+          sudo apt install -y crossbuild-essential-arm64
+          cargo build --target ${{ matrix.settings.target }} --release
+
```
Hinton, line 60: Use cross for this, that's what we do in the other workflows.

tangowithfoxtrot: Actually, [looking at this](https://github.com/bitwarden/sdk/actions/runs/8281072317/job/22658818729#step:7:110), cross apparently doesn't have the right build tools installed? I thought `cross` pulled a container image that was configured for whatever target you specified.

tangowithfoxtrot: [This](https://github.com/bitwarden/sdk/pull/653/commits/be03087f123511668469417f6c104b77e2058918#) isn't running with cross? I referenced the Build CLI job to base that off of.

PR #657, opened 2024-03-12
[PM-6761] Add fido credentials to data model

Hinton:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Add `Fido2Credentials` to the `Login` struct to support storing passkeys. Currently we don't decrypt passkeys but rather pass them along encrypted to ensure we can re-encrypt the login without data loss.

## Before you submit

### DIFF
--- a/crates/bitwarden/src/vault/cipher/login.rs
+++ b/crates/bitwarden/src/vault/cipher/login.rs
@@ -129,7 +129,7 @@ impl KeyDecryptable<SymmetricCryptoKey, LoginView> for Login {
             uris: self.uris.decrypt_with_key(key).ok().flatten(),
             totp: self.totp.decrypt_with_key(key).ok().flatten(),
             autofill_on_page_load: self.autofill_on_page_load,
-            fido2_credentials: self.fido2_credentials.as_ref().map(|v| v.to_vec()),
+            fido2_credentials: self.fido2_credentials.clone(),
         })
     }
 }
--- a/crates/bitwarden/src/vault/cipher/login.rs
+++ b/crates/bitwarden/src/vault/cipher/login.rs
@@ -38,6 +38,25 @@ pub struct LoginUriView {
     pub r#match: Option<UriMatchType>,
 }
 
+#[derive(Serialize, Deserialize, Debug, JsonSchema, Clone)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+#[cfg_attr(feature = "mobile", derive(uniffi::Record))]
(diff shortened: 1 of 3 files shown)


crates/bitwarden/src/vault/cipher/login.rs
```diff
@@ -62,6 +83,9 @@ pub struct LoginView {
     pub uris: Option<Vec<LoginUriView>>,
     pub totp: Option<String>,
     pub autofill_on_page_load: Option<bool>,
+
+    // TODO: Remove this once the SDK supports state
+    pub fido2_credentials: Option<Vec<Fido2Credential>>,
```
coroiu, line 88: ❓ question: I'm curious why? Would it move somewhere else?

coroiu (commented):
Looks good from my perspective, but I'll let Dani make the final approval

crates/bitwarden/src/vault/cipher/login.rs
```diff
@@ -38,6 +38,25 @@ pub struct LoginUriView {
...
+    pub key_type: EncString,
+    pub key_algorithm: EncString,
+    pub key_curve: EncString,
+    pub key_value: EncString,
+    pub rp_id: EncString,
+    pub user_handle: Option<EncString>,
+    pub user_name: Option<EncString>,
+    pub counter: EncString,
+    pub rp_name: Option<EncString>,
+    pub user_display_name: Option<EncString>,
```
dani-garcia, line 55: I feel like we're going to need `user_name`, `rp_name`, or `user_display_name` decrypted eventually to display them in the UI, but if it's not a need now we can leave that for the future.

crates/bitwarden/src/vault/cipher/login.rs
```diff
@@ -104,6 +129,7 @@ impl KeyDecryptable<SymmetricCryptoKey, LoginView> for Login {
             uris: self.uris.decrypt_with_key(key).ok().flatten(),
             totp: self.totp.decrypt_with_key(key).ok().flatten(),
             autofill_on_page_load: self.autofill_on_page_load,
+            fido2_credentials: self.fido2_credentials.as_ref().map(|v| v.to_vec()),
```
dani-garcia, line 132: Nit: This can be simplified by just using clone():
```suggestion
            fido2_credentials: self.fido2_credentials.clone(),
```

dani-garcia (approved):

dani-garcia (approved):

PR #660, opened 2024-03-14
Mark exporters and generators as internal dependencies

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

We accidentally exposed exporters and generators without feature flagging them. This PR properly puts them under the internal feature flag.

## Before you submit

### DIFF
--- a/crates/bitwarden/Cargo.toml
+++ b/crates/bitwarden/Cargo.toml
@@ -16,7 +16,10 @@ rust-version = "1.71"
 default = ["secrets"]
 
 secrets = [] # Secrets manager API
-internal = [] # Internal testing methods
+internal = [
+    "bitwarden-exporters",
+    "bitwarden-generators",
+] # Internal testing methods
 mobile = [
     "uniffi",
     "internal",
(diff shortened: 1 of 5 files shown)


crates/bitwarden/Cargo.toml
```diff
@@ -16,7 +16,10 @@ rust-version = "1.71"
 default = ["secrets"]
 
 secrets = [] # Secrets manager API
-internal = [] # Internal testing methods
+internal = [
+    "bitwarden-exporters",
+    "bitwarden-generators",
```
dani-garcia, line 21:
```suggestion
    "dep:bitwarden-exporters",
    "dep:bitwarden-generators",
```

If we don't use the `dep:` prefix, cargo tries to be helpful and expose a feature with the crate name, so you can do `cargo b -p bitwarden --features bitwarden-generators`

Using `dep:` removes the implicit feature

dani-garcia (approved):

PR #662, opened 2024-03-15
[PM-6760] Add support for URI checksums

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
Validate the URI checksums of ciphers. This is done in a similar way to the `enable_cipher_key_encryption` flag check


crates/bitwarden/src/mobile/vault/client_ciphers.rs
```diff
@@ -24,6 +24,11 @@ impl<'a> ClientCiphers<'a> {
             cipher_view.generate_cipher_key(key)?;
         }
 
+        // For compatibility reasons, we only create checksums for ciphers that have a key
+        if cipher_view.key.is_some() {
+            cipher_view.generate_checksums();
+        }
```
Hinton, line 30: I think this should move into the cipher logic. This is not temporary behaviour, and I believe we could easily do this as part of the encyrpt operation?

dani-garcia: Yeah, my thought was to group it with the cipher key behaviour and eventually when the flag was removed move both into the encrypt operation, but we can do it now with the checksums. I've updated the PR

crates/bitwarden/src/vault/cipher/cipher.rs
```diff
@@ -139,10 +139,15 @@ pub struct CipherListView {
 }
 
 impl KeyEncryptable<SymmetricCryptoKey, Cipher> for CipherView {
-    fn encrypt_with_key(self, key: &SymmetricCryptoKey) -> Result<Cipher, CryptoError> {
+    fn encrypt_with_key(mut self, key: &SymmetricCryptoKey) -> Result<Cipher, CryptoError> {
```
Hinton, line 142: Do we need to actually mutate self? We should be able to mutate just the URIs in login.

dani-garcia: We could change the mutability to only affect the login like this, but to me it seems harder to read:

This code will be nicer when moved to LoginUriView, but we can't do that until we decide to unconditionally generate the checksums, as the LoginUriView doesn't have knowledge of whether a cipher key is used or not.

```rust
    fn encrypt_with_key(self, key: &SymmetricCryptoKey) -> Result<Cipher, CryptoError> {
        let ciphers_key = Cipher::get_cipher_key(key, &self.key)?;
        let key = ciphers_key.as_ref().unwrap_or(key);

        let mut login = self.login;

        // For compatibility reasons, we only create checksums for ciphers that have a key
        if ciphers_key.is_some() {
            login.generate_checksums();
        }

        Ok(Cipher {
            id: self.id,
            organization_id: self.organization_id,
            ...
            login: login.encrypt_with_key(key)?,
```

PR #663, opened 2024-03-15
Use same version for bitwarden crates

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Rather than maintaining separate versions for every crate let's keep them locked. That means the rust sdk packages will always have the same version. Unfortunately a side effect of this is that every package will be released even if that package hasn't changed.

## Before you submit


vgrassia (dismissed):

dani-garcia (approved):
LGTM

michalchecinski (approved):

PR #672, opened 2024-03-20
Add sensitive types to SymmetricCryptoKey

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
This is a small subset of #536, which only contains the minimum code to protect the import/export functions on SymmetricCryptoKey. It also enables the from_base64 test on SymmetricCryptoKey as that passes now.

After this PR is merged I'll be expanding the use of Sensitive to other parts of the codebase while adding some extra memory testing checks to validate that it works for them.

### DIFF
--- a/crates/bitwarden-crypto/src/keys/device_key.rs
+++ b/crates/bitwarden-crypto/src/keys/device_key.rs
@@ -1,8 +1,6 @@
-use std::str::FromStr;
-
 use crate::{
-    error::Result, AsymmetricCryptoKey, AsymmetricEncString, CryptoError, EncString,
-    KeyDecryptable, KeyEncryptable, SymmetricCryptoKey,
+    error::Result, AsymmetricCryptoKey, AsymmetricEncString, CryptoError, DecryptedVec, EncString,
+    KeyDecryptable, KeyEncryptable, SensitiveString, SymmetricCryptoKey,
 };
 
 /// Device Key
@@ -40,7 +38,7 @@ impl DeviceKey {
         let data = user_key.to_vec();
 
         let protected_user_key =
-            AsymmetricEncString::encrypt_rsa2048_oaep_sha1(&data, &device_private_key)?;
+            AsymmetricEncString::encrypt_rsa2048_oaep_sha1(data.expose(), &device_private_key)?;
 
         let protected_device_public_key = device_private_key
             .to_public_der()?
@@ -51,7 +49,7 @@ impl DeviceKey {
             .encrypt_with_key(&device_key.0)?;
 
         Ok(TrustDeviceResponse {
-            device_key: device_key.to_base64(),
+            device_key: device_key.to_base64().expose().to_owned(),
             protected_user_key,
             protected_device_private_key,
             protected_device_public_key,
@@ -65,25 +63,25 @@ impl DeviceKey {
         protected_user_key: AsymmetricEncString,
     ) -> Result<SymmetricCryptoKey> {
         let device_private_key: Vec<u8> = protected_device_private_key.decrypt_with_key(&self.0)?;
-        let device_private_key = AsymmetricCryptoKey::from_der(device_private_key.as_slice())?;
+        let device_private_key = AsymmetricCryptoKey::from_der(&device_private_key)?;
 
-        let mut dec: Vec<u8> = protected_user_key.decrypt_with_key(&device_private_key)?;
-        let user_key: SymmetricCryptoKey = dec.as_mut_slice().try_into()?;
+        let dec: Vec<u8> = protected_user_key.decrypt_with_key(&device_private_key)?;
+        let dec = DecryptedVec::new(Box::new(dec));
+        let user_key = SymmetricCryptoKey::try_from(dec)?;
 
         Ok(user_key)
     }
 
-    fn to_base64(&self) -> String {
+    fn to_base64(&self) -> SensitiveString {
         self.0.to_base64()
     }
 }
 
-impl FromStr for DeviceKey {
-    type Err = CryptoError;
+impl TryFrom<SensitiveString> for DeviceKey {
+    type Error = CryptoError;
 
-    fn from_str(s: &str) -> Result<Self, Self::Err> {
-        let key = s.parse::<SymmetricCryptoKey>()?;
-        Ok(DeviceKey(key))
+    fn try_from(value: SensitiveString) -> Result<Self, Self::Error> {
+        SymmetricCryptoKey::try_from(value).map(DeviceKey)
     }
 }
 
@@ -98,10 +96,9 @@ mod tests {
 
         let result = DeviceKey::trust_device(&key).unwrap();
 
-        let decrypted = result
-            .device_key
-            .parse::<DeviceKey>()
-            .unwrap()
+        let device_key = SensitiveString::new(Box::new(result.device_key));
+        let device_key = DeviceKey::try_from(device_key).unwrap();
+        let decrypted = device_key
             .decrypt_user_key(
                 result.protected_device_private_key,
                 result.protected_user_key,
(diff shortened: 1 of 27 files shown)


crates/bitwarden/src/mobile/crypto.rs
```diff
@@ -172,7 +174,7 @@ pub async fn get_user_encryption_key(client: &mut Client) -> Result<String> {
         .get_key(&None)
         .ok_or(Error::VaultLocked)?;
 
-    Ok(user_key.to_base64())
+    Ok(user_key.to_base64().expose().to_owned())
```
Hinton, line 177: Should we add support to uniffi for Sensitive and avoid exposing them explicitly?

dani-garcia: Yeah that sounds reasonable, added it in a5a8097

dani-garcia: @Hinton I've had to add a uniffi Record using the new type to make sure it gets generated by uniffi correctly, which is a bit weird, have we hit this limitation before? 2c1ff1cc1e6f0f96bc626a2150dde73a2ed8010f

crates/bitwarden-crypto/src/sensitive/sensitive.rs
```diff
@@ -0,0 +1,211 @@
...
+        let rtn = String::from_utf8(*value).map_err(|_| CryptoError::InvalidUtf8String);
+        rtn.map(|v| Sensitive::new(Box::new(v)))
+    }
+}
+
+impl SensitiveString {
+    pub fn decode_base64<T: base64::Engine>(self, engine: T) -> Result<SensitiveVec, CryptoError> {
+        // Preallocate a Vec with the necessary capacity
+        let len = base64::decoded_len_estimate(self.value.len());
+        let mut value = SensitiveVec::new(Box::new(Vec::with_capacity(len)));
```
Hinton, line 65: Do we know what happens when a vector grows, will it make a copy of the data or have pointers to the new locations.

crates/bitwarden-crypto/src/sensitive/sensitive.rs
```diff
@@ -0,0 +1,211 @@
...
+    }
+}
+
+impl Sensitive<String> {
+    // We use a lot of `&str` in our tests, so we expose this helper
+    // to make it easier.
+    // IMPORTANT: This should not be used outside of test code
+    // Note that we can't just mark it with #[cfg(test)] because that only applies
+    // when testing this crate, not when testing other crates that depend on it.
+    // By at least limiting it to &'static str we should be able to avoid accidental usages
```
Hinton, line 139: Would exposing a specific feature and only using that feature in tests work?

dani-garcia: Hmm I think it might, we could even enable the feature in dev-dependencies too to avoid requiring it when calling `cargo test`. I'll give it a try!

crates/bitwarden-crypto/src/keys/device_key.rs
```diff
@@ -51,7 +49,7 @@ impl DeviceKey {
             .encrypt_with_key(&device_key.0)?;
 
         Ok(TrustDeviceResponse {
-            device_key: device_key.to_base64(),
+            device_key: device_key.to_base64().expose().to_owned(),
```
Hinton, line 52: Do we need to expose the key or could we change the response?

crates/bitwarden-crypto/src/keys/master_key.rs
```diff
@@ -67,7 +67,7 @@ impl MasterKey {
         let stretched_key = stretch_kdf_key(&self.0)?;
 
         EncString::encrypt_aes256_hmac(
-            user_key.to_vec().as_slice(),
+            user_key.to_vec().expose(),
```
Hinton, line 70: Should we look into if the encrypt methods should accept sensitive values?

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -59,31 +59,35 @@ impl SymmetricCryptoKey {
...
-        buf.zeroize();
-        result
+    pub fn to_base64(&self) -> SensitiveString {
+        self.to_vec().encode_base64(STANDARD)
     }
 
-    pub fn to_vec(&self) -> Zeroizing<Vec<u8>> {
-        let mut buf = Vec::with_capacity(self.total_len());
+    pub fn to_vec(&self) -> SensitiveVec {
+        let mut buf = SensitiveVec::new(Box::new(Vec::with_capacity(self.total_len())));
```
Hinton, line 67: Maybe add a comment why it's important to allocate the full length. It could be seen as an optimization method otherwise.
```suggestion
        // Prevent accidental copies by allocating the full size
        let mut buf = SensitiveVec::new(Box::new(Vec::with_capacity(self.total_len())));
```

crates/bitwarden/src/auth/renew.rs
```diff
@@ -63,8 +63,10 @@ pub(crate) async fn renew_token(client: &mut Client) -> Result<()> {
                     ) = (&result, state_file, client.get_encryption_settings())
                     {
                         if let Some(enc_key) = enc_settings.get_key(&None) {
-                            let state =
-                                ClientState::new(r.access_token.clone(), enc_key.to_base64());
+                            let state = ClientState::new(
```
Hinton, line 66: Should state expect sensitive values directly?

Hinton (commented):
Looks good, some open questions on expectations for inputs/outputs.

crates/bitwarden/src/auth/renew.rs
dani-garcia: Just checked and it's not too big a diff, changed it

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
dani-garcia: Done

crates/bitwarden-crypto/src/keys/master_key.rs
dani-garcia: Definitely, but it needs to be updated at the same time as the KeyEncryptable impls, as they use this function and we don't want to be boxing/unboxing all the time.

crates/bitwarden/src/secrets_manager/state.rs
```diff
@@ -15,11 +15,11 @@ const STATE_VERSION: u32 = 1;
 pub struct ClientState {
     pub(crate) version: u32,
     pub(crate) token: String,
-    pub(crate) encryption_key: String,
+    pub(crate) encryption_key: SensitiveString,
```
Hinton, line 18: Any concerns with backwards compatibility here?

Hinton (approved):
SM state needs to be backwards compatible, did we check if the struct works with old state files?

crates/bitwarden/src/secrets_manager/state.rs
dani-garcia: I gave it a test to be sure now and it works yeah. The struct is only used for serialization/deserialization and the sensitive type works the same way as a plain string there.

crates/bitwarden-crypto/src/uniffi_support.rs
```diff
@@ -43,3 +45,24 @@ impl UniffiCustomTypeConverter for AsymmetricEncString {
...
+        obj.expose().to_owned()
+    }
+}
+
+/// Uniffi doesn't seem to be generating the SensitiveString unless it's being used by
+/// a record somewhere. This is a workaround to make sure the type is generated.
+#[derive(uniffi::Record)]
+struct SupportSensitiveString {
+    sensitive_string: SensitiveString,
+}
```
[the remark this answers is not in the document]
dani-garcia: Opened PM-7214

PR #680, opened 2024-03-26
Expose argon2 in wasm

Hinton:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

We need to expose the argon2 module in WASM for the CLI.

## Before you submit

### DIFF
--- a/crates/bitwarden-wasm/src/client.rs
+++ b/crates/bitwarden-wasm/src/client.rs
@@ -1,6 +1,8 @@
 extern crate console_error_panic_hook;
 use std::rc::Rc;
 
+use argon2::{Algorithm, Argon2, Params, Version};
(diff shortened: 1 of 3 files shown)


crates/bitwarden-wasm/src/client.rs
```diff
@@ -54,3 +56,29 @@ impl BitwardenClient {
...
 }
+
+#[wasm_bindgen]
+pub fn argon2(
+    password: &[u8],
+    salt: &[u8],
+    iterations: u32,
+    memory: u32,
+    parallelism: u32,
+) -> String {
```
Hinton, line 67: We may be able to use Uint8Array here instead and remove the need for b64

Hinton:
Replaced by #691

PR #681, opened 2024-03-26
[PM-7068] - Use a distroless container image for `bws`

tangowithfoxtrot:
## Type of change

- [ ] Bug fix
- [ ] New feature development
- [x] Other

## Objective

Addresses [PM-7068](https://bitwarden.atlassian.net/browse/PM-7068). Build the `bws` Docker image from an empty file system. This results in a much smaller Docker image (~16MB, uncompressed) with a smaller threat surface than bundling it with a distro.

## Code changes

- **./crates/bws/Dockerfile:** Use `scratch` for the final build stage. This results in a distroless image that only contains our binary, the libraries that it depends on, and the CA certificates needed for SSL to work. The `ldd` line automatically determines what dependencies we need to copy over so we don't have to manually maintain a list of them.

## Before you submit

[PM-7068]: https://bitwarden.atlassian.net/browse/PM-7068?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/bws/Dockerfile
+++ b/crates/bws/Dockerfile
@@ -17,10 +17,14 @@ COPY . /app
 WORKDIR /app/crates/bws
 RUN cargo build --release --bin bws
 
+# Bundle bws dependencies
+RUN mkdir /lib-bws
+RUN ldd /app/target/release/bws | tr -s '[:blank:]' '\n' | grep '^/' | xargs -I % cp % /lib-bws
+
 ###############################################
 #                  App stage                  #
 ###############################################
-FROM debian:bookworm-slim
+FROM scratch
 
 ARG TARGETPLATFORM
 LABEL com.bitwarden.product="bitwarden"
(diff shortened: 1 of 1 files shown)


withinfocus:
Novel approach! Is this feasible for what we're offering here? I haven't seen `scratch` used directly like this.

tangowithfoxtrot:
Thanks! I borrowed this approach from some of the ["hello-docker" examples](https://github.com/docker-library/hello-world/blob/master/amd64/hello-world/Dockerfile).

I also use it for [another container](https://github.com/tangowithfoxtrot/dockerfiles/blob/main/speedtest/Dockerfile) that I made for personal use.

This runs every major function of our CLI without issue: listing secrets, getting secrets, mounting config + state files, etc.

Using `ldd` to grab the dynamic dependencies has worked well in my experience. The image could be simplified to remove this in the future as well, if we can statically-link `bws`.

michalchecinski (approved):

Thomas-Avery (approved):
Very cool, thank you!

PR #685, opened 2024-03-28
[PM-7051] Implement make_register_tde_keys

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
Add support in the SDK to create the keys necessary for the just in time user for TDE

### DIFF
--- a/crates/bitwarden/src/auth/tde.rs
+++ b/crates/bitwarden/src/auth/tde.rs
@@ -1,0 +1,52 @@
+use base64::{engine::general_purpose::STANDARD, Engine};
+use bitwarden_crypto::{
+    AsymmetricEncString, AsymmetricPublicCryptoKey, DeviceKey, EncString, SymmetricCryptoKey,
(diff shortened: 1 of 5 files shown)


crates/bitwarden/src/auth/tde.rs
```diff
@@ -0,0 +1,52 @@
+use base64::{engine::general_purpose::STANDARD, Engine};
+use bitwarden_crypto::{
+    AsymmetricEncString, AsymmetricPublicCryptoKey, DeviceKey, EncString, SymmetricCryptoKey,
+    TrustDeviceResponse, UserKey,
+};
+
+use crate::{error::Result, Client};
+
+pub(super) fn make_register_tde_keys(
```
Hinton, line 9: Should we lift up the inline documentation here?  I think the code is pretty self-explanatory but it's good to describe what's happening for consumers.

Hinton (approved):

PR #691, opened 2024-04-01
Publish WASM package to enable argon2 support on CLI

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
Continuation of #680 

I've created workflows to publish the WASM builds to NPM, so that it can be used from the CLI client.

The publishing step is based on the NAPI builds, and I've tested it against the Github NPM registry. We'll need to get a build going and published on the NPM registry to be able to use it on the CLI.


languages/js/wasm/package.json
```diff
@@ -15,6 +15,7 @@
   "main": "node/bitwarden_wasm.js",
   "module": "index.js",
   "types": "bitwarden_wasm.d.ts",
+  "scripts": {},
```
dani-garcia, line 18: This isn't technically required, but npm would output a confusing warning if it wasn't there, so I decided to add it.

crates/bitwarden-wasm/src/client.rs
```diff
@@ -54,3 +55,28 @@ impl BitwardenClient {
...
+            iterations,
+            parallelism,
+            Some(32),
+        )
+        .unwrap(),
+    );
+
+    let mut hash = [0u8; 32];
+    argon.hash_password_into(password, salt, &mut hash).unwrap();
+    hash.to_vec()
```
dani-garcia, line 81: @Hinton Changing this to Vec<u8> generates the Uint8Array bindings and works correctly, as you mentioned in the comment

languages/js/wasm/package.json
```diff
@@ -4,11 +4,13 @@
   "files": [
     "bitwarden_wasm_bg.js",
     "bitwarden_wasm_bg.wasm",
+    "bitwarden_wasm_bg.wasm.d.ts",
+    "bitwarden_wasm_bg.wasm.js",
```
dani-garcia, line 8: This file was missing from here, which causes the package to be published without it. That caused an error when trying to bundle it, as the index.js file tries to import it.

While I was at it, I also added a missing *.d.ts file, and sorted the file list in node/

michalchecinski (changes requested):

michalchecinski (approved):

.github/workflows/release-wasm.yml
```diff
@@ -0,0 +1,133 @@
...
+    inputs:
+      release_type:
+        description: "Release Options"
+        required: true
+        default: "Release"
+        type: choice
+        options:
+          - Release
+          - Redeploy
+          - Dry Run
```
michalchecinski, line 16:
```suggestion
        options:
          - Release
          - Dry Run
```

michalchecinski: @dani-garcia I forgot about this one yesterday, sorry.

michalchecinski (changes requested):

.github/workflows/release-wasm.yml
dani-garcia: No problem, I've removed it now!

michalchecinski: Thanks!

michalchecinski (dismissed):

crates/bitwarden-wasm/Cargo.toml
```diff
@@ -15,6 +15,15 @@ keywords.workspace = true
 crate-type = ["cdylib"]
 
 [dependencies]
+argon2 = { version = ">=0.5.0, <0.6", features = [
+    "alloc",
+    "zeroize",
+], default-features = false }
+base64 = ">=0.21.2, <0.22"
```
Hinton, line 22: I think we can remove base64 again.

crates/bitwarden-wasm/src/client.rs
```diff
@@ -72,11 +72,10 @@ pub fn argon2(
...
-        )
-        .unwrap(),
+        )?,
     );
 
     let mut hash = [0u8; 32];
-    argon.hash_password_into(password, salt, &mut hash).unwrap();
-    hash.to_vec()
+    argon.hash_password_into(password, salt, &mut hash)?;
+    Ok(Vec::new())
```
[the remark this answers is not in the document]
dani-garcia: 🤦 Yeah, I was trying to test the Result type initially and then forgot to revert that

Hinton (approved):

.github/workflows/release-wasm.yml
Hinton, line 1: I think you can get rid of the SETUP NPM by using the `actions/setup-node` action with `registry-url` per https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages#publishing-packages-to-the-npm-registry. You can also omit writing the token to disk and just pass it in as an env during the publish command.

dani-garcia: I'm hesitant to change this here as all our other NPM publishing workflows in sdk and clients are doing it this way, so I wouldn't want a single workflow being different than the rest.

I definitely think this is a good improvement over manually creating the file, and we definitely want to switch to it, but I think it would be better for a separate PR to update all workflows at once.

michalchecinski (approved):

PR #693, opened 2024-04-03
Forbid unwrap usage outside of tests

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Replaced uses of unwrap by ? or .expect(), and added lint to forbid them outside of tests.

### DIFF
--- a/crates/bitwarden-crypto/src/keys/shareable_key.rs
+++ b/crates/bitwarden-crypto/src/keys/shareable_key.rs
@@ -20,14 +20,15 @@ pub fn derive_shareable_key(
     // TODO: Are these the final `key` and `info` parameters or should we change them? I followed
     // the pattern used for sends
     let res = Hmac::<sha2::Sha256>::new_from_slice(format!("bitwarden-{}", name).as_bytes())
-        .unwrap()
+        .expect("hmac new_from_slice should not fail")
         .chain_update(secret)
         .finalize()
         .into_bytes();
 
-    let mut key: Pin<Box<GenericArray<u8, U64>>> = hkdf_expand(&res, info).unwrap();
+    let mut key: Pin<Box<GenericArray<u8, U64>>> =
+        hkdf_expand(&res, info).expect("Input is a valid size");
 
-    SymmetricCryptoKey::try_from(key.as_mut_slice()).unwrap()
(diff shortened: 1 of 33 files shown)


crates/bitwarden-crypto/src/keys/shareable_key.rs
```diff
@@ -20,14 +20,15 @@ pub fn derive_shareable_key(
     // TODO: Are these the final `key` and `info` parameters or should we change them? I followed
     // the pattern used for sends
     let res = Hmac::<sha2::Sha256>::new_from_slice(format!("bitwarden-{}", name).as_bytes())
```
Hinton, line 22: Mind changing this to be `PbkdfSha256Hmac::new_from_slice` while at it?

crates/bws/src/config.rs
```diff
@@ -48,13 +48,13 @@ pub(crate) const DIRECTORY: &str = ".bws";
 
 pub(crate) fn get_config_path(config_file: Option<&Path>, ensure_folder_exists: bool) -> PathBuf {
```
Hinton, line 49: I wonder if this should actually return an error which is propagated up to the CLI?

crates/bitwarden-crypto/src/keys/shareable_key.rs
dani-garcia: Done

crates/bws/src/config.rs
dani-garcia: I didn't want to change it too much in case the function was used in a lot of places but turns out it's only used in this file by functions that already return Result, so I've changed it now.

Hinton (approved):

crates/bws/src/state.rs
```diff
@@ -8,7 +8,7 @@ pub(crate) fn get_state_file_path(
         state_file_path.push(access_token_id);
 
         if let Some(parent_folder) = state_file_path.parent() {
-            std::fs::create_dir_all(parent_folder).unwrap();
+            std::fs::create_dir_all(parent_folder).expect("Parent directory should be writable");
```
Hinton, line 11: Should we convert this to result as well?

Hinton (approved):

PR #695, opened 2024-04-04
[PM-6764] Move cipher to organization

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
Added function to allow moving a cipher to an organization, which requires reencrypting the cipher key if it has one.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -515,6 +515,7 @@ dependencies = [
  "env_logger",
  "schemars",
  "uniffi",
+ "uuid",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


Hinton (approved):
Do we need a clone operation? So that you can move an org item to yourself.

dani-garcia:
I figured that on cipher clone the clients would delete the ID and key themselves so that a different pair can be generated, in which case a separate function isn't needed, but we could add a function to keep the symmetry.

PR #711, opened 2024-04-12
Set login_method when registering for tde

Hinton:
## Type of change

```
- [x] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

We should set login_method when registering TDE keys

## Before you submit


crates/bitwarden/src/auth/login/auth_request.rs
```diff
@@ -86,9 +84,7 @@ pub(crate) async fn complete_auth_request(
     .await?;
 
     if let IdentityTokenResponse::Authenticated(r) = response {
-        let kdf = Kdf::PBKDF2 {
-            iterations: NonZeroU32::new(600_000).expect("Non-zero number"),
-        };
+        let kdf = default_kdf();
```
dani-garcia, line 87: Should we just implement the `Default` trait in `Kdf` instead?

PR #716, opened 2024-04-15
[PM-7447] Publish prerelease to GitHub packages

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

We have a desire to publish every build of the wasm to gh package for faster use in the clients repository.

## Before you submit

### DIFF
--- a/.github/workflows/build-wasm.yml
+++ b/.github/workflows/build-wasm.yml
@@ -62,8 +62,10 @@ jobs:
         run: |
           npm version --no-git-tag-version $(npm view @bitwarden/sdk-wasm@latest version)
           npm version --no-git-tag-version prerelease
+        working-directory: languages/js/wasm
 
       - name: Publish NPM
         run: npm publish
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-wasm.yml
```diff
@@ -56,3 +57,20 @@ jobs:
...
+        run: |
+          npm version --no-git-tag-version $(npm view @bitwarden/sdk-wasm@latest version)
+          npm version --no-git-tag-version prerelease
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        working-directory: languages/js/wasm
+
+      - name: Publish NPM
+        if: ${{ github.ref == 'refs/head/main' }}
+        run: npm publish
```
dani-garcia, line 73: I think this will need to be `npm publish --access public`, otherwise the dependency will be private and only accessible to members of the bitwarden organization, as it's scoped under the @bitwarden name

michalchecinski (approved):

dani-garcia (approved):

PR #720, opened 2024-04-18
Update development section in ruby readme

Thomas-Avery:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [X] Other
```

## Objective

Update README for Ruby language local development. 

## Code changes

## Before you submit

### DIFF
--- a/languages/ruby/README.md
+++ b/languages/ruby/README.md
@@ -106,7 +106,7 @@ Prerequisites:
 cd languages/ruby
 
 # Make the binary folder if it doesn't exist already
-mkdir ./bitwarden_sdk_secrets/lib/macos-arm64
+mkdir -p ./bitwarden_sdk_secrets/lib/macos-arm64
 
 # Build and copy the bitwarden-c library
 cargo build --package bitwarden-c
--- a/languages/ruby/README.md
+++ b/languages/ruby/README.md
@@ -96,7 +96,18 @@ puts response
 
 ## Development
 
+Prerequisites:
+
+- Ruby >= 3.0 installed
+- Generate schemas `npm run schemas`
+
 ```bash
+# Navigate to the ruby language folder
+cd languages/ruby
+
+# Make the binary folder if it doesn't exist already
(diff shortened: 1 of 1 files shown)


languages/ruby/README.md
```diff
@@ -96,7 +96,18 @@ puts response
...
+
+- Ruby >= 3.0 installed
+- Generate schemas `npm run schemas`
+
 ```bash
+# Navigate to the ruby language folder
+cd languages/ruby
+
+# Make the binary folder if it doesn't exist already
+mkdir ./bitwarden_sdk_secrets/lib/macos-arm64
```
dani-garcia, line 109: I think we should change the command to make it create the parent directories if they are missing, and not output an error if the folder exists:
```suggestion
mkdir -p ./bitwarden_sdk_secrets/lib/macos-arm64
```

Thomas-Avery: Great idea added in https://github.com/bitwarden/sdk/pull/720/commits/d597a11b60ff22eeb11bd49ca9ce35a64e377857

languages/ruby/README.md
```diff
@@ -113,8 +124,8 @@ cd ..
 export ACCESS_TOKEN=""
 export ORGANIZATION_ID=""
 
-export API_URL=https://localhost:8080/api
-export IDENTITY_URL=https://localhost:8080/identity
+export API_URL=http://localhost:4000
+export IDENTITY_URL=http://localhost:33656
```
Thomas-Avery, line 128: After messing with this for a bit, I got these proxy endpoints working if: 

- You generate a valid SSL Certificate following instructions [here](https://contributing.bitwarden.com/getting-started/clients/web-vault/#ssl-certificate)
- Run `export NODE_EXTRA_CA_CERTS="$(mkcert -CAROOT)/rootCA.pem"`
- Have the web client running to have these proxies running.

I don't have a strong opinion either way, but we probably need more clarity on prerequisites for using the proxy endpoints.

dani-garcia (approved):

coltonhurst (approved):

PR #723, opened 2024-04-19
Document developer tools

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [ ] New feature development
- [x] Other
```

## Objective
Documented the tools we use for development. Sadly couldn't integrate `nextest` into `bacon` as the output is too different and `bacon` relies on parsing it to work. I still included it in the list for the cases where devs run it manually.

I've also removed the `lto = true` instruction from bitwarden-napi, as it was ignored in favor of the workspace setting and was producing a warning.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,5 +140,24 @@ VALUES
   );
 ```
 
+## Developer tools
+
+This project recommends the use of certain developer tools, and also includes configurations for
+them to make developers lives easier. The use of these tools is optional and they might require a
+separate installation step.
+
+The list of developer tools is:
+
+- `Visual Studio Code`: We provide a recommended extension list which should show under the
+  `Extensions` tab when opening this project with the editor. We also offer a few launch settings
+  and tasks to build and run the SDK
+- `bacon`: This is a CLI background code checker. We provide a configuration file with some of the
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -140,5 +140,24 @@ VALUES
...
+
+The list of developer tools is:
+
+- `Visual Studio Code`: We provide a recommended extension list which should show under the
+  `Extensions` tab when opening this project with the editor. We also offer a few launch settings
+  and tasks to build and run the SDK
+- `bacon`: This is a CLI background code checker. We provide a configuration file with some of the
+  most common tasks to run (`check`, `clippy`, `test`, `doc` - run `bacon -l` to see them all). This
+  tool needs to be installed separately by running `cargo install bacon --locked`.
+- `nexttest`: This is a new and faster test runner, capable of running tests in parallel and with a
```
Hinton, line 157: Should we update bacon test to run nexttest?

Hinton (approved):

README.md
dani-garcia: That's what I tried initially, but bacon relies on parsing the output from cargo test to show the failed test output, which didn't work correctly with cargo nextest sadly.

PR #724, opened 2024-04-22
Make init_crypto take MasterKey

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
By making the `init_crypto_*` functions take `MasterKey` directly, we remove the need for `login_method` to be set before the `init_crypto` call. We also centralize the use of the raw password to the `MasterKey::derive` function, which allows us to more easily secure it.


crates/bitwarden/src/client/client.rs
```diff
@@ -247,23 +247,14 @@ impl Client {
     }
 
     #[cfg(feature = "internal")]
-    pub(crate) fn initialize_user_crypto(
+    pub(crate) fn initialize_user_crypto_password(
```
Hinton, line 250:
```suggestion
    pub(crate) fn initialize_user_crypto_master_key(
```

Hinton (approved):

PR #725, opened 2024-04-22
[PM-7066] Use Sensitive in MasterKey

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Use Sensitive in MasterKey and bubble up the Sensitive change for all the requests that contain a password or pin. Note that this requires changing the request structs to be taken by value.

### DIFF
--- a/crates/bitwarden-crypto/src/keys/master_key.rs
+++ b/crates/bitwarden-crypto/src/keys/master_key.rs
@@ -5,7 +5,9 @@ use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
 use super::utils::{derive_kdf_key, stretch_kdf_key};
-use crate::{util, CryptoError, EncString, KeyDecryptable, Result, SymmetricCryptoKey, UserKey};
+use crate::{
+    util, CryptoError, EncString, KeyDecryptable, Result, SensitiveVec, SymmetricCryptoKey, UserKey,
+};
 
 /// Key Derivation Function for Bitwarden Account
 ///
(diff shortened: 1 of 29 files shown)


crates/bitwarden-crypto/src/keys/master_key.rs
```diff
@@ -69,13 +71,17 @@ impl MasterKey {
     }
 
     /// Derives a users master key from their password, email and KDF.
-    pub fn derive(password: &[u8], email: &[u8], kdf: &Kdf) -> Result<Self> {
-        derive_kdf_key(password, email, kdf).map(Self)
+    pub fn derive(password: &SensitiveVec, email: &[u8], kdf: &Kdf) -> Result<Self> {
+        derive_kdf_key(password.expose(), email, kdf).map(Self)
```
Hinton, line 75: Are we planning on changing `derive_kdf_key` to also accept a sensitive?

crates/bitwarden-crypto/src/sensitive/sensitive.rs
```diff
@@ -67,6 +67,13 @@ impl TryFrom<SensitiveVec> for SensitiveString {
     }
 }
 
+impl From<SensitiveString> for SensitiveVec {
+    fn from(mut s: SensitiveString) -> Self {
+        let value = std::mem::take(&mut s.value);
+        Sensitive::new(Box::new(value.into_bytes()))
```
Hinton, line 73: Is this getting rid of `value` after the move?

crates/bw/src/main.rs
```diff
@@ -192,10 +193,11 @@ async fn process_commands() -> Result<()> {
 
             let email = text_prompt_when_none("Email", email)?;
             let password = Password::new("Password").prompt()?;
+            let password = SensitiveString::new(Box::new(password));
```
Hinton, line 196: Should we inline password::new to prevent possibly leaking?

crates/bw/src/auth/login.rs
```diff
@@ -15,12 +16,13 @@ pub(crate) async fn login_password(mut client: Client, email: Option<String>) ->
     let email = text_prompt_when_none("Email", email)?;
 
     let password = Password::new("Password").without_confirmation().prompt()?;
+    let password = SensitiveString::new(Box::new(password));
```
Hinton, line 19: Inline needed?

crates/bitwarden-crypto/src/keys/master_key.rs
dani-garcia: derive_kdf_key needs a bit of extra work yet, as the current impl is writing the resulting hash to a fixed size array which is causing leak problems. I'm adding some tests for it now and I'll leave it to a separate PR.

crates/bw/src/main.rs
dani-garcia: Done, we'll need to add tests to make sure that the library isn't leaking anything by itself. Not sure how to properly test that yet. Thankfully this isn't used anywhere other than the `bw` crate.

crates/bw/src/auth/login.rs
dani-garcia: Resolved with abead06

crates/bitwarden-crypto/src/sensitive/sensitive.rs
dani-garcia: `into_bytes` consumes the String and returns the internal Vec, so there's no data copied or moved.

Hinton (approved):

PR #726, opened 2024-04-22
Mark generate_random_bytes to return sensitive

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Make `generate_random_bytes` return `Sensitive`. Also migrate `derive_shareable_key` to accept a sensitive input.

## Before you submit

### DIFF
--- a/crates/bitwarden-crypto/src/sensitive/sensitive.rs
+++ b/crates/bitwarden-crypto/src/sensitive/sensitive.rs
@@ -54,16 +54,9 @@ impl<V: Zeroize> Sensitive<V> {
     }
 }
 
-/// Helper to convert a `Sensitive<[u8, 16]>` to a `SensitiveVec`.
-impl From<Sensitive<[u8; 16]>> for SensitiveVec {
-    fn from(sensitive: Sensitive<[u8; 16]>) -> Self {
-        SensitiveVec::new(Box::new(sensitive.value.to_vec()))
-    }
-}
-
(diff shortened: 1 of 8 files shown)


crates/bitwarden-crypto/src/sensitive/sensitive.rs
```diff
@@ -54,6 +54,20 @@ impl<V: Zeroize> Sensitive<V> {
...
+        SensitiveVec::new(Box::new(sensitive.value.to_vec()))
+    }
+}
+
+/// Helper to convert a `Sensitive<[u8, 32]>` to a `SensitiveVec`.
+impl From<Sensitive<[u8; 32]>> for SensitiveVec {
+    fn from(sensitive: Sensitive<[u8; 32]>) -> Self {
+        SensitiveVec::new(Box::new(sensitive.value.to_vec()))
+    }
+}
```
dani-garcia, line 69:
```suggestion
impl<const S: usize> From<Sensitive<[u8; S]>> for SensitiveVec {
    fn from(sensitive: Sensitive<[u8; S]>) -> Self {
        SensitiveVec::new(Box::new(sensitive.value.to_vec()))
    }
}
```
We can replace these two with const generics

Hinton: Nice, I was looking for alternatives but couldn't find any.

dani-garcia (approved):

PR #748, opened 2024-04-29
Convert `to_der` to use Sensitive

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Convert `to_der` and `to_public_der` to return Sensitive wrapped values. To ensure they get dropped when the values are out of scope.

## Before you submit


crates/bitwarden/src/platform/generate_fingerprint.rs
```diff
@@ -45,7 +45,7 @@ pub(crate) fn generate_user_fingerprint(
         .ok_or("Missing private key")?;
 
     let public_key = private_key.to_public_der()?;
-    let fingerprint = fingerprint(&fingerprint_material, &public_key)?;
+    let fingerprint = fingerprint(&fingerprint_material, public_key.expose())?;
```
dani-garcia, line 48: Should we update fingerprint to take the sensitive value directly? Then we would only need to use expose in a single place.

Hinton: I changed public keys to not be sensitive.

PR #752, opened 2024-04-29
[SM-1221] - Statically-linked bins

tangowithfoxtrot:
## Type of change

- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

Produce statically-linked binaries that can be used on GLIBC or MUSL-based Linux distros.

## Code changes

- **.github/workflows/build-cli.yml:** Install and configure Zig as the compiler. This gives us free statically-linked binaries that can also be cross-compiled without the need for `cross`.

## Before you submit

### DIFF
--- a/.github/workflows/build-cli.yml
+++ b/.github/workflows/build-cli.yml
@@ -271,6 +271,10 @@ jobs:
           toolchain: stable
           targets: ${{ matrix.settings.target }}
 
+      - uses: goto-bus-stop/setup-zig@7ab2955eb728f5440978d5824358023be3a2802d # v2.2.0
+        with:
+          version: 0.12.0
+
       - name: Cache cargo registry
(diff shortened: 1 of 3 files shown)


tangowithfoxtrot:
To clarify on removing `cross`: I removed it to simplify the workflow, since we can produce binaries for that target the same way. If we want to use `cross` for that build target, [this commit](https://github.com/bitwarden/sdk/pull/752/commits/aa69f28e2e7a0407d76cd494129a7399fa31023e) can be easily reverted.

Thomas-Avery (approved):
Looks good to me, but I will defer to those with more rust experience if they would like changes.

dani-garcia (approved):
Nice! I knew some people have used the zig compiler in the past to cross compile rust and c dependencies, but didn't know it was this easy!

LGTM

vgrassia (approved):

PR #753, opened 2024-04-30
Add sensitive to password strength and password policy

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Add sensitive to password strength and password policy. This required adding some ways to get the length of a Sensitive String and check if all the characters match a predicate. 

The is_empty added to SensitiveString is because there's a lint for `len` implementation without a corresponding `is_empty`.

### DIFF
--- a/crates/bitwarden/src/auth/password/strength.rs
+++ b/crates/bitwarden/src/auth/password/strength.rs
@@ -1,9 +1,10 @@
+use bitwarden_crypto::SensitiveString;
 use zxcvbn::zxcvbn;
 
 const GLOBAL_INPUTS: [&str; 3] = ["bitwarden", "bit", "warden"];
 
 pub(crate) fn password_strength(
-    password: String,
+    password: SensitiveString,
     email: String,
     additional_inputs: Vec<String>,
 ) -> u8 {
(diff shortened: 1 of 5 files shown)


crates/bitwarden/src/auth/password/strength.rs
```diff
@@ -13,7 +14,7 @@ pub(crate) fn password_strength(
     let mut arr: Vec<_> = inputs.iter().map(String::as_str).collect();
     arr.extend(GLOBAL_INPUTS);
 
-    zxcvbn(&password, &arr).map_or(0, |e| e.score())
+    zxcvbn(password.expose(), &arr).map_or(0, |e| e.score())
```
Hinton, line 17: I guess `zxcvbn` will create copies on their end?

Hinton (approved):

crates/bitwarden/src/auth/password/strength.rs
dani-garcia: They make at least one yeah, this is inside the `zxcvbn` function:
```
let password = password.chars().take(100).collect::<String>();
```

I'll look later into seeing how that can be improved.

PR #763, opened 2024-05-02
Change dependency format for bws

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

The CLI should use the same format for dependency versions.

## Before you submit


dani-garcia (commented):
These are equivalent in Cargo https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements

If we want exact versions we need to use `=x.y.z`

That said I'm okay removing them for consistency with the rest at least

Hinton:
@dani-garcia true, Let's change it to be consistent for now.

dani-garcia (approved):

PR #764, opened 2024-05-03
Hide Sensitive::test behind flag

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

Hide `Sensitive::test` behind a feature flag which is only added as a dev-dependency.

## Before you submit

### DIFF
--- a/crates/bitwarden-crypto/Cargo.toml
+++ b/crates/bitwarden-crypto/Cargo.toml
@@ -17,6 +17,7 @@ keywords.workspace = true
 default = []
 
 mobile = ["dep:uniffi"] # Mobile-specific features
+test = []               # Test methods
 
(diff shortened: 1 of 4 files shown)


dani-garcia (approved):
Nice! LGTM.

About the memory leak test, I've made some changes to the testing framework in this PR https://github.com/bitwarden/sdk/pull/766, and with those changes the test passes again (CI run here: https://github.com/bitwarden/sdk/pull/768) 

I think it is indeed related to the debug builds code generation, as that PR is changing the test to use release builds. I'm not quite sure what part of this PR is triggering the Rust compiler to generate different code than before though.

Edit: I've merged the changes in main and it's working now! 🚀

PR #765, opened 2024-05-03
[WIP] Passkey Uniffi API layer

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective
Created passkeys Uniffi API layer

### DIFF
--- a/crates/bitwarden-uniffi/src/error.rs
+++ b/crates/bitwarden-uniffi/src/error.rs
@@ -14,6 +14,22 @@ impl From<bitwarden::error::Error> for BitwardenError {
     }
 }
 
+impl Into<bitwarden::error::Error> for BitwardenError {
+    fn into(self) -> bitwarden::error::Error {
+        match self {
+            BitwardenError::E(e) => e,
+        }
+    }
+}
+
+// Need to implement this From<> impl in order to handle unexpected callback errors.  See the
+// Callback Interfaces section of the handbook for more info.
+impl From<uniffi::UnexpectedUniFFICallbackError> for BitwardenError {
+    fn from(e: uniffi::UnexpectedUniFFICallbackError) -> Self {
+        Self::E(bitwarden::error::Error::UniffiCallback(e))
+    }
+}
+
 impl Display for BitwardenError {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         match self {
--- a/crates/bitwarden-uniffi/src/error.rs
+++ b/crates/bitwarden-uniffi/src/error.rs
@@ -14,16 +14,17 @@ impl From<bitwarden::error::Error> for BitwardenError {
     }
 }
 
-impl Into<bitwarden::error::Error> for BitwardenError {
-    fn into(self) -> bitwarden::error::Error {
-        match self {
+impl From<BitwardenError> for bitwarden::error::Error {
+    fn from(val: BitwardenError) -> Self {
+        match val {
             BitwardenError::E(e) => e,
         }
     }
 }
 
 // Need to implement this From<> impl in order to handle unexpected callback errors.  See the
-// Callback Interfaces section of the handbook for more info.
+// following page in the Uniffi user guide:
+// <https://mozilla.github.io/uniffi-rs/foreign_traits.html#error-handling>
 impl From<uniffi::UnexpectedUniFFICallbackError> for BitwardenError {
     fn from(e: uniffi::UnexpectedUniFFICallbackError) -> Self {
         Self::E(bitwarden::error::Error::UniffiCallback(e))
--- a/crates/bitwarden-uniffi/src/platform/fido2.rs
+++ b/crates/bitwarden-uniffi/src/platform/fido2.rs
@@ -13,6 +13,8 @@ use bitwarden::{
 
 use crate::{error::Result, Client};
 
+/// At the moment this is just a stub implementation that doesn't do anything. It's here to make
+/// it possible to check the usability API on the native clients.
 #[derive(uniffi::Object)]
 pub struct ClientFido2(pub(crate) Arc<Client>);
 
@@ -181,6 +183,10 @@ pub trait CredentialStore: Send + Sync {
     async fn save_credential(&self, cred: Cipher) -> Result<()>;
 }
 
+// Because uniffi doesn't support external traits, we have to make a copy of the trait here.
+// Ideally we'd want to implement the original trait for every item that implements our local copy,
+// but the orphan rules don't allow us to blanket implement an external trait. So we have to wrap
+// the trait in a newtype and implement the trait for the newtype.
 struct UniffiTraitBridge<T>(T);
 
 #[async_trait::async_trait]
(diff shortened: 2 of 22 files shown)


crates/bitwarden-uniffi/src/error.rs
```diff
@@ -14,6 +14,22 @@ impl From<bitwarden::error::Error> for BitwardenError {
...
+impl Into<bitwarden::error::Error> for BitwardenError {
+    fn into(self) -> bitwarden::error::Error {
+        match self {
+            BitwardenError::E(e) => e,
+        }
+    }
+}
+
+// Need to implement this From<> impl in order to handle unexpected callback errors.  See the
+// Callback Interfaces section of the handbook for more info.
```
Hinton, line 26: Link to the handbook?

crates/bitwarden-uniffi/src/platform/fido2.rs
```diff
@@ -0,0 +1,246 @@
...
+        PublicKeyCredentialAuthenticatorAssertionResponse,
+        PublicKeyCredentialAuthenticatorAttestationResponse,
+    },
+    vault::{Cipher, CipherView, Fido2Credential, Fido2CredentialView},
+};
+
+use crate::{error::Result, Client};
+
+#[derive(uniffi::Object)]
+pub struct ClientFido2(pub(crate) Arc<Client>);
```
Hinton, line 17: Let's document that everything in this struct is stubbed and not actually working.

crates/bitwarden-uniffi/src/platform/fido2.rs
```diff
@@ -0,0 +1,246 @@
...
+#[allow(dead_code)]
+#[derive(uniffi::Record)]
+pub struct CheckUserResult {
+    user_present: bool,
+    user_verified: bool,
+}
+
+#[uniffi::export(with_foreign)]
+#[async_trait::async_trait]
+pub trait UserInterface: Send + Sync {
```
Hinton, line 155: We should document these traits. Follow up ticket?

crates/bitwarden-uniffi/src/platform/fido2.rs
```diff
@@ -0,0 +1,246 @@
...
+    async fn find_credentials(
+        &self,
+        ids: Option<Vec<Vec<u8>>>,
+        rip_id: String,
+    ) -> Result<Vec<Cipher>>;
+
+    async fn save_credential(&self, cred: Cipher) -> Result<()>;
+}
+
+struct UniffiTraitBridge<T>(T);
```
Hinton, line 184: The naming of this is a bit unfortunate, can we add some comments what purpose it serves?

crates/bitwarden-uniffi/src/platform/fido2.rs
```diff
@@ -0,0 +1,246 @@
...
+            .map_err(Into::into)
+    }
+
+    async fn save_credential(&self, cred: Cipher) -> BitResult<()> {
+        self.0.save_credential(cred).await.map_err(Into::into)
+    }
+}
+
+#[derive(uniffi::Record)]
+pub struct CipherViewWrapper {
```
Hinton, line 205: Why do we need a wrapper?

crates/bitwarden/src/error.rs
```diff
@@ -68,10 +69,23 @@ pub enum Error {
     #[error(transparent)]
     ExportError(#[from] ExportError),
 
+    #[error("Webauthn error: {0:?}")]
+    WebauthnError(passkey::client::WebauthnError),
```
Hinton, line 73: Sidenote, webauthn feels like a bad error name. Not our error though.

crates/bitwarden/src/platform/client_platform.rs
```diff
@@ -16,6 +16,12 @@ impl<'a> ClientPlatform<'a> {
     pub fn user_fingerprint(self, fingerprint_material: String) -> Result<String> {
         generate_user_fingerprint(self.client, fingerprint_material)
     }
+
+    pub fn fido2(&'a mut self) -> ClientFido2<'a> {
```
Hinton, line 20: Document that this is a stub?

crates/bitwarden/src/platform/fido2/authenticator.rs
```diff
@@ -0,0 +1,333 @@
...
+    fn is_verification_enabled(&self) -> Option<bool> {
+        Some(true)
+    }
+}
+
+// What type do we need this to be? We probably can't use Serialize over the FFI boundary
+pub type Extensions = Option<std::collections::HashMap<String, String>>;
+
+#[cfg_attr(feature = "mobile", derive(uniffi::Record))]
+pub struct PublicKeyCredentialRpEntity {
```
Hinton, line 257: We should document these.

crates/bitwarden/src/vault/cipher/login.rs
```diff
@@ -97,7 +97,26 @@ pub struct Fido2Credential {
     pub creation_date: DateTime<Utc>,
 }
 
-#[derive(Serialize, Deserialize, Debug, JsonSchema)]
+#[derive(Serialize, Deserialize, Debug, JsonSchema, Clone)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+#[cfg_attr(feature = "mobile", derive(uniffi::Record))]
+pub struct Fido2CredentialView {
```
Hinton, line 103: Future todo. Should we expose all fido stuff decrypted?

Hinton (approved):
Lets add some comments that the implementation is stub for now and we should be good to merge this. We can look into the todos later.

crates/bitwarden/src/vault/cipher/login.rs
dani-garcia: That's something I'm not sure yet, we might want to expose only some parts by having some kind of `Fido2CredentialListView`, like we do with others. This will probably change once we get to actually implement this.

crates/bitwarden-uniffi/src/platform/fido2.rs
dani-garcia: Seems to be a Uniffi limitation, which was breaking the android builds.

When a trait is the defined in the current crate but the return type is defined in an external crate, this would cause issues on the generated Kotlin code.

Using a wrapper is the best way I found to make the return type be in the same crate without creating an entire copy of `CipherView` here.

Hinton: Sounds good, lets make sure to document it though.

crates/bitwarden-uniffi/src/platform/fido2.rs
dani-garcia: This is to work around the fact that we can't blanket implement the `bitwarden` trait over the copy in `bitwarden-uniffi`, I've added a comment now. I agree the name isn't great but it should be pretty limited in scope, let me know if you can think of a better name.

crates/bitwarden-uniffi/src/error.rs
dani-garcia: Done

crates/bitwarden-uniffi/src/platform/fido2.rs
dani-garcia: Done

crates/bitwarden/src/platform/client_platform.rs
dani-garcia: Done

crates/bitwarden-uniffi/src/platform/fido2.rs
```diff
@@ -0,0 +1,256 @@
...
+            .map_err(Into::into)
+    }
+
+    async fn save_credential(&self, cred: Cipher) -> BitResult<()> {
+        self.0.save_credential(cred).await.map_err(Into::into)
+    }
+}
+
+// Uniffi seems to have trouble generating code for Android when a local trait returns a type from
+// an external crate. If the type is small we  can just copy it over and convert back and forth, but
```
Hinton, line 211:
```suggestion
// an external crate. If the type is small we can just copy it over and convert back and forth, but
```

Hinton (approved):

PR #767, opened 2024-05-03
Passkey iOS example

Hinton:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Quick. example of how to use the Passkey APIs on iOS. Android should be mostly the same.

## Before you submit

### DIFF
--- a/languages/swift/iOS/App/ContentView.swift
+++ b/languages/swift/iOS/App/ContentView.swift
@@ -357,9 +357,9 @@ struct ContentView: View {
             options: Options(rk: true, uv: Uv.preferred),
             extensions: nil
         ))
-        
+
         try await authenticator.silentlyDiscoverCredentials(rpId: "")
-        
+
         // Only on android!
         let client = clientFido.client(userInterface: ui, credentialStore: cs)
         try await client.authenticate(origin: "test", request: "test", clientData: ClientData.defaultWithExtraData(androidPackageName: "abc"))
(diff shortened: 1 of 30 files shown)


languages/swift/iOS/App/ContentView.swift
```diff
@@ -320,6 +332,40 @@ struct ContentView: View {
...
+            clientDataHash: Data(),
+            allowList: nil,
+            options: Options(rk: true, uv: Uv.preferred),
+            extensions: nil
+        ))
+        
+        try await authenticator.silentlyDiscoverCredentials(rpId: "")
+        
+        // Only on android!
+        let client = clientFido.client(userInterface: ui, credentialStore: cs)
```
dani-garcia, line 364: Do we want to hide this impl behind a cfg to make sure it's not usable here?

dani-garcia (dismissed):
LGTM, how did you format this code? I couldn't get Xcode to do it for me when I wrote it 😭

Hinton:
Format, `brew install swiftlint, `swiftlint --fix`

languages/swift/iOS/App/ContentView.swift
Hinton: Maybe, I think it's better to do it once we fully understand everything though.

dani-garcia (approved):

PR #772, opened 2024-05-10
[SM-1236] - Add new targets for Go wrapper

tangowithfoxtrot:
## Tracking

[SM-1236](https://bitwarden.atlassian.net/browse/SM-1236)

## Type of change

- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Add the following targets to the build workflow:
- x86_64-pc-windows-gnu
- x86_64-unknown-linux-musl
- aarch64-unknown-linux-musl

These are needed to produced statically-linked binaries with our Go language wrapper.

## Code changes

- Cross-compile the MUSL targets with Zig and add the Windows GNU target. 
- Adjusted the workflow to _just_ build the `bitwarden-c` crate. Previously, it was building every crate in the workspace.

## Before you submit

[SM-1236]: https://bitwarden.atlassian.net/browse/SM-1236?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/build-rust-cross-platform.yml
+++ b/.github/workflows/build-rust-cross-platform.yml
@@ -19,18 +19,19 @@ jobs:
       matrix:
         settings:
           - os: macos-12
-            target:
-              - x86_64-apple-darwin
-              - aarch64-apple-darwin
+            target: x86_64-apple-darwin
+          - os: macos-12
+            target: arch64-apple-darwin
           - os: windows-2022
-            target:
-              - x86_64-pc-windows-msvc
-              - x86_64-pc-windows-gnu
+            target: x86_64-pc-windows-msvc
+          - os: windows-2022
+            target: x86_64-pc-windows-gnu
           - os: ubuntu-22.04
-            target:
-              - x86_64-unknown-linux-gnu
-              - x86_64-unknown-linux-musl
-              - aarch64-unknown-linux-musl
+            target: x86_64-unknown-linux-gnu
+          - os: ubuntu-22.04
+            target: x86_64-unknown-linux-musl
+          - os: ubuntu-22.04
+            target: aarch64-unknown-linux-musl
 
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-rust-cross-platform.yml
```diff
@@ -39,13 +45,29 @@ jobs:
...
+        env:
+          RUSTFLAGS: "-D warnings"
+        run: cargo zigbuild -p bitwarden-c --target ${{ matrix.settings.target }} --release
+
+      - name: Build Rust for - ${{ matrix.settings.target }}
+        if: ${{ !contains(matrix.settings.target, 'musl') }}
         env:
           RUSTFLAGS: "-D warnings"
-        run: cargo build --target ${{ matrix.settings.target }} --release
+        run: cargo build -p bitwarden-c --target ${{ matrix.settings.target }} --release
```
tangowithfoxtrot, line 70: I adjusted this workflow so that it _only_ builds the artifacts for the `bitwarden-c` crate. Let me know if the other builds are actually necessary in this workflow. As far as I can tell, they aren't getting used anywhere else, but I could have missed something

dani-garcia: Yeah, we are only uploading artifacts for files matching `*bitwarden_c*`, so this seems fine to me.

dani-garcia (approved):
LGTM!

coltonhurst (approved):
@tangowithfoxtrot and I discussed how we are adding more toolchains and a desire to reduce the surface in this area. We'll address this idea after this project as Zig has already been introduced in a recent PR.

Thanks for the work on this!

michalchecinski (approved):

PR #788, opened 2024-05-16
[PM-8275] Implement a zeroizing allocator

Hinton:
## Type of change

```
- [ ] Bug fix
- [x] New feature development
- [ ] Other
```

## Objective

Add a custom allocator which zerorizes memory on drop. This should ensure everything heap allocated gets sanitized before deallocated, and avoid sensitive memory being exposed more reliably than using `Sensitive`.

## Before you submit


crates/bitwarden-crypto/src/allocator.rs
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    unsafe fn alloc_zeroed(&self, layout: Layout) -> *mut u8 {
+        self.0.alloc_zeroed(layout)
+    }
+}
+
+/*
+#[cfg(test)]
+mod tests {
```
Hinton, line 32: These tests are disabled since they work inconsitently. Sometimes the memory is freed and immediately overwritten.

It's never the same value as what used to be there though.

dani-garcia: We can mark the test as ignored instead. That way it won't be run by default, but can be run with `cargo test --ignored`:
```rust
#[test]
#[ignore = "It produces inconsistent results on some platforms"]
fn test_fn() {
}
```

dani-garcia (commented):
Allocator look good to me and seems to be working correctly, but I'm not sure about having to define `#[global_allocator]` in every crate. It might make us accidentally forget to enable it in some future new crate. We should also probably enable it for `bitwarden-uniffi`.

What if instead we defined the global allocator in the `bitwarden` crate under some `secure-allocator` feature flag which is enabled by default? That way all our crates will get the new allocator for free, and anyone using our Rust crate who might prefer a different allocator can just disable the default flags.

crates/bitwarden/Cargo.toml
```diff
@@ -16,26 +16,28 @@ license-file.workspace = true
...
+secrets = [] # Secrets manager API
 wasm-bindgen = ["chrono/wasmbind"]
 
 [dependencies]
 async-trait = ">=0.1.80, <0.2"
 base64 = ">=0.21.2, <0.23"
 bitwarden-api-api = { workspace = true }
 bitwarden-api-identity = { workspace = true }
-bitwarden-crypto = { workspace = true }
+bitwarden-crypto = { workspace = true, no-default-features = true }
```
dani-garcia, line 40: Should this not be `default-features = false`?  That's how it works outside of workspace dependencies at least
https://github.com/bitwarden/sdk/blob/68a0e4edda5d408ea2a499082c39879d03e87b9b/crates/bitwarden-uniffi/Cargo.toml#L26-L29

Weird that it doesn't give an error or warning either way.

crates/bitwarden-c/Cargo.toml
```diff
@@ -16,7 +16,7 @@ bench = false
 
 [target.'cfg(not(target_arch="wasm32"))'.dependencies]
 tokio = { version = ">=1.28.2, <2.0", features = ["rt-multi-thread", "macros"] }
-
+bitwarden = { path = "../bitwarden", features = ["secrets"] }
```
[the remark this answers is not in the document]
Hinton: It was needed before with the allocator. Moving it to crypto this is no longer the case.

crates/bitwarden-crypto/Cargo.toml
```diff
@@ -16,8 +16,9 @@ keywords.workspace = true
 [features]
 default = []
 
-mobile = ["dep:uniffi"] # Mobile-specific features
-test = []               # Test methods
+no-memory-hardening = [] # Disable memory hardening features
```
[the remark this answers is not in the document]
Hinton: Yes, with an additive feature the benchmarks breaks. I also wanted to avoid people opting out of secrets manager from missing the memory hardening.

dani-garcia (approved):

crates/bitwarden-crypto/src/lib.rs
MGibson1, line 1: Please add features documentation to this front page as described in https://doc.rust-lang.org/cargo/reference/features.html#feature-documentation-and-discovery.

Especially since you're worried about unintentionally disabling memory hardening and thus not following the additive feature standard

Hinton: Good point. I've added some documentation.

dani-garcia (approved):

dani-garcia (approved):

PR #790, opened 2024-05-17
SM-1237: Decrease bitwarden-c static archive size

coltonhurst:
## Type of change

- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

Reduce static library size when building the `bitwarden-c` crate.

## Code changes

- **crates/bitwarden-c/Cargo.toml:** Update the crate types for `bitwarden-c`

## Before you submit

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,8 +41,8 @@ opt-level = 1
 
 # Turn on LTO on release mode
 [profile.release]
-lto = "fat"
+lto = "thin"
 codegen-units = 1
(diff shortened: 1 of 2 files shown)


dani-garcia (approved):
LGTM!

Not sure what it's about the `lib` type that causes the compiler to build the staticlib unoptimally, but with this change, `libbitwarden_c.a` goes from 60MB to 14MB on my system.

Thankfully this shouldn't affect anything negatively as we never use `bitwarden-c` as a rust library.

PR #796, opened 2024-05-22
Update OpenAPI templates and generator

dani-garcia:
## Type of change
```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective
Update OpenAPI templates and generator to the latest version available:
- Generator: 7.6.0
- Templates: https://github.com/OpenAPITools/openapi-generator/commit/638af0fe37e2786ca3d6f5a8d764b346b2cc74a6

The first commit (556440739798c2735ad1e1b12e930f4e64e7b57c) contains the changes from upstream as-is, the second commit (1249b1240ce0bf5746a41018c3f8eac5ccaea2a3) contains all the changes we need to make. This way it should be easier to apply the changes in the future.

The diff with upstream is a bit smaller now and also does less unnecessary allocations. Also I've updated the Cargo.toml template to more closely match ours. Once this is approved I'll make a separate PR to regenerate the bindings.

The changes to the generated bindings are fairly minor:
- The bindings are now capable of base64 decoding values that are marked as `[JsonConverter(typeof(Base64UrlConverter))]` on the server. We have some API endpoints that make use of it on the server but at the moment we're not using them.
- The model structs now implement Default
- The resulting code generates less warnings now, so we can be more selective with what warnings we disable.


Hinton (commented):
Can we open a PR with the changes for the bindings? I think the templates looks good but would be nice to look at the bindings before merging this.

dani-garcia:
@Hinton Sounds good, here are the bindings https://github.com/bitwarden/sdk/pull/797

Hinton (approved):

PR #798, opened 2024-05-24
[PM-8301] Invert crates relationship

Hinton:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [ ] Other
```

## Objective

This PR inverts the relationship between most crates in an effort to make the code base easier to maintain.

- Moves most of the logic into `bitwarden-core`.
- `bitwarden` provides re-exports from the various crates.
- Moved `ClientFeature` structs to the various crates, and introduced `ClientFeatureExt` which exposes it on `Client`. This now becomes the typical way to use functionality.
- Created `InternalClient` which contains the internal functionality for hooking up logic between the various crates.

## Before you submit


crates/bitwarden-core/src/auth/client_auth.rs
```diff
@@ -169,6 +165,7 @@ impl<'a> Client {
     }
 }
 
+/*
```
dani-garcia, line 168: This test is commented, should this be fixed or moved somewhere else?

dani-garcia (commented):
Overall looks good to me, but some imports in the Kotlin demo need to be updated, and there's a commented test

crates/bitwarden-core/src/auth/client_auth.rs
Hinton: It's more of an integration test. I'm thinking we could move them to the bitwarden crate under a separate test directory.

dani-garcia (approved):

PR #820, opened 2024-06-06
[PM-8657] Handle attachments with cipher keys

dani-garcia:
## 📔 Objective

- Added support for decrypting attachments without keys, for backwards compatibility
- Updated `move_to_organization` and `generate_cipher_key` to reencrypt attachment keys when possible or error out when it's not.
- Fix `generate_cipher_key` incorrectly encrypting the new cipher key with the old cipher key when the cipher already had a key.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden/src/mobile/vault/client_ciphers.rs
+++ b/crates/bitwarden/src/mobile/vault/client_ciphers.rs
@@ -17,7 +17,18 @@ impl<'a> ClientCiphers<'a> {
 
         // TODO: Once this flag is removed, the key generation logic should
         // be moved directly into the KeyEncryptable implementation
-        if cipher_view.key.is_none() && self.client.get_flags().enable_cipher_key_encryption {
+
+        // If any attachments have no keys, they will need to be reuploaded before we can
+        // generate them a cipher key, so in that case we skip them even if the flag is set
+        let all_attachments_have_keys = cipher_view
+            .attachments
+            .iter()
+            .flatten()
+            .all(|a| a.key.is_some());
+        if cipher_view.key.is_none()
+            && self.client.get_flags().enable_cipher_key_encryption
+            && all_attachments_have_keys
+        {
             let key = cipher_view
                 .locate_key(enc, &None)
                 .ok_or(Error::VaultLocked)?;
--- a/crates/bitwarden/src/mobile/vault/client_ciphers.rs
+++ b/crates/bitwarden/src/mobile/vault/client_ciphers.rs
@@ -155,14 +155,7 @@ mod tests {
             edit: true,
             view_password: true,
             local_data: None,
-            attachments: Some(vec![Attachment {
-                id: Some("uf7bkexzag04d3cw04jsbqqkbpbwhxs0".to_string()),
-                url: Some("http://localhost:4000/attachments//358f2b2b-9326-4e5e-94a8-b18100bb0908/uf7bkexzag04d3cw04jsbqqkbpbwhxs0".to_string()),
-                file_name: Some("2.mV50WiLq6duhwGbhM1TO0A==|dTufWNH8YTPP0EMlNLIpFA==|QHp+7OM8xHtEmCfc9QPXJ0Ro2BeakzvLgxJZ7NdLuDc=".parse().unwrap()),
-                key: None,
-                size: Some("65".to_string()),
-                size_name: Some("65 Bytes".to_string()),
-            }]),
+            attachments: None,
             fields:  None,
             password_history: None,
             creation_date: "2024-05-31T11:20:58.4566667Z".parse().unwrap(),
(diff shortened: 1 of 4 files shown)


crates/bitwarden/src/vault/cipher/attachment.rs
```diff
@@ -80,15 +80,20 @@ impl KeyDecryptable<SymmetricCryptoKey, Vec<u8>> for AttachmentFile {
         let ciphers_key = Cipher::get_cipher_key(key, &self.cipher.key)?;
         let ciphers_key = ciphers_key.as_ref().unwrap_or(key);
 
-        let mut attachment_key: Vec<u8> = self
+        // Decrypt the attachment key if there is one
```
Hinton, line 83: I wonder if this code would be more easily understood,

```rust
/// Version 2 or 3, `AttachmentKey` or `CipherKey(AttachmentKey)`
if let Some(attachment_key) = &self.attachment.key {
    let mut content_key: Vec<u8> = attachment_key.decrypt_with_key(ciphers_key)?;
    let content_key = SymmetricCryptoKey::try_from(content_key.as_mut_slice())?;

    self.contents.decrypt_with_key(&content_key)
} else {
    /// Legacy attachment, use user/org key
    self.contents.decrypt_with_key(key)
}
```

Hinton (commented):
Can we add the following documentation to `pub contents: EncString`

```rust
    /// There are three different ways attachments are encrypted.
    /// 1. UserKey / OrgKey (Contents) - Legacy
    /// 2. AttachmentKey(Contents) - Pre CipherKey
    /// 3. CipherKey(AttachmentKey(Contents)) - Current
    pub contents: EncString,
```

crates/bitwarden/src/vault/cipher/attachment.rs
dani-garcia: Agreed, changed it in 0cbed5cbbaa01749092781d66be891a7aa6ffa02

dani-garcia:
Both `move_to_organization` and `generate_cipher_key` should correctly reencrypt attachment keys now, or error if they don't have keys.

I've also fixed a bug with `generate_cipher_key`, previously if the cipher already had a key, we would encrypt the new cipher key with the old cipher key, instead of with the user key. This isn't a problem as we don't regenerate the key for ciphers that already have one, but I noticed it while implementing this. I can split that change to another PR if you prefer.

crates/bitwarden/src/vault/cipher/cipher.rs
```diff
@@ -366,23 +368,49 @@ impl CipherView {
...
+                return Err("This cipher contains attachments without keys. Those attachments need to be reuploaded before the cipher can be moved".into());
+            }
+
+            for attachment in attachments {
+                let attachment_key = attachment
+                    .key
+                    .as_mut()
+                    .expect("Key is asserted to be present");
+                let dec_attachment_key: Vec<u8> = attachment_key.decrypt_with_key(old_key)?;
+                *attachment_key = dec_attachment_key.encrypt_with_key(new_key)?;
```
Hinton, line 389: Should we expose a helper on the `AttachmentView` instead?

crates/bitwarden/src/vault/cipher/cipher.rs
```diff
@@ -343,11 +343,13 @@ fn build_subtitle_identity(first_name: Option<String>, last_name: Option<String>
...
 impl CipherView {
     pub fn generate_cipher_key(&mut self, key: &SymmetricCryptoKey) -> Result<()> {
-        let ciphers_key = Cipher::get_cipher_key(key, &self.key)?;
-        let key = ciphers_key.as_ref().unwrap_or(key);
+        let old_ciphers_key = Cipher::get_cipher_key(key, &self.key)?;
+        let old_key = old_ciphers_key.as_ref().unwrap_or(key);
 
         let new_key = SymmetricCryptoKey::generate(rand::thread_rng());
 
+        self.reencrypt_attachment_keys(old_key, &new_key)?;
```
Hinton, line 351: Could this be confusing for the mobile apps? Since we won't support saving older items without the attachments being updated first. Should we avoid generating a cipher key in that scenario.

crates/bitwarden/src/vault/cipher/cipher.rs
dani-garcia: My thinking here was that we always want to reencrypt all the attachments in one go or return an error, which is also why I'm checking if all the attachments have a key before doing the reencryption.

If we exposed this as a helper in `AttachmentView` it would be easier to end up in a situation where we reencrypt one attachment but not another.

That said I think we're doing reencryption in enough places that it would be convenient to have some generic helper functions for it, and that would also simplify this code a bit more.

crates/bitwarden/src/vault/cipher/cipher.rs
Hinton: After discussion we should only update attachments keys if they use cipher keys here. If not it should still be the user key 🤯

crates/bitwarden/src/mobile/vault/client_ciphers.rs
```diff
@@ -17,7 +17,18 @@ impl<'a> ClientCiphers<'a> {
...
+        // generate them a cipher key, so in that case we skip them even if the flag is set
+        let all_attachments_have_keys = cipher_view
+            .attachments
+            .iter()
+            .flatten()
+            .all(|a| a.key.is_some());
+        if cipher_view.key.is_none()
+            && self.client.get_flags().enable_cipher_key_encryption
+            && all_attachments_have_keys
+        {
```
Hinton, line 31: I believe the other clients do re-encrypt ciphers with attachments. They just leave attachments keys as null.

dani-garcia: Ah okay, I changed it in 817c9fffe978b6cd024766fb7613940827455c33.

Now when encrypting a cipher we always leave null attachment keys as null (so it continues using the org/user key). When the keys are not null we reencrypt as usual.

In the move to org case we still need to error out if the keys are null.

Does that sound about right?

crates/bitwarden/src/vault/cipher/cipher.rs
```diff
@@ -366,23 +368,48 @@ impl CipherView {
...
 
-            let new_key = enc
-                .get_key(&Some(organization_id))
-                .ok_or(Error::VaultLocked)?;
+        let new_key = enc
+            .get_key(&Some(organization_id))
+            .ok_or(Error::VaultLocked)?;
 
+        // If the cipher has a key, we need to re-encrypt it with the new organization key
+        if let Some(cipher_key) = &mut self.key {
```
Hinton, line 407: I think we need to check the attachments here too?

dani-garcia: Ah yes, I lifted the condition before this point so it covers both cases

Hinton (approved):

PR #827, opened 2024-06-10
Clear stack after argon2

dani-garcia:
## 📔 Objective

The argon2 library doesn't seem to clear the stack after doing a hash, but we can overwrite the stack by calling a function that creates a big stack allocated array right after the hash operation.

I've tried changing the parameters to the argon2 function and tweaking the array size to see where the limit should be, and in most of my tests clearing around 3700 bytes is enough. I've decided to round it up to the page size to be safe.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/memory-testing/cases.json
+++ b/crates/memory-testing/cases.json
@@ -80,8 +80,7 @@
       "memory_lookups": [
         {
           "name": "Key",
-          "hex": "3bc0520a0abff0097d521ce0ee5e5b1cee301939a84742623c0c1697d7a4bd46",
-          "allowed_count": 3
+          "hex": "3bc0520a0abff0097d521ce0ee5e5b1cee301939a84742623c0c1697d7a4bd46"
         },
         {
           "name": "Hash B64",
--- a/crates/bitwarden-crypto/src/keys/utils.rs
+++ b/crates/bitwarden-crypto/src/keys/utils.rs
@@ -32,6 +32,15 @@ pub(super) fn derive_kdf_key(secret: &[u8], salt: &[u8], kdf: &Kdf) -> Result<Sy
 
             let mut hash = [0u8; 32];
             argon.hash_password_into(secret, &salt_sha, &mut hash)?;
+
+            // Argon2 is using some stack memory that is not zeroed. Eventually some function will
+            // overwrite the stack, but we use this trick to force the used stack to be zeroed.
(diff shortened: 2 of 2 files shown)


crates/memory-testing/cases.json
```diff
@@ -80,8 +80,7 @@
       "memory_lookups": [
         {
           "name": "Key",
-          "hex": "3bc0520a0abff0097d521ce0ee5e5b1cee301939a84742623c0c1697d7a4bd46",
-          "allowed_count": 3
```
Hinton, line 84: Should we remove support for `allowed_count`?

Hinton (approved):
Nice, did you explore how large the stack clear has to be?

dani-garcia:
With the current test case the array had to be defined to be exactly 3872 bytes long. If I reduced it to 3871 it would consistently fail.

I also tried to increase the memory used by argon2 from 4MB to 64MB and the amount of memory needed is exactly the same. Previously I got a passing test with a smaller stack value but I can't reproduce it anymore, so it was probably a mistake.

The test case used:
```
    {
      "name": "Master Key Argon2 64MB",
      "master_key": {
        "password": "asdfasdfasdf",
        "email": "test@mail.com",

        "kdf": {
          "argon2id": {
            "iterations": 4,
            "memory": 64,
            "parallelism": 1
          }
        }
      },
      "memory_lookups": [
        {
          "name": "Key",
          "hex": "606ba237ae8c02df6ebf7835cdc3a4fe82a6a4156cc8ec124faf96ab401bbfaf"
        },
        {
          "name": "Hash B64",
          "string": "9AKG6mjR9NlqfICE+zfwO3hhXQB0TeluknocoMI4G7U="
        },
        {
          "name": "Hash bytes",
          "hex": "f40286ea68d1f4d96a7c8084fb37f03b78615d00744de96e927a1ca0c2381bb5"
        }
      ]
    }
```

crates/memory-testing/cases.json
dani-garcia: I'm okay removing it, but it's only two lines of code so I don't think it's a problem either way. 

That said now with the custom allocator hopefully we won't need to allow many leaks.

AaronFeickert:
Has this been reported or investigated upstream?

PR #831, opened 2024-06-11
Fix save_credential failing when the selected cipher doesn't have a Passkey

dani-garcia:
## 📔 Objective

The current call to `get_selected_credential` inside `save_credential` will try to fetch and decrypt the ciphers FIDO2 credentials, and error if they are not there. 

This can only happen when creating a new Passkey, so instead of calling `get_selected_credential` we just get the value from the lock. The other places where `get_selected_credential` is used is in `update_credential` and right at the end of the `assertion`, `register` and `authenticate` operations, so those should be safe.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden/src/platform/fido2/authenticator.rs
+++ b/crates/bitwarden/src/platform/fido2/authenticator.rs
@@ -311,7 +311,14 @@ impl passkey::authenticator::CredentialStore for CredentialStoreImpl<'_> {
             let cred = try_from_credential_full(cred, user, rp)?;
 
             // Get the previously selected cipher and add the new credential to it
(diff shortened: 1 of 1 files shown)


coroiu (dismissed):
Just one small suggestion

dani-garcia:
I also thought about entirely renaming the variable from `selected_credential` to `selected_cipher` here, knowing that the value doesn't necessarily need to contain a credential.

I didn't do it in this PR so we could get it resolved quickly but it might be worth considering.

coroiu (approved):
Ah yeah good point, let's follow up with that then

Hinton (approved):

PR #70, opened 2023-06-14
SM-805: Put ApiConfigurations & EncryptionSettings in Client behind RwLock & Arc

coltonhurst:
## Type of change

```
- [ ] Bug fix
- [ ] New feature development
- [x] Tech debt (refactoring, code cleanup, dependency upgrades, etc)
- [ ] Build/deploy pipeline (DevOps)
- [ ] Other
```

## Objective

Make client functions immutable

## Code changes

- **file.ext:** Description of what was changed and why

## Before you submit

- Please add **unit tests** where it makes sense to do so (encouraged but not required)

PR #837, opened 2024-06-12
Make client internally mutable

dani-garcia:
## 📔 Objective

We're having some deadlock problems with the Passkey API by calling certain SDK functions on the `passkey-rs` callbacks. These issues are caused because we're holding a reference to the RwLock during the execution of the Passkey operation and so calling into the SDK from inside that will deadlock.

To solve it we can't just use a lock at the client level, and so we need to move to using interior mutability, I've revived an old PR (#70) where we started doing that.

This PR is split into two commits, for ease of review: 
- [Remove all the muts and locks for uniffi](https://github.com/bitwarden/sdk/commit/c56add5c181ecd6028df3c6710da6327b2ad9166): This removes the use of RwLock in bitwarden-uniffi and the `&mut` references used in the bitwarden Client API. Note that this alone won't compile.
- [Make client internally mutable](https://github.com/bitwarden/sdk/commit/f61d10496ea5724b639ba991a7b953a7852ca34c): Implements RwLocks in Client to make it internally mutable.

Some notes of what the change entails:
- I had to wrap `EncryptionSettings` in a `RwLock<Arc<>>` to get the API working, this is quite unfortunate as it forces us to clone `EncryptionSettings` to modify it when calling `set_org_keys`. We can't wrap the `org_keys` in a `RwLock` as we will get lifetime issues in `get_key`. I think the best solution for this will be the work on creating a Crypto store that only returns opaque key references, so we might want to prioritize that.
- I extracted `external_client` from `ApiConfigurations`, as it's never mutated and it allows us to keep the same API for `get_http_client`

I hate that `Mutex` and `RwLock` return a Result in case they are poisoned, that's never our case and it makes their usage so cumbersome 😢 

Note: I've tried to change the smallest amount of code to make the change, so the secrets manager APIs (and bitwarden-json) are still mutable. If we're okay with these changes I can go back to update those in a separate PR.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden/src/client/client.rs
+++ b/crates/bitwarden/src/client/client.rs
@@ -152,7 +152,7 @@ impl Client {
     }
 
     #[cfg(feature = "internal")]
-    pub fn load_flags(&mut self, flags: std::collections::HashMap<String, bool>) {
+    pub fn load_flags(&self, flags: std::collections::HashMap<String, bool>) {
         self.flags = Flags::load_from_map(flags);
     }
 
@@ -161,7 +161,7 @@ impl Client {
         &self.flags
     }
 
-    pub(crate) async fn get_api_configurations(&mut self) -> &ApiConfigurations {
+    pub(crate) async fn get_api_configurations(&self) -> &ApiConfigurations {
         // At the moment we ignore the error result from the token renewal, if it fails,
         // the token will end up expiring and the next operation is going to fail anyway.
         self.auth().renew_token().await.ok();
@@ -192,19 +192,14 @@ impl Client {
         self.encryption_settings.as_ref().ok_or(VaultLocked.into())
     }
 
-    pub(crate) fn set_login_method(&mut self, login_method: LoginMethod) {
+    pub(crate) fn set_login_method(&self, login_method: LoginMethod) {
         use log::debug;
 
         debug! {"setting login method: {:#?}", login_method}
         self.login_method = Some(login_method);
     }
 
-    pub(crate) fn set_tokens(
-        &mut self,
-        token: String,
-        refresh_token: Option<String>,
-        expires_in: u64,
-    ) {
+    pub(crate) fn set_tokens(&self, token: String, refresh_token: Option<String>, expires_in: u64) {
         self.token = Some(token.clone());
         self.refresh_token = refresh_token;
         self.token_expires_on = Some(Utc::now().timestamp() + expires_in as i64);
@@ -219,7 +214,7 @@ impl Client {
 
     #[cfg(feature = "internal")]
     pub(crate) fn initialize_user_crypto_master_key(
-        &mut self,
+        &self,
         master_key: MasterKey,
         user_key: EncString,
         private_key: EncString,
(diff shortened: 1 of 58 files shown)


crates/bitwarden/src/secrets_manager/secrets/update.rs
```diff
@@ -44,5 +44,5 @@ pub(crate) async fn update_secret(
 
     let enc = client.get_encryption_settings()?;
```
Hinton, line 45: I think we can reuse the same encryption setting as above?

crates/bitwarden/src/secrets_manager/secrets/create.rs
```diff
@@ -48,5 +48,5 @@ pub(crate) async fn create_secret(
 
     let enc = client.get_encryption_settings()?;
```
Hinton, line 49: I think we can reuse the same encryption setting above?

crates/bitwarden/src/secrets_manager/projects/update.rs
```diff
@@ -39,5 +39,5 @@ pub(crate) async fn update_project(
 
     let enc = client.get_encryption_settings()?;
```
Hinton, line 40: I think we can reuse the same `enc` as above?

crates/bitwarden/src/secrets_manager/projects/create.rs
```diff
@@ -40,5 +40,5 @@ pub(crate) async fn create_project(
 
     let enc = client.get_encryption_settings()?;
```
Hinton, line 41: I think we can reuse the same `enc` as above?

crates/bitwarden/src/mobile/tool/client_sends.rs
```diff
@@ -75,11 +75,8 @@ impl<'a> ClientSends<'a> {
...
-        let key = self
-            .client
-            .get_encryption_settings()?
-            .get_key(&None)
-            .ok_or(VaultLocked)?;
+        let enc = self.client.get_encryption_settings()?;
+        let key = enc.get_key(&None).ok_or(VaultLocked)?;
         let key = Send::get_key(&send.key, key)?;
 
         let enc = buffer.encrypt_with_key(&key)?;
```
Hinton, line 82: Should we rename this now that `enc` exists above but is a different type.

crates/bitwarden/src/client/client.rs
```diff
@@ -69,20 +69,24 @@ pub(crate) enum ServiceAccountLoginMethod {
 /// The main struct to interact with the Bitwarden SDK.
 #[derive(Debug)]
 pub struct Client {
-    token: Option<String>,
-    pub(crate) refresh_token: Option<String>,
-    pub(crate) token_expires_on: Option<i64>,
-    pub(crate) login_method: Option<LoginMethod>,
+    token: RwLock<Option<String>>,
+    pub(crate) refresh_token: RwLock<Option<String>>,
+    pub(crate) token_expires_on: RwLock<Option<i64>>,
```
Hinton, line 74: I wonder if we should find some way to group these together. You shouldn't be able to write to just one of these.

dani-garcia: Yeah, that's true, I extracted them to a tokens struct, which at least simplifies a bit all the locking and unlocking.

crates/bitwarden/src/mobile/tool/client_sends.rs
dani-garcia: Good point, changed

Hinton (approved):

PR #849, opened 2024-06-17
Add fido2 credential autofill view

coroiu:
## 📔 Objective

This PR adds a function on the Fido2 authenticator which returns all available credentials as a View tailored specifically for integrating with OS-level autofill APIs

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden-fido/Cargo.toml
+++ b/crates/bitwarden-fido/Cargo.toml
@@ -27,12 +27,12 @@ chrono = { version = ">=0.4.26, <0.5", features = [
     "serde",
 ], default-features = false }
 coset = { version = "0.3.7" }
-schemars = { version = ">=0.8.9, <0.9", features = ["uuid1", "chrono"] }
 itertools = "0.13.0"
 log = ">=0.4.18, <0.5"
 p256 = { version = ">=0.13.2, <0.14" }
 passkey = { git = "https://github.com/bitwarden/passkey-rs", rev = "c48c2ddfd6b884b2d754432576c66cb2b1985a3a" }
 reqwest = { version = ">=0.12, <0.13", default-features = false }
+schemars = { version = "0.8.21", features = ["uuid1", "chrono"] }
 serde = { version = ">=1.0, <2.0", features = ["derive"] }
 serde_json = ">=1.0.96, <2.0"
 thiserror = ">=1.0.40, <2.0"
--- a/crates/bitwarden-fido/Cargo.toml
+++ b/crates/bitwarden-fido/Cargo.toml
@@ -27,6 +27,8 @@ chrono = { version = ">=0.4.26, <0.5", features = [
     "serde",
 ], default-features = false }
 coset = { version = "0.3.7" }
+schemars = { version = ">=0.8.9, <0.9", features = ["uuid1", "chrono"] }
+itertools = "0.13.0"
 log = ">=0.4.18, <0.5"
 p256 = { version = ">=0.13.2, <0.14" }
 passkey = { git = "https://github.com/bitwarden/passkey-rs", rev = "c48c2ddfd6b884b2d754432576c66cb2b1985a3a" }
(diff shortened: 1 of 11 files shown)


crates/bitwarden-uniffi/src/platform/fido2.rs
```diff
@@ -216,6 +246,8 @@ pub trait Fido2CredentialStore: Send + Sync {
         rip_id: String,
     ) -> Result<Vec<CipherView>, Fido2CallbackError>;
 
+    async fn all_credentials(&self) -> Result<Vec<CipherView>, Fido2CallbackError>;
```
dani-garcia, line 249: Instead of adding a separate function, could we update `find_credentials` to make `rp_id` optional just like the `ids` list is?

Then we could also have `silently_discover_credentials` take an optional `rp_id` which would save us from having a mostly duplicate `credentials_for_autofill` function and error type.

crates/bitwarden-fido/src/types.rs
```diff
@@ -1,8 +1,84 @@
...
+    pub fn from_cipher_view(
+        cipher: &CipherView,
+        enc: &dyn KeyContainer,
+    ) -> Result<Vec<Fido2CredentialAutofillView>, Fido2CredentialAutofillViewError> {
+        let credentials = cipher.decrypt_fido2_credentials(enc)?;
+
+        credentials
+            .into_iter()
+            .filter_map(|c| -> Option<Result<_, Fido2CredentialAutofillViewError>> {
+                c.user_handle.map(|user_handle| {
```
dani-garcia, line 63: Could we move this to an `impl TryFrom<Fido2CredentialView> for Fido2CredentialAutofillView`?

dani-garcia (commented):
Overall looks good to me, just a question about some duplication introduced by the new `credentials_for_autofill`

crates/bitwarden-fido/src/types.rs
coroiu: No because `Fido2CredentialAutofillView` has information from other places than just the `Fido2CredentialView`:

- `cipher.id` and
- `cipher.name`

crates/bitwarden-fido/Cargo.toml
```diff
@@ -27,6 +27,8 @@ chrono = { version = ">=0.4.26, <0.5", features = [
     "serde",
 ], default-features = false }
 coset = { version = "0.3.7" }
+schemars = { version = ">=0.8.9, <0.9", features = ["uuid1", "chrono"] }
+itertools = "0.13.0"
```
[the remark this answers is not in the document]
coroiu: arggg xD, will fix!

crates/bitwarden-uniffi/src/platform/fido2.rs
dani-garcia: Ah okay I didn't know that was part of the spec, I though it was just a utility function for our clients. That makes sense then!

crates/bitwarden-fido/src/types.rs
```diff
@@ -1,8 +1,84 @@
...
+                    Ok(Fido2CredentialAutofillView {
+                        credential_id: string_to_guid_bytes(&c.credential_id)?,
+                        cipher_id: cipher
+                            .id
+                            .expect("cipher must be saved to server for autofill"),
```
dani-garcia, line 68: Seeing as this function is already returning their own error type, I think it would be better to return an error here instead of using `expect` and causing a panic if any ciphers don't have ID.

coroiu: Fixed!

crates/bitwarden-fido/Cargo.toml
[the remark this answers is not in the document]
coroiu: Fixed!

dani-garcia (approved):

PR #891, opened 2024-07-09
Use oslogger for macOS logging

coroiu:
## 📔 Objective

Xcode doesn't seem to be able to pick up `stdio` logs from secondary processes like autofill extensions. This PR switches to using native iOS logging which Xcode supports.

## 📸 Screenshots

image

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden-uniffi/src/lib.rs
+++ b/crates/bitwarden-uniffi/src/lib.rs
@@ -75,9 +75,14 @@ impl Client {
 }
 
 fn init_logger() {
-    #[cfg(not(target_os = "android"))]
+    #[cfg(not(any(target_os = "android", target_os = "ios")))]
     let _ = env_logger::Builder::from_env(env_logger::Env::default().default_filter_or("info"))
         .try_init();
+
+    #[cfg(target_os = "ios")]
+    let _ = oslog::OsLogger::new("com.8bit.bitwarden")
+        .level_filter(log::LevelFilter::Info)
+        .init();
 
     #[cfg(target_os = "android")]
     android_logger::init_once(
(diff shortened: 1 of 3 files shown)


crates/bitwarden-uniffi/src/lib.rs
```diff
@@ -75,10 +75,15 @@ impl Client {
 }
 
 fn init_logger() {
-    #[cfg(not(target_os = "android"))]
+    #[cfg(not(any(target_os = "android", target_os = "ios")))]
     let _ = env_logger::Builder::from_env(env_logger::Env::default().default_filter_or("info"))
         .try_init();
 
+    #[cfg(target_os = "ios")]
+    let _ = oslog::OsLogger::new("com.8bit.bitwarden")
```
dani-garcia, line 83: :question: Does this need to match the app identifier? Would it be more helpful to use a separate one to differentiate between app logs and SDK logs?

dani-garcia (approved):

crates/bitwarden-uniffi/src/lib.rs
coroiu: Oh I have no idea, good question. I don't think it matters much. Here is the difference when outputing all metadata

image

Do you have any good suggestions?

dani-garcia: Hmm I guess it doesn't matter much seeing those logs, the error is already printing `bitwarden_fido::crypto`, which I think is good enough to identify SDK calls.

We can leave it as is then.

PR #892, opened 2024-07-09
Fix unreadable passkeys

coroiu:
## 📔 Objective

This PR contains 3 bug fixes

- The PKCS8 key decode function wasn’t actually trying to read the key as PKCS8 
- The SDK decrypted `key_value` and `user_handle` straight to an array buffer, but these fields are actually first encoded as b64 and then encrypted
- The Attestation result tried to use `attestation_data.credential_id`  to return a `credential_id`, but attestation data is not present during an assertion

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-fido/src/authenticator.rs
```diff
@@ -229,12 +229,8 @@ impl<'a> Fido2Authenticator<'a> {
...
 
         let authenticator_data = response.auth_data.to_vec();
-        let credential_id = response
-            .auth_data
-            .attested_credential_data
-            .ok_or(GetAssertionError::MissingAttestedCredentialData)?
-            .credential_id()
-            .to_vec();
+        let credential_id =
+            string_to_guid_bytes(&self.get_selected_credential()?.credential.credential_id)?;
```
dani-garcia, line 233: We should probably extract the call to `get_selected_credential()` to a separate variable, so we don't have to call it twice (once here and another when setting `selected_credential` on line 244)

crates/bitwarden-fido/src/lib.rs
```diff
@@ -102,14 +105,21 @@ impl TryFrom<CipherViewContainer> for Passkey {
 fn try_from_credential_full_view(value: Fido2CredentialFullView) -> Result<Passkey, Fido2Error> {
     let counter: u32 = value.counter.parse().expect("Invalid counter");
     let counter = (counter != 0).then_some(counter);
+    log::info!("Converting Credential to Passkey: {:?}", value);
+    let key_value = URL_SAFE_NO_PAD.decode(value.key_value)?;
+    let user_handle = value
+        .user_handle
+        .map(|u| URL_SAFE_NO_PAD.decode(u))
+        .transpose()?;
+    log::info!("Key value: {:?}", key_value);
```
dani-garcia, line 114: I think we should be careful about logging key values, or at least use `log::debug!` or `log::trace!`, which should not print on release builds by default.

coroiu: Oooooooooooff I completely missed this, it was a debug log and I definitely did not intend on this being logged. Thanks for the save!

crates/bitwarden-fido/src/authenticator.rs
coroiu: Fixed!

crates/bitwarden-fido/src/lib.rs
coroiu: Removed!

dani-garcia (approved):

PR #900, opened 2024-07-12
Return a more detailed MissingKey error from get_key()

dani-garcia:
## 📔 Objective

This change adds the organization UUID to the `MissingKey` error type, and changes the `get_key` function to return a Result instead of an Option to avoid having to add the error all over the place.

Note that some of the `encryption_settings.get_key()` calls were being mapped to `VaultLocked`, that wasn't correct as having the `EncryptionSettings` means the vault is unlocked. Those were changed to `MissingKey(id)` as well.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden-crypto/src/error.rs
+++ b/crates/bitwarden-crypto/src/error.rs
@@ -1,6 +1,7 @@
 use std::fmt::Debug;
 
 use thiserror::Error;
+use uuid::Uuid;
 
 use crate::fingerprint::FingerprintError;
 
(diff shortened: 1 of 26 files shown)


crates/bitwarden-crypto/src/error.rs
```diff
@@ -16,8 +17,8 @@ pub enum CryptoError {
     InvalidKeyLen,
     #[error("The value is not a valid UTF8 String")]
     InvalidUtf8String,
-    #[error("Missing Key")]
-    MissingKey,
+    #[error("Missing Key for organization with ID {0}")]
+    MissingKey(Uuid),
```
Hinton, line 21: Would it be more appropriate to have `MissingKeyUser`, `MissingKeyOrg(id)` as variants of a MissingKey enum?

dani-garcia: A user key will never be missing, as `EncryptionSettings` always has it defined. Only case where a user key is not present is when `EncryptionSettings` is `None`, and in those cases we return `VaultLocked`. 

`enc.get_key(&None)` always returns `Ok` with the user key.

We could rename `MissingKey` to `MissingOrgKey` to be more specific though.

Hinton (approved):

mzieniukbw (approved):
LGTM

PR #910, opened 2024-07-17
[PM-6271] Propose cipher versioning scheme dynamic data

coroiu:
## 📔 Objective

This is a version of the proposal that uses a "dynamic data object" i.e. the cipher data object is not statically typed. Migrators instead rely on runtime checks to perform their work.

Of course, they don't have to use dynamic objects. A migrator is just a function and it can be implemented however you like

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


Cargo.toml
```diff
@@ -28,6 +28,7 @@ bitwarden-fido = { path = "crates/bitwarden-fido", version = "=0.5.0" }
 bitwarden-generators = { path = "crates/bitwarden-generators", version = "=0.5.0" }
 bitwarden-send = { path = "crates/bitwarden-send", version = "=0.5.0" }
 bitwarden-vault = { path = "crates/bitwarden-vault", version = "=0.5.0" }
+bitwarden-versioning = { path = "crates/bitwarden-versioning", version = "=0.5.0" }
```
Hinton, line 31: **thought:** A `versioning` crate feels a bit odd. I think it's fine to put this in the core crate.

Hinton (commented):
I've unfortunately only taken a quick look at this, and it deserves significantly more time to wrap my mind around everything.

There are quite a few things to like here, but also some concerns.

I noticed password_history is not in `data` currently. And data contains items that are currently not always encrypted. I'm worried we will get into scenarios where we need to migrate fields outside the data blob.

It also feels like we're doing much of the work for the cipher refactor in the client without doing the slightly additional work of updating the server too.

I'm curious what items if any we access in the `Cipher` in a decrypted state. Would it be viable to do this:

```
struct Cipher {
  id: UUID,
  meta_data: UnreadableBlob,
  full_data: UnreadableBlob,
  version: 1
}
```

Migrating requires the whole context of `Cipher`, and would affect content in the data blobs, and the version field.

`meta_data` and `full_data` in this struct can't be parsed. Migration is implicitly done while decrypting, or through an explicit migrate step.

```
struct CipherViewFull {
  id: UUID,
  data: {
    login: {
      ..
    },
    edit: true
  }
}
```

coroiu:
@Hinton

> I noticed password_history is not in data currently. 

I forgot to add `password_history` to the `data` object and then I just never got around to it because of how fast I was trying to get this out. In other words, it can definitely move into `Data`

> And data contains items that are currently not always encrypted.

What I tried to do was to think of `Data` as a structure defined entirely by the client, while everything outside of `Data` is defined by the server. We can definitely move those fields into another object called `Metadata` if we want to make that distinction clearer.

> I'm worried we will get into scenarios where we need to migrate fields outside the data blob.

Given this distinction, migrating fields outside of the data blob should never be a thing. At least not on the client. Those fields are owned by the server and should be handled appropriately.

> It also feels like we're doing much of the work for the cipher refactor in the client without doing the slightly additional work of updating the server too.

Not sure I follow, but I have skipped the server at this stage because I think those changes are much more straight forward. In other words I didn't think we needed to prove the concept on the server side.

Cargo.toml
coroiu: Yeah I agree, I thought this was gonna be much bigger but alas :)

PR #912, opened 2024-07-18
[PM-9527] Add PIN validation support

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-9527

## 📔 Objective

Fede mentioned on Slack some time ago that we currently don't have a simple way to validate the user's PIN like we do with the password. Instead the way to do it was to call `init_crypto` and check that the result is not an error.

This requires more data than needed and does a lot of unnecessary operations, so I think it makes sense to expose a simpler way to do PIN verification.

In this PR we're decrypting the PIN protected user key and comparing it with the stored user key, similar to what we're doing with `validate_password_user_key`.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden-core/src/auth/pin.rs
+++ b/crates/bitwarden-core/src/auth/pin.rs
@@ -16,6 +16,7 @@ pub(crate) fn validate_pin(
         .get_login_method()
         .ok_or(Error::NotAuthenticated)?;
 
+    #[allow(irrefutable_let_patterns)]
     let LoginMethod::User(login_method) = login_method.as_ref() else {
         return Err(Error::NotAuthenticated);
     };
--- a/crates/bitwarden-core/src/auth/pin.rs
+++ b/crates/bitwarden-core/src/auth/pin.rs
@@ -1,0 +1,100 @@
+use bitwarden_crypto::{EncString, PinKey};
+
+use crate::{
+    client::{LoginMethod, UserLoginMethod},
+    error::{Error, Result},
+    Client,
+};
+
+pub(crate) fn validate_pin(
+    client: &Client,
+    pin: String,
+    pin_protected_user_key: EncString,
+) -> Result<bool> {
+    let login_method = client
+        .internal
+        .get_login_method()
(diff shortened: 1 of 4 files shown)


crates/bitwarden-core/src/auth/pin.rs
```diff
@@ -0,0 +1,101 @@
...
+        .ok_or(Error::NotAuthenticated)?;
+
+    #[allow(irrefutable_let_patterns)]
+    let LoginMethod::User(login_method) = login_method.as_ref() else {
+        return Err(Error::NotAuthenticated);
+    };
+
+    match login_method {
+        UserLoginMethod::Username { email, kdf, .. }
+        | UserLoginMethod::ApiKey { email, kdf, .. } => {
```
Hinton, line 26: **question:** Do we expect Api key logins to use pin?

crates/bitwarden-uniffi/src/auth/mod.rs
```diff
@@ -110,6 +110,17 @@ impl ClientAuth {
             .validate_password_user_key(password, encrypted_user_key)?)
     }
 
+    /// Validate the user PIN
+    ///
+    /// To validate the user PIN, you need to have the user's pin_protected_user_key. This key is
+    /// obtained when enabling PIN unlock on the account with the `derive_pin_key` method.
+    ///
+    /// This works by the decrypted user key with the current user key, so the client must be
```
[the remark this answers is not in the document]
dani-garcia: A good eye, I did miss a word there, fixed now in ab5427f

crates/bitwarden-core/src/auth/pin.rs
dani-garcia: I don't think so, and I'm not sure if they would need password validation either. I think this is mostly a case of copying over the `match` to get `email` and `kdf`. We can return an error for `ApiKey` if you prefer though.

Hinton (approved):

PR #676, opened 2024-03-21
SM-1147: Switch to try_init with pyo3_log

coltonhurst:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Tech debt (refactoring, code cleanup, dependency upgrades, etc)
- [ ] Build/deploy pipeline (DevOps)
- [ ] Other

## Objective

When running the `ansible-playbook` command, we needed to export the following variable: `export OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES`. This suppresses fork safety warnings on Mac OS. With this set, you can safely query for individual secrets. However, if you query for secrets within an Ansible loop or with a Jinja2 template, we receive an error that is caused by setting up a logger more than once. (This error comes from the `pyo3_log` crate.)

If we switch to using the `pyo3_log::try_init` function, we ignore the case where a logger is already set up if the attempt fails. This happens because the `pyo3_log::init` function [can panic](https://github.com/vorner/pyo3-log/blob/70819388bfcb58a7ca91a27179a73cea0abde31e/src/lib.rs#L569). This also introduces panic safety.

## Code changes

- **client.rs:** Switch to `try_init`

## Before you submit

- Please add **unit tests** where it makes sense to do so

PR #935, opened 2024-08-02
[SM-1384] Fix panic on re-registering logger | WASM

Thomas-Avery:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/SM-1384

## 📔 Objective

When creating multiple WASM clients, a panic occurs `failed to initialize logger: SetLoggerError())`.

Looks to be the same thing we fixed: 
https://github.com/bitwarden/sdk/pull/181
https://github.com/bitwarden/sdk/pull/676

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


dani-garcia (approved):

crates/bitwarden-wasm/src/client.rs
```diff
@@ -37,11 +37,9 @@ impl BitwardenClient {
...
     pub fn new(settings_input: Option<String>, log_level: Option<LogLevel>) -> Self {
         console_error_panic_hook::set_once();
-        if let Err(e) =
-            console_log::init_with_level(convert_level(log_level.unwrap_or(LogLevel::Info)))
-        {
-            panic!("failed to initialize logger: {:?}", e);
-        }
+        // This will only fail if another logger was already initialized, so we can ignore the
+        // result.
+        let _ = console_log::init_with_level(convert_level(log_level.unwrap_or(LogLevel::Info)));
```
mzieniukbw, line 42: I wonder if maybe it would be better to log warn when this happens, instead of ignoring it, since it potentially reduces a chance for a customer raising an issue about log level not being set.

Thomas-Avery: Great question, I will look into this more.

Thomas-Avery: Findings: 
  The first log level set with the first client is honored, all subsequent client creation logging levels are ignored.
  
  I added a warning as suggested here. https://github.com/bitwarden/sdk/pull/935/commits/2da01df2393596b7205c708ad36905ab6f0d4f33

mzieniukbw (dismissed):

crates/bitwarden-wasm/src/client.rs
```diff
@@ -37,10 +37,10 @@ impl BitwardenClient {
     #[wasm_bindgen(constructor)]
     pub fn new(settings_input: Option<String>, log_level: Option<LogLevel>) -> Self {
         console_error_panic_hook::set_once();
-        if let Err(e) =
+        if let Err(_e) =
             console_log::init_with_level(convert_level(log_level.unwrap_or(LogLevel::Info)))
         {
-            panic!("failed to initialize logger: {:?}", e);
+            warn!("Logger already initialized, log level will not be changed.");
```
Hinton, line 43: It's probably going to be confusing to get a warning when you setup multiple sdk clients. If we do want to warn about this we should check if the log level is different.

Thomas-Avery: Agreed, at the time of looking at the `console_log` crate, I didn't think it was possible. 

After further research, I found out we can use the log crate functions, and we can just set the level. 

Thoughts on this approach?

From an SDK consumer perspective, this seems ideal.

https://docs.rs/log/0.4.22/log/fn.set_max_level.html

Edit here https://github.com/bitwarden/sdk/pull/935/commits/3a86e8e51083021fd593b0d28ba34dd5d819fed7

Hinton (approved):
Seems reasonable. Do we need to update the other bindings with similar behaviour?

dani-garcia (approved):

coltonhurst:
As discussed with the SM team, merging this. QA will handle testing on `main` prior to the future release.

PR #942, opened 2024-08-07
[SM-1399] Update the README for Windows on ARM

coltonhurst:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/SM-1399

## 📔 Objective

Just `cargo build` is not sufficient for building the SDK on Windows on ARM. We should add clarity to what is required so that `cargo build` will run successfully.

Python is required. Clang is also required for [ring](https://crates.io/crates/ring) (more info [here](https://github.com/briansmith/ring/blob/main/BUILDING.md)).

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,8 @@ To build, you will need the following in your PATH:
 
 - [Python](https://www.python.org)
 - [Clang](https://clang.llvm.org)
-  - We recommend installing this via [Visual Studio](https://learn.microsoft.com/en-us/cpp/build/clang-support-msbuild?view=msvc-170#install-1)
+  - We recommend installing this via
+    [Visual Studio](https://learn.microsoft.com/en-us/cpp/build/clang-support-msbuild?view=msvc-170#install-1)
 
 ## Crates
 
--- a/README.md
+++ b/README.md
@@ -20,9 +20,19 @@ are currently open as well as what it's like to work at Bitwarden.
 
 ## Getting Started
 
+### Linux / Mac / Windows
+
 ```bash
 cargo build
 ```
+
+### Windows on ARM
+
+To build, you will need the following in your PATH:
+
+- [Python](https://www.python.org)
+- [Clang](https://clang.llvm.org)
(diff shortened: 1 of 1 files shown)


dani-garcia (commented):
I didn't see the Python requirement in the ring docs, but yeah the clang requirement is true.

Rather than full visual studio I think we can recommend installing the [Visual Studio Build Tools](https://visualstudio.microsoft.com/downloads/#build-tools-for-visual-studio-2022) which should be somewhat lighter than a full visual studio install. You still need to manually select the `clang` component and add it to the PATH though.

coltonhurst:
> I didn't see the Python requirement in the ring docs, but yeah the clang requirement is true.
> 
> Rather than full visual studio I think we can recommend installing the [Visual Studio Build Tools](https://visualstudio.microsoft.com/downloads/#build-tools-for-visual-studio-2022) which should be somewhat lighter than a full visual studio install. You still need to manually select the `clang` component and add it to the PATH though.

Do we know if those build tools include Clang? I attempted this the first time around in my tests and I couldn't find Clang in the filesystem until after I installed the full suite.

I would much rather only recommend Build Tools because it is much smaller as you mentioned.

dani-garcia:
I think for both the build tools and the full visual studio package you have to manually select `clang` from the optional packages list inside the `Visual Studio Installer`, but I haven't checked.

dani-garcia:
I think that's the wrong download, that is the runtime redist and not the build tools, this is what i get with the build tools:
Screenshot 2024-08-07 at 18 46 36
Screenshot 2024-08-07 at 18 50 04

coltonhurst:
> I think that's the wrong download, that is the runtime redist and not the build tools, this is what i get with the build tools: [image]

Ah, I did have the wrong download. Thanks @dani-garcia!

Updated in https://github.com/bitwarden/sdk/pull/942/commits/941fbaea845162e1eb8b6242bf2a2d0a594dbb85

dani-garcia (approved):

PR #979, opened 2024-08-19
[PM-5693] CryptoService using memfd_secret on Linux

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-5693

## 📔 Objective

I've been looking into using `memfd_secret` to protect keys in memory on Linux.

The `memfd_secret` API is similar to malloc in that we get some memory range allocated for us, but this memory is only visible to the process that has access to the file descriptor, which should protect us from any memory dumping from anything other than a kernel driver.

https://man.archlinux.org/man/memfd_secret.2.en

Using this requires changing how we store keys, so this is the perfect moment to go through removing the raw keys from the API surface.

## Changes
- Added a `KeyRef` trait and `SymmetricKeyRef`/`AsymmetricKeyRef` subtraits to represent the key types. Also a small macro to quickly implement them. `KeyRef` implementations are user defined types that implements `Eq+Hash+Copy`, and don't contain any key material
- `KeyEncryptable/KeyDecryptable` are now `Decryptable/Encryptable`, and instead of taking the key as parameter, they take a `CryptoServiceContext` and a `KeyRef`. This way keys are never exposed to the clients.
- `KeyLocator` trait is replaced by `UsesKey`. This trait only returns the `KeyRef` of the key required to decrypt it, instead of fetching the key itself.
- Created a `KeyStore` trait, with some simple CRUD methods. We have two implementations, one based on `memfd_secret` for Linux, and another one based on a Rust boxed slice, that also applies `mlock` if possible.
-  Because both `mlock` and `memfd_secret` apply their protection over a specified memory area, we need a Rust data structure that is laid out linearly and also doesn't reallocate by itself. Also because `memfd_secret` allocates the memory for us, we can't use a std type like Vec (reason is the Vec must be allocated by the Rust allocator [[ref](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.from_raw_parts)]). This basically only leaves us with a Rust slice, on top of which we'd need to implement insert/get/delete. To allow for reuse and easier testing I've created `SliceKeyStore`, which implements the CRUD methods from the `KeyStore` trait on top of a plain slice. Then the actual `mlock` and `memfd_secret` implementations just need to implement a trait casting their data to a slice. The data is stored as a slice of `Option<(KeyRef, KeyMaterial)>`, and the keys are unique and sorted in the slice for easier lookups.
- Created `CryptoService`, which contains the `KeyStore`s and some encrypt/decrypt utility functions. From a `CryptoService` you can also initialize a `CryptoServiceContext`
- A `CryptoServiceContext` contains a read only view of the keys inside the `CryptoService`, plus a read-write ephemeral key store, for use by `Decryptable/Encryptable` implementations when they need to temporarily store some keys (Cipher keys, attachment keys, send keys...).

Migrated the codebase to use these changes in a separate PR: https://github.com/bitwarden/sdk/pull/1117

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-crypto/Cargo.toml
```diff
@@ -46,8 +46,13 @@ uniffi = { version = "=0.28.1", optional = true }
 uuid = { version = ">=1.3.3, <2.0", features = ["serde"] }
 zeroize = { version = ">=1.7.0, <2.0", features = ["derive", "aarch64"] }
 
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+memsec = { version = "0.7.0", features = [
+    "alloc_ext",
+], git = "https://github.com/dani-garcia/memsec", rev = "3d2e272d284442637bac0a7d94f76883960db7e2" }
```
dani-garcia, line 52: `memsec` is using `windows-sys` 0.45, which was causing build errors similar to the ones fixed by #1053.

Using the latest `windows-sys` 0.59 seems to solve the problem for me. TODO: Open a PR upstream?

crates/bitwarden-crypto/src/keys/key_ref.rs
```diff
@@ -0,0 +1,89 @@
...
+    /// the system, an end user would not implement this trait directly, and would instead use
+    /// the `SymmetricKeyRef` and `AsymmetricKeyRef` traits.
+    pub trait KeyRef:
+        Debug + Clone + Copy + Hash + Eq + PartialEq + Ord + PartialOrd + Send + Sync + 'static
+    {
+        type KeyValue: CryptoKey + Send + Sync + ZeroizeOnDrop;
+
+        /// Returns whether the key is local to the current context or shared globally by the
+        /// service.
+        fn is_local(&self) -> bool;
```
coroiu, line 24: question: what does local mean?

coroiu: Oh ok I just found it 

https://github.com/bitwarden/sdk/blob/0c74d80e314f9aeaccd840fc6c9c0f31474268f2/crates/bitwarden-vault/src/cipher/cipher.rs#L175-L176

crates/bitwarden-crypto/src/service/key_store/mod.rs
```diff
@@ -0,0 +1,20 @@
...
+/// This trait represents a platform that can securely store and return keys. The `SliceKeyStore`
+/// implementation is a simple in-memory store with some platform-specific security features. Other
+/// implementations could use secure enclaves or HSMs, or OS provided keychains.
+pub trait KeyStore<Key: KeyRef>: ZeroizeOnDrop + Send + Sync {
+    fn insert(&mut self, key_ref: Key, key: Key::KeyValue);
+    fn get(&self, key_ref: Key) -> Option<&Key::KeyValue>;
+    fn remove(&mut self, key_ref: Key);
+    fn clear(&mut self);
+
+    fn retain(&mut self, f: fn(Key) -> bool);
```
coroiu, line 19: suggestion: add some documentation to these functions describing their behavior and use-cases. Admittedly most are pretty self-explanatory, but e.g. `retain` isn't

PR #993, opened 2024-08-23
Add `run_command_async`

justindbaur:
## 📔 Objective

- Bumps .NET to 8 (latest LTS)
- Uses `[LibraryImport]` for auto-marshalling of function
- Adds `run_command_async` method in `bitwarden-c` that allows for a non-blocking call that can notify the caller when the action is completed.
- `Task`-ifies one of the C# SDK calls into the SDK. 
  - This only updates one method to be async for now but all of them can be changed
  - This is a breaking change so would be a good idea before a `1.0.0` release.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden-c/src/c.rs
+++ b/crates/bitwarden-c/src/c.rs
@@ -28,6 +28,32 @@ pub extern "C" fn run_command(c_str_ptr: *const c_char, client_ptr: *const CClie
     }
 }
 
+type OnCompletedCallback = unsafe extern "C" fn(result: *mut c_char) -> ();
+
+#[no_mangle]
+pub extern "C" fn run_command_async(c_str_ptr: *const c_char, client_ptr: *const CClient, on_completed_callback: OnCompletedCallback) -> () {
+    let client = unsafe { ffi_ref!(client_ptr) };
+    let input_str = str::from_utf8(unsafe { CStr::from_ptr(c_str_ptr) }.to_bytes())
+        .expect("Input should be a valid string")
+        // Languages may assume that the string is collectable as soon as this method exits
+        // but it's not since the request will be run in the background
+        // so we need to make our own copy.
+        .to_owned();
+
+    client
+        .runtime
+        .spawn(async move {
+            let result = client.client.run_command(input_str.as_str()).await;
+            let str_result = match std::ffi::CString::new(result) {
+                Ok(cstr) => cstr.into_raw(),
+                Err(_) => panic!("failed to return comment result: null encountered"),
+            };
+
+            // run completed function
+            unsafe { on_completed_callback(str_result) }
+        });
+}
+
 // Init client, potential leak! You need to call free_mem after this!
 #[no_mangle]
 pub extern "C" fn init(c_str_ptr: *const c_char) -> *mut CClient {
--- a/languages/csharp/Bitwarden.Sdk/Bitwarden.Sdk.csproj
+++ b/languages/csharp/Bitwarden.Sdk/Bitwarden.Sdk.csproj
@@ -1,11 +1,11 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net8.0</TargetFramework>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
     <RootNamespace>Bitwarden.Sdk</RootNamespace>
-
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <Title>Bitwarden Secrets Manager SDK</Title>
     <Authors>Bitwarden Inc.</Authors>
     <Description>.NET bindings for interacting with the Bitwarden Secrets Manager</Description>
--- a/languages/csharp/Bitwarden.Sdk/BitwardenLibrary.cs
+++ b/languages/csharp/Bitwarden.Sdk/BitwardenLibrary.cs
@@ -30,18 +30,16 @@ internal static partial class BitwardenLibrary
 
         try
         {
-            run_command_async(json, handle, (pointer) =>
+            run_command_async(json, handle, (resultPointer) =>
             {
                 try
                 {
-                    var stringResult = Marshal.PtrToStringUTF8(pointer);
-                    Console.WriteLine($"Setting Result {stringResult}");
+                    var stringResult = Marshal.PtrToStringUTF8(resultPointer);
                     tcs.SetResult(stringResult);
-                    Console.WriteLine("Set Result");
                 }
                 finally
                 {
-                    Marshal.FreeCoTaskMem(pointer);
+                    Marshal.FreeCoTaskMem(resultPointer);
                 }
             });
         }
--- a/languages/csharp/Bitwarden.Sdk/BitwardenLibrary.cs
+++ b/languages/csharp/Bitwarden.Sdk/BitwardenLibrary.cs
@@ -2,20 +2,56 @@
 
 namespace Bitwarden.Sdk;
 
-internal static class BitwardenLibrary
+internal static partial class BitwardenLibrary
 {
-    [DllImport("bitwarden_c", CallingConvention = CallingConvention.Cdecl)]
-    private static extern BitwardenSafeHandle init(string settings);
+    [LibraryImport("bitwarden_c", StringMarshalling = StringMarshalling.Utf8)]
+    private static partial BitwardenSafeHandle init(string settings);
 
-    [DllImport("bitwarden_c", CallingConvention = CallingConvention.Cdecl)]
-    private static extern void free_mem(IntPtr handle);
+    [LibraryImport("bitwarden_c", StringMarshalling = StringMarshalling.Utf8)]
+    private static partial void free_mem(IntPtr handle);
 
-    [DllImport("bitwarden_c", CallingConvention = CallingConvention.Cdecl)]
-    private static extern string run_command(string json, BitwardenSafeHandle handle);
+    [LibraryImport("bitwarden_c", StringMarshalling = StringMarshalling.Utf8)]
+    private static partial string run_command(string json, BitwardenSafeHandle handle);
+
+    internal delegate void OnCompleteCallback(IntPtr json);
+
+    [LibraryImport("bitwarden_c", StringMarshalling = StringMarshalling.Utf8)]
+    private static partial void run_command_async(string json, BitwardenSafeHandle handle, OnCompleteCallback on_completed_callback);
 
     internal static BitwardenSafeHandle Init(string settings) => init(settings);
 
     internal static void FreeMemory(IntPtr handle) => free_mem(handle);
 
     internal static string RunCommand(string json, BitwardenSafeHandle handle) => run_command(json, handle);
+
+    internal static Task<string> RunCommandAsync(string json, BitwardenSafeHandle handle, CancellationToken token = default)
+    {
+        var tcs = new TaskCompletionSource<string>(TaskCreationOptions.RunContinuationsAsynchronously);
+
+        try
+        {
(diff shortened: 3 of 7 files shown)


languages/csharp/Bitwarden.Sdk/Bitwarden.Sdk.csproj
```diff
@@ -1,11 +1,11 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net6.0</TargetFramework>
```
dani-garcia, line 4: Question: Does increasing this affect people using the SDK in older .NET versions?

languages/csharp/Bitwarden.Sdk/BitwardenLibrary.cs
```diff
@@ -2,20 +2,54 @@
...
+            run_command_async(json, handle, (resultPointer) =>
+            {
+                try
+                {
+                    var stringResult = Marshal.PtrToStringUTF8(resultPointer);
+                    tcs.SetResult(stringResult);
+                }
+                finally
+                {
+                    Marshal.FreeCoTaskMem(resultPointer);
```
dani-garcia, line 42: Is this freeing the Rust created result string? I don't think that's allowed according to the Rust docs, we'd need to free it from the Rust side:

https://doc.rust-lang.org/std/ffi/struct.CString.html#method.into_raw

Instead we should be freeing it from Rust after the callback was executed

dani-garcia (commented):
Overall LGTM, just some small details.

languages/csharp/Bitwarden.Sdk/Bitwarden.Sdk.csproj
justindbaur: Yes :/ .NET 6 does go out of support in November though. My current thinking is that I will try and convert more of the commands to be Async and then maybe we do a `1.0.0` release after .NET 6 goes out of support with the breaking changes of converting them.

languages/csharp/Bitwarden.Sdk/BitwardenLibrary.cs
justindbaur: Good call, moved the freeing to rust.

languages/csharp/Bitwarden.Sdk/Bitwarden.Sdk.csproj
```diff
@@ -1,11 +1,11 @@
...
+    <TargetFramework>net8.0</TargetFramework>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
     <RootNamespace>Bitwarden.Sdk</RootNamespace>
-
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
Hinton, line 8: Is anything using unsafe on the c# side?

languages/csharp/Bitwarden.Sdk/Bitwarden.Sdk.csproj
Hinton: Do we need to upgrade target framework? We've had requests to use NET Standard 2.0 to support .NET Framework.

languages/csharp/Bitwarden.Sdk/BitwardenLibrary.cs
```diff
@@ -2,20 +2,47 @@
...
+                var stringResult = Marshal.PtrToStringUTF8(resultPointer);
+                tcs.SetResult(stringResult);
+            });
+        }
+        catch (Exception ex)
+        {
+            tcs.SetException(ex);
+        }
+
+        // TODO: Register cancellation on token
```
Hinton, line 44: todo?

languages/csharp/Bitwarden.Sdk/Bitwarden.Sdk.csproj
justindbaur: Yes, `[LibraryImport] does, it generates code like this that has unsafe blocks:

```csharp
private static partial void run_command_async(string json, global::Bitwarden.Sdk.BitwardenSafeHandle handle, global::Bitwarden.Sdk.BitwardenLibrary.OnCompleteCallback on_completed_callback)
{
    byte* __json_native = default;
    nint __handle_native = default;
    System.IntPtr __on_completed_callback_native = default;
    // Setup - Perform required setup.
    global::System.Runtime.InteropServices.Marshalling.SafeHandleMarshaller<global::Bitwarden.Sdk.BitwardenSafeHandle>.ManagedToUnmanagedIn __handle_native__marshaller = new();
    scoped global::System.Runtime.InteropServices.Marshalling.Utf8StringMarshaller.ManagedToUnmanagedIn __json_native__marshaller = new();
    try
    {
        // Marshal - Convert managed data to native data.
        __on_completed_callback_native = on_completed_callback != null ? global::System.Runtime.InteropServices.Marshal.GetFunctionPointerForDelegate(on_completed_callback) : default;
        __handle_native__marshaller.FromManaged(handle);
        __json_native__marshaller.FromManaged(json, stackalloc byte[global::System.Runtime.InteropServices.Marshalling.Utf8StringMarshaller.ManagedToUnmanagedIn.BufferSize]);
        {
            // PinnedMarshal - Convert managed data to native data that requires the managed data to be pinned.
            __handle_native = __handle_native__marshaller.ToUnmanaged();
            __json_native = __json_native__marshaller.ToUnmanaged();
            __PInvoke(__json_native, __handle_native, __on_completed_callback_native);
        }

        // NotifyForSuccessfulInvoke - Keep alive any managed objects that need to stay alive across the call.
        global::System.GC.KeepAlive(on_completed_callback);
    }
    finally
    {
        // CleanupCallerAllocated - Perform cleanup of caller allocated resources.
        __handle_native__marshaller.Free();
        __json_native__marshaller.Free();
    }

    // Local P/Invoke
    [global::System.Runtime.InteropServices.DllImportAttribute("bitwarden_c", EntryPoint = "run_command_async", ExactSpelling = true)]
    static extern unsafe void __PInvoke(byte* __json_native, nint __handle_native, System.IntPtr __on_completed_callback_native);
}
```

languages/csharp/Bitwarden.Sdk/BitwardenLibrary.cs
justindbaur: It's still something I want to try out but I think it can be a future enhancement. 

It would require boxing the pointer to the returned `JoinHandle` on the rust side. Sending it back to C# to that we can register a cancellation action and then we likely have to add two new methods on the rust side. Probably `abort_and_drop_handle` and `drop_handle`. Do you have any feelings about that?

languages/csharp/Bitwarden.Sdk/Bitwarden.Sdk.csproj
justindbaur: `[LibraryImport]` requires .NET 7+, in my opinion supporting .NET Framework would not be worth it. C# may soon be getting built in `Result` and `Option` types that we could take advantage of here and getting locked into .NET Framework support would either mean a of conditional compilation or just holding back the .NET targets.

crates/bitwarden-core/Cargo.toml
```diff
@@ -54,6 +54,9 @@ validator = { version = "0.18.1", features = ["derive"] }
 zeroize = { version = ">=1.7.0, <2.0", features = ["derive", "aarch64"] }
 zxcvbn = { version = ">=3.0.1, <4.0", optional = true }
 
+[target.'cfg(debug_assertions)'.dependencies]
+tokio = { version = "1.36.0", features = ["rt", "macros", "time"] }
```
justindbaur, line 58: I barely know what I did here, I need the time feature but only in debug, is this the right way to go?

justindbaur:
I added support for `CancellationToken` which will cancel an ongoing call to rust. To ensure I did this correctly I added some debug only commands. I also converted all the C# methods to be `Task` returning and to optionally take a `CancellationToken`.

languages/csharp/Bitwarden.Sdk/BitwardenLibrary.cs
justindbaur: I ended up doing this, let me know what you think about the implementation.

crates/bitwarden-c/src/c.rs
```diff
@@ -28,6 +33,48 @@ pub extern "C" fn run_command(c_str_ptr: *const c_char, client_ptr: *const CClie
...
+        unsafe {
+            on_completed_callback(str_result);
+            let _ = CString::from_raw(str_result);
+        }
+    });
+
+    // We only want to box the join handle the caller
+    // has said that they may want to cancel, essentially
+    // promising to us that they will take care of the
+    // returned pointer.
```
Hinton, line 70: This seems fairly compact, `cargo fmt` generally uses 100 as the rule for line breaks.

```suggestion
    // We only want to box the join handle the caller has said that they may want to cancel,
    // essentially promising to us that they will take care of the returned pointer.
```

crates/bitwarden-core/Cargo.toml
Hinton: It is, but I can't say I'm a fan of different behaviors on debug and release builds.

Hinton:
I added SM as reviewers, we should try and get this in before the 1.0 release of the C# sdk.

crates/bitwarden-core/Cargo.toml
justindbaur: This feels like the exact thing a debug build is for, it helps you write tests to make sure you've built the language wrapper correctly. That said I was able to move it into the `bitwarden-json` crate instead of `core`.

crates/bitwarden-core/src/platform/client_platform.rs
```diff
@@ -24,6 +24,28 @@ impl<'a> ClientPlatform<'a> {
...
+    }
+
+    #[cfg(debug_assertions)]
+    pub fn error_test(&mut self) -> Result<i32> {
+        use crate::Error;
+
+        Err(Error::Internal(std::borrow::Cow::Borrowed(
+            "This is an error.",
+        )))
+    }
```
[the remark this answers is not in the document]
justindbaur: I moved them to `bitwarden-json`, if I moved it into `bitwarden-c` I would have to double parse the input string to check for special commands. Do you think that is alright?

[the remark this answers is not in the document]
Hinton: Why not just expose two regular functions for it?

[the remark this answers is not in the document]
justindbaur: The cancellation test is mostly testing the very code written in `run_command_async` so we'd have to expose a method with very similar internals to that and repeat ourselves. The error test is testing that the C# wrapper properly handles how errors are given back to it through the bitwarden-json crate.

crates/bitwarden-json/src/command.rs
```diff
@@ -188,3 +190,11 @@ pub enum GeneratorsCommand {
...
     GeneratePassword(PasswordGeneratorRequest),
 }
+
+#[cfg(debug_assertions)]
+#[derive(Serialize, Deserialize, JsonSchema, Debug)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub enum DebugCommand {
+    CancellationTest { duration_millis: u64 },
+    ErrorTest {},
+}
```
Hinton, line 200: An alternative way. of doing this would be to add one or two separate commands to bitwarden-c.

Hinton (commented):
I think this seems fine from my perspective. I'll let SM do the final review and cover the csharp files.

coltonhurst (approved):
This looks good to me, thanks @justindbaur!

Have you tested that state files still work on the client creation and during token renewal?

tangowithfoxtrot (approved):

PR #1036, opened 2024-09-11
[SM-1402] Automatic PHP schema generation

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/SM-1402
(Not part of this ticket, but agreed with @tangowithfoxtrot that it would make sense to do so)

## 📔 Objective

Automatic PHP Schema generation using quicktype.
The schema is not compatible with previously generated Swaggest based schema - PHP code changes are required (does not affect the usage, `README.md` and `example.php` works as before)
Notable code changes:
- Schema class fields are private, no setters or getters. Can only init the classes via constructor. (Could enable getters in schema generation, but the quicktype does not respond well with nested arrays that are part of schema, as a result PHP generated schema file does not compile)
- Schema class constructors are initialized with named arguments, which is more readable than `new Command(null, null, null, null, null, $secrets_command, null, null)
- Uses PHP native json serialization, but SDK does not react well to null fields, so adjusted the code to remove it just before JSON serialization.
- Separated `AuthClient` into separate file
- Imports with `uses`

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


languages/php/example.php
mzieniukbw, line 1: no changes in this file (except for `$bitwarden_client->auth`), except widening the try-catch across all file

tangowithfoxtrot (approved):
Looks great! Thanks for automating the schema generation!

support/scripts/schemas.ts
```diff
@@ -117,9 +117,30 @@ async function main() {
...
+  });
+
+  const phpDir = "./languages/php/src/Schemas/";
+  if (!fs.existsSync(phpDir)) {
+    fs.mkdirSync(phpDir);
+  }
+
+  php.lines.splice(1, 0, "namespace Bitwarden\\Sdk\\Schemas;", "use stdClass;", "use Exception;");
+
+  writeToFile("./languages/php/src/Schemas/Schemas.php", php.lines);
```
coltonhurst, line 139: Just a nit... isn't the `schemas` directory normally (and currently) lowercase? Any reason we are capitalizing it?

coltonhurst (changes requested):

coltonhurst:
Thanks for adding this by the way, this will be very helpful going forward! 👏

support/scripts/schemas.ts
mzieniukbw: namespaces are case sensitive in PHP. We imported with capitalised name (like most PHP projects to), but the directory were lowercase, which is wrong.
Many big PHP projects just use capitalised name, but i don't think there is any namespace naming convention. I would keep it capitalized as many PHP projects does.

mzieniukbw: Few examples from large projects, all use capitalised names:
https://github.com/laravel/laravel/tree/11.x/app
https://github.com/symfony/symfony/tree/7.2/src/Symfony
https://github.com/composer/composer/tree/main/src (we use this)

coltonhurst: Perfect, thank you!

tangowithfoxtrot (approved):

PR #1116, opened 2024-10-07
[PM-11764] Implement account switching and sdk initialization

coroiu:
## 📔 Objective

- Expose crypto client
- Add error handling

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


.editorconfig
Hinton, line 1: These changes seems unrelated? And we probably want to change quite_type for ts ...

crates/bitwarden-wasm-internal/src/crypto.rs
```diff
@@ -0,0 +1,37 @@
...
+impl ClientCrypto {
+    /// Initialization method for the user crypto. Needs to be called before any other crypto
+    /// operations.
+    pub async fn initialize_user_crypto(&self, req: InitUserCryptoRequest) -> Result<()> {
+        Ok(self
+            .0
+            .crypto()
+            .initialize_user_crypto(req)
+            .await
+            .map_err(bitwarden_core::Error::EncryptionSettings)?)
```
Hinton, line 24: **suggestion:** If we add bitwarden_core error as transparent #from in this crates error, you should be able to avoid the explicit error conversion and rely on the `thiserror` macro from implementation.

This is the pattern I updated uniffi to use in https://github.com/bitwarden/sdk/pull/1124.

crates/bitwarden-crypto/src/enc_string/asymmetric.rs
```diff
@@ -11,11 +11,16 @@ use crate::{
...
-
 // This module is a workaround to avoid deprecated warnings that come from the ZeroizeOnDrop
 // macro expansion
 #[allow(deprecated)]
 mod internal {
+    #[cfg(feature = "wasm")]
+    #[wasm_bindgen::prelude::wasm_bindgen(typescript_custom_section)]
+    const TS_CUSTOM_TYPES: &'static str = r#"
+    export type AsymmetricEncString = string;
+    "#;
```
Hinton, line 22: We have a `uniffi_support.rs` file, would it make sense to add something similar for these custom logics? Or should we encourage lifting in the uniffi logic to where it's needed.

coroiu: I added `custom_types.rs` which is inspired by that uniffi file, but I think adding the types where they are defined is a preferred approach. It's just that we sometimes use external type and so I needed both approaches

.editorconfig
coroiu: Oh that's why I suddenly got lots of prettier errors, lol, yeah ok I'll split this out

crates/bitwarden-wasm-internal/src/crypto.rs
coroiu: I have chosen another path which simply converts all errors to strings. I'll make sure we have a follow-up ticket to investigate the best method for catching errors in JS

coroiu:
Adding @dani-garcia as reviewer because @Hinton is traveling/at conference

crates/bitwarden-core/src/mobile/crypto.rs
```diff
@@ -6,6 +6,8 @@ use bitwarden_crypto::{
 };
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
+#[cfg(feature = "wasm")]
+use {tsify_next::Tsify, wasm_bindgen::prelude::*};
```
dani-garcia, line 10: Huh, I just discovered now thanks to this that you can use a top level `{}` block in a `use` statement, neat!

crates/bitwarden-wasm-internal/src/client.rs
```diff
@@ -27,7 +29,7 @@ fn convert_level(level: LogLevel) -> Level {
 // Rc<...> is to avoid needing to take ownership of the Client during our async run_command
 // function https://github.com/rustwasm/wasm-bindgen/issues/2195#issuecomment-799588401
 #[wasm_bindgen]
-pub struct BitwardenClient(Rc<Client>);
+pub struct BitwardenClient(pub(crate) Rc<Client>);
```
dani-garcia, line 32:
```suggestion
pub struct BitwardenClient(Rc<Client>);
```

Nit: This inner value is only used on it's own impl as far as I know, so we don't need it to be `pub(crate)`

dani-garcia (dismissed):
LGTM!

dani-garcia (approved):

PR #1515, opened 2026-05-14
chore: update sdk internal to 3.0.0

tangowithfoxtrot:
## 📔 Objective

Update to sdk-internal 3.0.0 in preparation for sdk-sm release.


tangowithfoxtrot:
Reverted [85375be](https://github.com/bitwarden/sdk-sm/pull/1515/commits/85375beb63978eb7877dbfc818b1edc754a558d5) because the root-cause of the workflow failures it was meant to address is actually caused by https://github.com/actions/runner-images/issues/14097, and not by the cache. Merely disabling the cache just causes other workflows to fail, so it wasn't a great way to workaround the issue in the first place.

tangowithfoxtrot:
Looks like the macOS Runner jank is gone. The workflows are green again :)

cd-bitwarden (approved):
👍

PR #1517, opened 2026-05-15
Qa 1715/sdk test poc

ddanielyan:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/QA-1715

## 📔 Objective

    - Added SDK test framework POC for Python and Go language bindings
    - Implemented cross-platform test runners (PythonTestRunner.cs, GoTestRunner.cs) with virtual environment support
    - Created unified test suite (test_suite.py, test_suite.go) for SDK integration testing
    - Added Go test infrastructure with module management and build automation
    - Integrated fake-server lifecycle management for isolated testing environments
    - Implemented process execution abstraction layer for OS-agnostic command handling
    - Added test result aggregation and JSON output formatting for CI/CD pipelines
    - Implemented test independence with per-test resource creation/cleanup
    - Added helper methods to eliminate code duplication (createTestProject, cleanupProject, verifySecretDeleted)
    - Added real vs fake server detection with conditional verification logic
    - Ensured all secret operations associate with projects (real server requirement)


tests/README.md
```diff
@@ -0,0 +1,284 @@
...
+cd tests
+cp SdkTestFramework.Tests/Configuration/.env.example \
+   SdkTestFramework.Tests/Configuration/.env
+```
+
+### Step 3: Run Tests
+
+```bash
+cd tests
+dotnet test
```
tangowithfoxtrot, line 71: I'm following the steps as laid out by this README and I noticed that the tests are passing without actually making any requests to the fake-server at all:

In one terminal:
```console
> cargo run -p fake-server
warning: /private/tmp/sdk-sm/crates/bitwarden-json/Cargo.toml: Found `debug_assertions` in `target.'cfg(...)'.dependencies`. This value is not supported for selecting dependencies and will not work as expected. To learn more visit https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#platform-specific-dependencies
   Compiling serde_json v1.0.149
   Compiling bitflags v2.11.1
   Compiling tower v0.5.3
   Compiling hyper-util v0.1.20
   Compiling aho-corasick v1.1.4
   Compiling time v0.3.47
   Compiling sharded-slab v0.1.7
   Compiling axum-core v0.5.6
   Compiling security-framework v3.7.0
   Compiling rusqlite v0.38.0
   Compiling tower-http v0.6.8
   Compiling regex-automata v0.4.14
   Compiling rustls-platform-verifier v0.6.2
   Compiling hyper-rustls v0.27.9
   Compiling schemars v1.2.1
   Compiling bitwarden-state v2.0.0 (https://github.com/bitwarden/sdk-internal?rev=73cc150a4109bb3de65dd63269cb91ac355a9b1a#73cc150a)
   Compiling reqwest v0.12.28
   Compiling axum v0.8.9
   Compiling reqwest-middleware v0.4.2
   Compiling regex v1.12.3
   Compiling fancy-regex v0.13.0
   Compiling matchers v0.2.0
   Compiling tracing-subscriber v0.3.23
   Compiling bitwarden-api-base v2.0.0 (https://github.com/bitwarden/sdk-internal?rev=73cc150a4109bb3de65dd63269cb91ac355a9b1a#73cc150a)
   Compiling bitwarden-api-key-connector v2.0.0 (https://github.com/bitwarden/sdk-internal?rev=73cc150a4109bb3de65dd63269cb91ac355a9b1a#73cc150a)
   Compiling bitwarden-api-api v2.0.0 (https://github.com/bitwarden/sdk-internal?rev=73cc150a4109bb3de65dd63269cb91ac355a9b1a#73cc150a)
   Compiling bitwarden-crypto v2.0.0 (https://github.com/bitwarden/sdk-internal?rev=73cc150a4109bb3de65dd63269cb91ac355a9b1a#73cc150a)
   Compiling bitwarden-api-identity v2.0.0 (https://github.com/bitwarden/sdk-internal?rev=73cc150a4109bb3de65dd63269cb91ac355a9b1a#73cc150a)
   Compiling zxcvbn v3.1.0
   Compiling validator v0.20.0
   Compiling bitwarden-core v2.0.0 (https://github.com/bitwarden/sdk-internal?rev=73cc150a4109bb3de65dd63269cb91ac355a9b1a#73cc150a)
   Compiling bitwarden-policies v2.0.0 (https://github.com/bitwarden/sdk-internal?rev=73cc150a4109bb3de65dd63269cb91ac355a9b1a#73cc150a)
   Compiling bitwarden-generators v2.0.0 (https://github.com/bitwarden/sdk-internal?rev=73cc150a4109bb3de65dd63269cb91ac355a9b1a#73cc150a)
   Compiling bitwarden-auth v2.0.0 (https://github.com/bitwarden/sdk-internal?rev=73cc150a4109bb3de65dd63269cb91ac355a9b1a#73cc150a)
   Compiling bitwarden-sm v2.0.0 (https://github.com/bitwarden/sdk-internal?rev=73cc150a4109bb3de65dd63269cb91ac355a9b1a#73cc150a)
   Compiling fake-server v0.0.0 (/private/tmp/sdk-sm/crates/fake-server)
    Finished [`dev` profile [optimized + debuginfo]](https://doc.rust-lang.org/cargo/reference/profiles.html#default-profiles) target(s) in 1m 16s
     Running `target/debug/fake-server`
2026-05-18T13:17:24.002115Z  INFO fake_server: Starting fake server on 0.0.0.0:3000
2026-05-18T13:17:24.003327Z  INFO fake_server: Server listening on http://0.0.0.0:3000
```

In another terminal, with the fake-server still running:
```console
Executing Python SDK tests...
      ===================================================

      ═══════════════════════════════════════════════════════════════════
        Python SDK Test Results
      ═══════════════════════════════════════════════════════════════════

        Status: ✅ PASSED
        Platform: OSX (Arm64)
        Duration: 0.02s

        Test Summary:
        ├─ Total:     8 tests
        ├─ Passed:    8 ✅
        ├─ Failed:    0 ❌
        └─ Skipped:   0 ⏭️

        Test Operations:
        ──────────────────────────────────────────────────────────
        📦 Tests
           ├─ ✅ test_auth (5ms)
           ├─ ✅ test_secret_create (2ms)
           ├─ ✅ test_secret_get (1ms)
           ├─ ✅ test_secret_update (1ms)
           ├─ ✅ test_secret_sync
           ├─ ✅ test_secret_delete
           ├─ ✅ test_project_list
           └─ ✅ test_project_update

      ═══════════════════════════════════════════════════════════════════
```

tangowithfoxtrot: This is troubling, as the tests are passing without actually hitting a server in the first place.

tests/README.md
```diff
@@ -0,0 +1,284 @@
...
+{
+  "configuration": {
+    "TEST_MODE": "fake-server",    // or "real-server"
+    "BUILD_SDK": true,              // Build SDKs before testing
+    "AUTO_START_FAKE_SERVER": true, // Auto-start fake server
+    "FAKE_SERVER_PORT": 4000,
+    "ENABLED_LANGUAGES": ["python", "go"]
+  }
+}
+```
```
tangowithfoxtrot, line 137: Not sure how much effort it would take, but making these configurable via environment variables as well would make CI workflows easier.

Also, "fake-server" and "real-server" as strings feels a bit weird. Maybe we could do something like `USE_FAKE_SERVER=<true/false>`?

rdaugherty (commented):
LGTM 🚀

tests/README.md
```diff
@@ -0,0 +1,284 @@
...
+## 🚀 Quick Start
+
+### Step 1: Install Prerequisites
+
+You'll need these tools installed on your system:
+
+| Tool | Required For | Installation |
+|------|-------------|--------------|
+| **.NET 10** | Running tests | [Download](https://dotnet.microsoft.com/download/dotnet/10.0) |
+| **Rust** | Building SDKs & fake-server | `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs \| sh` |
```
tangowithfoxtrot: Some of these commands just dump the scripts to the terminal and throw a curl error:

```console
#!/bin/sh
# shellcheck shell=dash
# shellcheck disable=SC2039  # local is non-POSIX

# This is just a little script that can be downloaded from the internet to
# install rustup. It just does platform detection, downloads the installer
# and runs it.

# It runs on Unix shells like {a,ba,da,k,z}sh. It uses the common `local`
# extension. Note: Most shells limit `local` to 1 var per line, contra bash.

# Some versions of ksh have no `local` keyword. Alias it to `typeset`, but
# beware this makes variables global with f()-style function syntax in ksh93.
# mksh has this alias by default.
has_local() {
    # shellcheck disable=SC2034  # deliberately unused
    local _has_local
}

has_local 2>/dev/null || alias local=typeset

is_zsh() {
    [ -n "${ZSH_VERSION-}" ]
}

set -u

# If RUSTUP_UPDATE_ROOT is unset or empty, default it.
RUSTUP_UPDATE_ROOT="${RUSTUP_UPDATE_ROOT:-https://static.rust-lang.org/rustup}"
# Set quiet as a global for ease of use
RUSTUP_QUIET=no

# NOTICE: If you change anything here, please make the same changes in setup_mode.rs
usage() {
    cat <<EOF
rustup-init 1.29.0 (d243d7d4e 2026-03-04)

The installer for rustup

Usage: rustup-init[EXE] [OPTIONS]

Options:
  -v, --verbose
          Set log level to 'DEBUG' if 'RUSTUP_LOG' is unset
  -q, --quiet
          Disable progress output, set log level to 'WARN' if 'RUSTUP_LOG' is unset
  -y
          Disable confirmation prompt
      --default-host <DEFAULT_HOST>
          Choose a default host triple
      --default-toolchain <DEFAULT_TOOLCHAIN>
          Choose a default toolchain to install. Use 'none' to not install any toolchains at all
      --profile <PROFILE>
          [default: default] [possible values: minimal, default, complete]
  -c, --component <COMPONENT>
          Comma-separated list of component names to also install
  -t, --target <TARGET>
          Comma-separated list of target names to also install
      --no-update-default-toolchain
          Don't update any existing default toolchain after install
      --no-modify-path
          Don't configure the PATH environment variable
  -h, --help
          Print help
  -V, --version
          Print version
EOF
}

main() {
    downloader --check
    need_cmd uname
    need_cmd mktemp
    need_cmd chmod
    need_cmd mkdir
    need_cmd rm
    need_cmd rmdir

    get_architecture || return 1
    local _arch="$RETVAL"
    assert_nz "$_arch" "arch"

    local _default_host_override=""
    case "$_arch" in
        *windows*)
            _default_host_override="--default-host=$_arch"
            ;;
    esac

    local _ext=""
    case "$_arch" in
        *windows*)
            _ext=".exe"
            ;;
    esac

    local _url
    if [ "${RUSTUP_VERSION+set}" = 'set' ]; then
        say "\`RUSTUP_VERSION\` has been set to \`${RUSTUP_VERSION}\`"
        _url="${RUSTUP_UPDATE_ROOT}/archive/${RUSTUP_VERSION}"
    else
        _url="${RUSTUP_UPDATE_ROOT}/dist"
    fi
    _url="${_url}/${_arch}/rustup-init${_ext}"


    local _dir
    if ! _dir="$(ensure mktemp -d)"; then
        # Because the previous command ran in a subshell, we must manually
        # propagate exit status.
        exit 1
    fi
    local _file="${_dir}/rustup-init${_ext}"

    local _ansi_escapes_are_valid=false
    if [ -t 2 ]; then
        if [ "${TERM+set}" = 'set' ]; then
            case "$TERM" in
                xterm*|rxvt*|urxvt*|linux*|vt*)
                    _ansi_escapes_are_valid=true
                ;;
            esac
        fi
    fi

    # check if we have to use /dev/tty to prompt the user
    local need_tty=yes
    for arg in "$@"; do
        case "$arg" in
            --help)
                usage
                exit 0
                ;;
            --quiet)
                RUSTUP_QUIET=yes
                ;;
            *)
                OPTIND=1
                if [ "${arg%%--*}" = "" ]; then
                    # Long option (other than --help);
                    # don't attempt to interpret it.
                    continue
                fi
                while getopts :hqy sub_arg "$arg"; do
                    case "$sub_arg" in
                        h)
                            usage
                            exit 0
                            ;;
                        q)
                            RUSTUP_QUIET=yes
                            ;;
                        y)
                            # user wants to skip the prompt --
                            # we don't need /dev/tty
                            need_tty=no
                            ;;
                        *)
                            ;;
                        esac
                done
                ;;
        esac
    done

    say 'downloading installer'

    ensure mkdir -p "$_dir"
    ensure downloader "$_url" "$_file" "$_arch"
    ensure chmod u+x "$_file"
    if [ ! -x "$_file" ]; then
        err "Cannot execute $_file (likely because of mounting /tmp as noexec)."
        err "Please copy the file to a location where you can execute binaries and run ./rustup-init${_ext}."
        exit 1
    fi

    if [ "$need_tty" = "yes" ] && [ ! -t 0 ]; then
        # The installer is going to want to ask for confirmation by
        # reading stdin.  This script was piped into `sh` though and
        # doesn't have stdin to pass to its children. Instead we're going
        # to explicitly connect /dev/tty to the installer's stdin.
        if [ ! -t 1 ]; then
            err "Unable to run interactively. Run with -y to accept defaults, --help for additional options"
            exit 1;
        fi

        ignore "$_file" ${_default_host_override:+"$_default_host_override"} "$@" < /dev/tty
    else
        ignore "$_file" ${_default_host_override:+"$_default_host_override"} "$@"
    fi

    local _retval=$?

    ignore rm "$_file"
    ignore rmdir "$_dir"

    return "$_retval"
}

get_current_exe() {
    # Returns the executable used for system architecture detection
    # This is only run on Linux
    local _current_exe
    if test -L /proc/self/exe ; then
        _current_exe=/proc/self/exe
    else
        warn "Unable to find /proc/self/exe. System architecture detection might be inaccurate."
        if test -n "$SHELL" ; then
            _current_exe=$SHELL
        else
            need_cmd /bin/sh
            _current_exe=/bin/sh
        fi
        warn "Falling back to $_current_exe."
    fi
    echo "$_current_exe"
}

get_bitness() {
    need_cmd head
    # Architecture detection without dependencies beyond coreutils.
    # ELF files start out "\x7fELF", and the following byte is
    #   0x01 for 32-bit and
    #   0x02 for 64-bit.
    # The printf builtin on some shells like dash only supports octal
    # escape sequences, so we use those.
    local _current_exe=$1
    local _current_exe_head
    _current_exe_head=$(head -c 5 "$_current_exe")
    if [ "$_current_exe_head" = "$(printf '\177ELF\001')" ]; then
        echo 32
    elif [ "$_current_exe_head" = "$(printf '\177ELF\002')" ]; then
        echo 64
    else
        err "unknown platform bitness"
        exit 1;
    fi
}

is_host_amd64_elf() {
    local _current_exe=$1

    need_cmd head
    need_cmd tail
    # ELF e_machine detection without dependencies beyond coreutils.
    # Two-byte field at offset 0x12 indicates the CPU,
    # but we're interested in it being 0x3E to indicate amd64, or not that.
    local _current_exe_machine
    _current_exe_machine=$(head -c 19 "$_current_exe" | tail -c 1)
    [ "$_current_exe_machine" = "$(printf '\076')" ]
}

get_endianness() {
    local _current_exe=$1
    local cputype=$2
    local suffix_eb=$3
    local suffix_el=$4

    # detect endianness without od/hexdump, like get_bitness() does.
    need_cmd head
    need_cmd tail

    local _current_exe_endianness
    _current_exe_endianness="$(head -c 6 "$_current_exe" | tail -c 1)"
    if [ "$_current_exe_endianness" = "$(printf '\001')" ]; then
        echo "${cputype}${suffix_el}"
    elif [ "$_current_exe_endianness" = "$(printf '\002')" ]; then
        echo "${cputype}${suffix_eb}"
    else
        err "unknown platform endianness"
        exit 1
    fi
}

# Detect the Linux/LoongArch UAPI flavor, with all errors being non-fatal.
# Returns 0 or 234 in case of successful detection, 1 otherwise (/tmp being
# noexec, or other causes).
check_loongarch_uapi() {
    need_cmd base64

    local _tmp
    if ! _tmp="$(ensure mktemp)"; then
        return 1
    fi

    # Minimal Linux/LoongArch UAPI detection, exiting with 0 in case of
    # upstream ("new world") UAPI, and 234 (-EINVAL truncated) in case of
    # old-world (as deployed on several early commercial Linux distributions
    # for LoongArch).
    #
    # See https://gist.github.com/xen0n/5ee04aaa6cecc5c7794b9a0c3b65fc7f for
    # source to this helper binary.
    ignore base64 -d > "$_tmp" <<EOF
f0VMRgIBAQAAAAAAAAAAAAIAAgEBAAAAeAAgAAAAAABAAAAAAAAAAAAAAAAAAAAAQQAAAEAAOAAB
AAAAAAAAAAEAAAAFAAAAAAAAAAAAAAAAACAAAAAAAAAAIAAAAAAAJAAAAAAAAAAkAAAAAAAAAAAA
AQAAAAAABCiAAwUAFQAGABUAByCAAwsYggMAACsAC3iBAwAAKwAxen0n
EOF

    ignore chmod u+x "$_tmp"
    if [ ! -x "$_tmp" ]; then
        ignore rm "$_tmp"
        return 1
    fi

    "$_tmp"
    local _retval=$?

    ignore rm "$_tmp"
    return "$_retval"
}

ensure_loongarch_uapi() {
    check_loongarch_uapi
    case $? in
        0)
            return 0
            ;;
        234)
            err 'Your Linux kernel does not provide the ABI required by this Rust distribution.'
            err 'Please check with your OS provider for how to obtain a compatible Rust package for your system.'
            exit 1
            ;;
        *)
            warn "Cannot determine current system's ABI flavor, continuing anyway."
            warn 'Note that the official Rust distribution only works with the upstream kernel ABI.'
            warn 'Installation will fail if your running kernel happens to be incompatible.'
            ;;
    esac
}

get_architecture() {
    local _ostype _cputype _bitness _arch _clibtype
    _ostype="$(uname -s)"
    _cputype="$(uname -m)"
    _clibtype="gnu"

    if [ "$_ostype" = Linux ]; then
        if [ "$(uname -o)" = Android ]; then
            _ostype=Android
        fi
        if ldd --version 2>&1 | grep -q 'musl'; then
            _clibtype="musl"
        fi
    fi

    if [ "$_ostype" = Darwin ]; then
        # Darwin `uname -m` can lie due to Rosetta shenanigans. If you manage to
        # invoke a native shell binary and then a native uname binary, you can
        # get the real answer, but that's hard to ensure, so instead we use
        # `sysctl` (which doesn't lie) to check for the actual architecture.
        if [ "$_cputype" = i386 ]; then
            # Handling i386 compatibility mode in older macOS versions (<10.15)
            # running on x86_64-based Macs.
            # Starting from 10.15, macOS explicitly bans all i386 binaries from running.
            # See: <https://support.apple.com/en-us/HT208436>

            # Avoid `sysctl: unknown oid` stderr output and/or non-zero exit code.
            if (sysctl hw.optional.x86_64 2> /dev/null || true) | grep -q ': 1'; then
                _cputype=x86_64
            fi
        elif [ "$_cputype" = x86_64 ]; then
            # Handling x86-64 compatibility mode (a.k.a. Rosetta 2)
            # in newer macOS versions (>=11) running on arm64-based Macs.
            # Rosetta 2 is built exclusively for x86-64 and cannot run i386 binaries.

            # Avoid `sysctl: unknown oid` stderr output and/or non-zero exit code.
            if (sysctl hw.optional.arm64 2> /dev/null || true) | grep -q ': 1'; then
                _cputype=arm64
            fi
        fi
    fi

    if [ "$_ostype" = SunOS ]; then
        # Both Solaris and illumos presently announce as "SunOS" in "uname -s"
        # so use "uname -o" to disambiguate.  We use the full path to the
        # system uname in case the user has coreutils uname first in PATH,
        # which has historically sometimes printed the wrong value here.
        if [ "$(/usr/bin/uname -o)" = illumos ]; then
            _ostype=illumos
        fi

        # illumos systems have multi-arch userlands, and "uname -m" reports the
        # machine hardware name; e.g., "i86pc" on both 32- and 64-bit x86
        # systems.  Check for the native (widest) instruction set on the
        # running kernel:
        if [ "$_cputype" = i86pc ]; then
            _cputype="$(isainfo -n)"
        fi
    fi

    local _current_exe
    case "$_ostype" in

        Android)
            _ostype=linux-android
            ;;

        Linux)
            _current_exe=$(get_current_exe)
            _ostype=unknown-linux-$_clibtype
            _bitness=$(get_bitness "$_current_exe")
            ;;

        FreeBSD)
            _ostype=unknown-freebsd
            ;;

        NetBSD)
            _ostype=unknown-netbsd
            ;;

        DragonFly)
            _ostype=unknown-dragonfly
            ;;

        Darwin)
            _ostype=apple-darwin
            ;;

        illumos)
            _ostype=unknown-illumos
            ;;

        SunOS)
            if [ "$_cputype" = sun4v ]; then
               _ostype=sun-solaris
            else
               _ostype=pc-solaris
            fi
            ;;

        MINGW* | MSYS* | CYGWIN* | Windows_NT)
            _ostype=pc-windows-gnu
            ;;

        *)
            err "unrecognized OS type: $_ostype"
            exit 1
            ;;

    esac

    case "$_cputype" in

        i386 | i486 | i686 | i786 | x86)
            _cputype=i686
            ;;

        xscale | arm)
            _cputype=arm
            if [ "$_ostype" = "linux-android" ]; then
                _ostype=linux-androideabi
            fi
            ;;

        armv6l)
            _cputype=arm
            if [ "$_ostype" = "linux-android" ]; then
                _ostype=linux-androideabi
            else
                _ostype="${_ostype}eabihf"
            fi
            ;;

        armv7l | armv8l)
            _cputype=armv7
            if [ "$_ostype" = "linux-android" ]; then
                _ostype=linux-androideabi
            else
                _ostype="${_ostype}eabihf"
            fi
            ;;

        aarch64 | arm64)
            _cputype=aarch64
            ;;

        x86_64 | x86-64 | x64 | amd64)
            _cputype=x86_64
            ;;

        mips)
            _cputype=$(get_endianness "$_current_exe" mips '' el)
            ;;

        mips64)
            if [ "$_bitness" -eq 64 ]; then
                # only n64 ABI is supported for now
                _ostype="${_ostype}abi64"
                _cputype=$(get_endianness "$_current_exe" mips64 '' el)
            fi
            ;;

        ppc)
            _cputype=powerpc
            ;;

        ppc64)
            _cputype=powerpc64
            ;;

        ppc64le)
            _cputype=powerpc64le
            ;;

        s390x)
            _cputype=s390x
            ;;
        sun4v)
            _cputype=sparcv9
            ;;
        riscv64)
            _cputype=riscv64gc
            ;;
        loongarch64)
            _cputype=loongarch64
            ensure_loongarch_uapi
            ;;
        *)
            err "unknown CPU type: $_cputype"
            exit 1

    esac

    # Detect 64-bit linux with 32-bit userland
    if [ "${_ostype}" = unknown-linux-gnu ] && [ "${_bitness}" -eq 32 ]; then
        case $_cputype in
            x86_64)
                if [ -n "${RUSTUP_CPUTYPE:-}" ]; then
                    _cputype="$RUSTUP_CPUTYPE"
                else {
                    # 32-bit executable for amd64 = x32
                    if is_host_amd64_elf "$_current_exe"; then {
                        err "This host is running an x32 userland, for which no native toolchain is provided."
                        err "You will have to install multiarch compatibility with i686 or amd64."
                        err "To do so, set the RUSTUP_CPUTYPE environment variable set to i686 or amd64 and re-run this script."
                        err "You will be able to add an x32 target after installation by running \`rustup target add x86_64-unknown-linux-gnux32\`."
                         exit 1
                    }; else
                        _cputype=i686
                    fi
                }; fi
                ;;
            mips64)
                _cputype=$(get_endianness "$_current_exe" mips '' el)
                ;;
            powerpc64)
                _cputype=powerpc
                ;;
            aarch64)
                _cputype=armv7
                if [ "$_ostype" = "linux-android" ]; then
                    _ostype=linux-androideabi
                else
                    _ostype="${_ostype}eabihf"
                fi
                ;;
            riscv64gc)
                err "riscv64 with 32-bit userland unsupported"
                exit 1
                ;;
        esac
    fi

    # Detect armv7 but without the CPU features Rust needs in that build,
    # and fall back to arm.
    # See https://github.com/rust-lang/rustup.rs/issues/587.
    if [ "$_ostype" = "unknown-linux-gnueabihf" ] && [ "$_cputype" = armv7 ]; then
        if ! (ensure grep '^Features' /proc/cpuinfo | grep -E -q 'neon|simd') ; then
            # Either `/proc/cpuinfo` is malformed or unavailable, or
            # at least one processor does not have NEON (which is asimd on armv8+).
            _cputype=arm
        fi
    fi

    _arch="${_cputype}-${_ostype}"

    RETVAL="$_arch"
}

__print() {
    if $_ansi_escapes_are_valid; then
        printf '\33[1m%s:\33[0m %s\n' "$1" "$2" >&2
    else
        printf '%s: %s\n' "$1" "$2" >&2
    fi
}

warn() {
    __print 'warn' "$1" >&2
}

say() {
    if [ "$RUSTUP_QUIET" = "no" ]; then
        __print 'info' "$1" >&2
    fi
}

# NOTE: you are required to exit yourself
# we don't do it here because of multiline errors
err() {
    __print 'error' "$1" >&2
}

need_cmd() {
    if ! check_cmd "$1"; then
        err "need '$1' (command not found)"
        exit 1
    fi
}

check_cmd() {
    command -v "$1" > /dev/null 2>&1
}

assert_nz() {
    if [ -z "$1" ]; then
        err "assert_nz $2"
        exit 1
    fi
}

# Run a command that should never fail. If the command fails execution
# will immediately terminate with an error showing the failing
# command.
ensure() {
    if ! "$@"; then
        err "command failed: $*"
        exit 1
    fi
}

# This is just for indicating that commands' results are being
# intentionally ignored. Usually, because it's being executed
# as part of error handling.
ignore() {
    "$@"
}

# This wraps curl or wget. Try curl first, if not installed,
# use wget instead.
downloader() {
    # zsh does not split words by default, Required for curl retry arguments below.
    is_zsh && setopt local_options shwordsplit

    local _dld
    local _ciphersuites
    local _err
    local _status
    local _retry
    if check_cmd curl; then
        # Check if we have a broken snap curl
        # https://github.com/boukendesho/curl-snap/issues/1
        _curl_path=$(command -v curl)
        if echo "$_curl_path" | grep "/snap/" > /dev/null 2>&1; then
            if check_cmd wget; then
                _dld=wget
            else
                err "curl installed with snap cannot be used to install Rust"
                err "due to missing permissions. Please uninstall it and"
                err "reinstall curl with a different package manager (e.g., apt)."
                err "See https://github.com/boukendesho/curl-snap/issues/1"
                exit 1
            fi
        else
            _dld=curl
        fi
    elif check_cmd wget; then
        _dld=wget
    else
        _dld='curl or wget' # to be used in error message of need_cmd
    fi

    if [ "$1" = --check ]; then
        need_cmd "$_dld"
    elif [ "$_dld" = curl ]; then
        check_curl_for_retry_support
        _retry="$RETVAL"
        get_ciphersuites_for_curl
        _ciphersuites="$RETVAL"
        if [ -n "$_ciphersuites" ]; then
            # shellcheck disable=SC2086
            _err=$(curl $_retry --proto '=https' --tlsv1.2 --ciphers "$_ciphersuites" --silent --show-error --fail --location "$1" --output "$2" 2>&1)
            _status=$?
        else
            warn "Not enforcing strong cipher suites for TLS, this is potentially less secure"
            if ! check_help_for "$3" curl --proto --tlsv1.2; then
                warn "Not enforcing TLS v1.2, this is potentially less secure"
                # shellcheck disable=SC2086
                _err=$(curl $_retry --silent --show-error --fail --location "$1" --output "$2" 2>&1)
                _status=$?
            else
                # shellcheck disable=SC2086
                _err=$(curl $_retry --proto '=https' --tlsv1.2 --silent --show-error --fail --location "$1" --output "$2" 2>&1)
                _status=$?
            fi
        fi
        if [ -n "$_err" ]; then
            warn "$_err"
            if echo "$_err" | grep -q 404$; then
                err "installer for platform '$3' not found, this may be unsupported"
                exit 1
            fi
        fi
        return $_status
    elif [ "$_dld" = wget ]; then
        if [ "$(wget -V 2>&1|head -2|tail -1|cut -f1 -d" ")" = "BusyBox" ]; then
            warn "using the BusyBox version of wget.  Not enforcing strong cipher suites for TLS or TLS v1.2, this is potentially less secure"
            _err=$(wget "$1" -O "$2" 2>&1)
            _status=$?
        else
            get_ciphersuites_for_wget
            _ciphersuites="$RETVAL"
            if [ -n "$_ciphersuites" ]; then
                _err=$(wget --https-only --secure-protocol=TLSv1_2 --ciphers "$_ciphersuites" "$1" -O "$2" 2>&1)
                _status=$?
            else
                warn "Not enforcing strong cipher suites for TLS, this is potentially less secure"
                if ! check_help_for "$3" wget --https-only --secure-protocol; then
                    warn "Not enforcing TLS v1.2, this is potentially less secure"
                    _err=$(wget "$1" -O "$2" 2>&1)
                    _status=$?
                else
                    _err=$(wget --https-only --secure-protocol=TLSv1_2 "$1" -O "$2" 2>&1)
                    _status=$?
                fi
            fi
        fi
        if [ -n "$_err" ]; then
            warn "$_err"
            if echo "$_err" | grep -q ' 404 Not Found$'; then
                err "installer for platform '$3' not found, this may be unsupported"
                exit 1
            fi
        fi
        return $_status
    else
        err "Unknown downloader"   # should not reach here
        exit 1
    fi
}

check_help_for() {
    local _arch
    local _cmd
    local _arg
    _arch="$1"
    shift
    _cmd="$1"
    shift

    local _category
    if "$_cmd" --help | grep -q '"--help all"'; then
      _category="all"
    else
      _category=""
    fi

    case "$_arch" in

        *darwin*)
        if check_cmd sw_vers; then
            local _os_version
            local _os_major
            _os_version=$(sw_vers -productVersion)
            _os_major=$(echo "$_os_version" | cut -d. -f1)
            case $_os_major in
                10)
                    # If we're running on macOS, older than 10.13, then we always
                    # fail to find these options to force fallback
                    if [ "$(echo "$_os_version" | cut -d. -f2)" -lt 13 ]; then
                        # Older than 10.13
                        warn "Detected macOS platform older than 10.13"
                        return 1
                    fi
                    ;;
                *)
                    if ! { [ "$_os_major" -eq "$_os_major" ] 2>/dev/null && [ "$_os_major" -ge 11 ]; }; then
                        # Unknown product version, warn and continue
                        warn "Detected unknown macOS major version: $_os_version"
                        warn "TLS capabilities detection may fail"
                    fi
                    ;; # We assume that macOS v11+ will always be okay.
            esac
        fi
        ;;

    esac

    for _arg in "$@"; do
        if ! "$_cmd" --help "$_category" | grep -q -- "$_arg"; then
            return 1
        fi
    done

    true # not strictly needed
}

# Check if curl supports the --retry flag, then pass it to the curl invocation.
check_curl_for_retry_support() {
    local _retry_supported=""
    # "unspecified" is for arch, allows for possibility old OS using macports, homebrew, etc.
    if check_help_for "notspecified" "curl" "--retry"; then
        _retry_supported="--retry 3"
        if check_help_for "notspecified" "curl" "--continue-at"; then
            # "-C -" tells curl to automatically find where to resume the download when retrying.
            _retry_supported="--retry 3 -C -"
        fi
    fi

    RETVAL="$_retry_supported"
}

# Return cipher suite string specified by user, otherwise return strong TLS 1.2-1.3 cipher suites
# if support by local tools is detected. Detection currently supports these curl backends:
# GnuTLS and OpenSSL (possibly also LibreSSL and BoringSSL). Return value can be empty.
get_ciphersuites_for_curl() {
    if [ -n "${RUSTUP_TLS_CIPHERSUITES-}" ]; then
        # user specified custom cipher suites, assume they know what they're doing
        RETVAL="$RUSTUP_TLS_CIPHERSUITES"
        return
    fi

    local _openssl_syntax="no"
    local _gnutls_syntax="no"
    local _backend_supported="yes"
    if curl -V | grep -q ' OpenSSL/'; then
        _openssl_syntax="yes"
    elif curl -V | grep -iq ' LibreSSL/'; then
        _openssl_syntax="yes"
    elif curl -V | grep -iq ' BoringSSL/'; then
        _openssl_syntax="yes"
    elif curl -V | grep -iq ' GnuTLS/'; then
        _gnutls_syntax="yes"
    else
        _backend_supported="no"
    fi

    local _args_supported="no"
    if [ "$_backend_supported" = "yes" ]; then
        # "unspecified" is for arch, allows for possibility old OS using macports, homebrew, etc.
        if check_help_for "notspecified" "curl" "--tlsv1.2" "--ciphers" "--proto"; then
            _args_supported="yes"
        fi
    fi

    local _cs=""
    if [ "$_args_supported" = "yes" ]; then
        if [ "$_openssl_syntax" = "yes" ]; then
            _cs=$(get_strong_ciphersuites_for "openssl")
        elif [ "$_gnutls_syntax" = "yes" ]; then
            _cs=$(get_strong_ciphersuites_for "gnutls")
        fi
    fi

    RETVAL="$_cs"
}

# Return cipher suite string specified by user, otherwise return strong TLS 1.2-1.3 cipher suites
# if support by local tools is detected. Detection currently supports these wget backends:
# GnuTLS and OpenSSL (possibly also LibreSSL and BoringSSL). Return value can be empty.
get_ciphersuites_for_wget() {
    if [ -n "${RUSTUP_TLS_CIPHERSUITES-}" ]; then
        # user specified custom cipher suites, assume they know what they're doing
        RETVAL="$RUSTUP_TLS_CIPHERSUITES"
        return
    fi

    local _cs=""
    if wget -V | grep -q '\-DHAVE_LIBSSL'; then
        # "unspecified" is for arch, allows for possibility old OS using macports, homebrew, etc.
        if check_help_for "notspecified" "wget" "TLSv1_2" "--ciphers" "--https-only" "--secure-protocol"; then
            _cs=$(get_strong_ciphersuites_for "openssl")
        fi
    elif wget -V | grep -q '\-DHAVE_LIBGNUTLS'; then
        # "unspecified" is for arch, allows for possibility old OS using macports, homebrew, etc.
        if check_help_for "notspecified" "wget" "TLSv1_2" "--ciphers" "--https-only" "--secure-protocol"; then
            _cs=$(get_strong_ciphersuites_for "gnutls")
        fi
    fi

    RETVAL="$_cs"
}

# Return strong TLS 1.2-1.3 cipher suites in OpenSSL or GnuTLS syntax. TLS 1.2
# excludes non-ECDHE and non-AEAD cipher suites. DHE is excluded due to bad
# DH params often found on servers (see RFC 7919). Sequence matches or is
# similar to Firefox 68 ESR with weak cipher suites disabled via about:config.
# $1 must be openssl or gnutls.
get_strong_ciphersuites_for() {
    if [ "$1" = "openssl" ]; then
        # OpenSSL is forgiving of unknown values, no problems with TLS 1.3 values on versions that don't support it yet.
        echo "TLS_AES_128_GCM_SHA256:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_256_GCM_SHA384:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384"
    elif [ "$1" = "gnutls" ]; then
        # GnuTLS isn't forgiving of unknown values, so this may require a GnuTLS version that supports TLS 1.3 even if wget doesn't.
        # Begin with SECURE128 (and higher) then remove/add to build cipher suites. Produces same 9 cipher suites as OpenSSL but in slightly different order.
        echo "SECURE128:-VERS-SSL3.0:-VERS-TLS1.0:-VERS-TLS1.1:-VERS-DTLS-ALL:-CIPHER-ALL:-MAC-ALL:-KX-ALL:+AEAD:+ECDHE-ECDSA:+ECDHE-RSA:+AES-128-GCM:+CHACHA20-POLY1305:+AES-256-GCM"
    fi
}

set +u
case "$RUSTUP_INIT_SH_PRINT" in
    arch | architecture)
        get_architecture || exit 1
        echo "$RETVAL"
        ;;
    *)
        main "$@" || exit 1
        ;;
esac
curl: (1) Protocol "http" disabled
curl: (1) Protocol "http" disabled
```

tangowithfoxtrot (changes requested):

tests/README.md
ddanielyan: Ill look more into this, TY, i tested with real-server and fake maybe something i recently changed might of broke

tests/README.md
ddanielyan: the configuration file is similar to a .env in way that we set values there but the difference .env being more for secrets and config.json being for everything else, i can look into adding these as optional to the .env

languages/go/test/test_suite.go
```diff
@@ -0,0 +1,657 @@
...
+		return fmt.Errorf("failed to create client: %w", err)
+	}
+
+	s.client = client
+	return nil
+}
+
+// createTestProject creates a project for testing and returns the project ID
+func (s *GoSDKTestSuite) createTestProject(purpose string) (string, error) {
+	projectName := fmt.Sprintf("test-project-%s-%s", purpose, uuid.Must(uuid.NewV4()).String()[:8])
```
tangowithfoxtrot: What is this UUID in the project name used for? It seems like we could get rid of it?

languages/go/test/test_suite.go
```diff
@@ -0,0 +1,657 @@
...
+	project, err := s.client.Projects().Create(s.organizationID, projectName)
+	if err != nil {
+		return "", fmt.Errorf("failed to create project: %w", err)
+	}
+	return project.ID, nil
+}
+
+// cleanupProject deletes a project and verifies cleanup on real server
```
tangowithfoxtrot, line 90: The cleanups are greatly appreciated 😄

languages/go/test/test_suite.go
```diff
@@ -0,0 +1,657 @@
...
+	}
+	return nil
+}
+
+// RunOperation executes a test operation and tracks the result
+func (s *GoSDKTestSuite) RunOperation(name string, testFunc func() (bool, map[string]interface{}, error), displayName string) bool {
+	if displayName == "" {
+		displayName = strings.ReplaceAll(name, "_", " ")
+		displayName = strings.Title(displayName)
+	}
```
tangowithfoxtrot, line 129: This might be a silly question, but when would the displayName be snake_case? Why do the ReplaceAll here?

languages/go/test/test_suite.go
```diff
@@ -0,0 +1,657 @@
...
+	// On real server, verify authentication worked by trying to sync
+	verified := false
+	if s.testMode == "real-server" {
+		// Try to sync - should work if authenticated
+		_, err := s.client.Secrets().Sync(s.organizationID, nil)
+		if err != nil {
+			return false, nil, fmt.Errorf("authentication verification failed: %w", err)
+		}
+		verified = true
+	}
```
tangowithfoxtrot: This is a different "sync" than the one that happens for password manager. It can be run on the fake server as well.

languages/go/test/test_suite.go
```diff
@@ -0,0 +1,657 @@
...
+
+	return true, map[string]interface{}{
+		"method":    "access_token",
```
tangowithfoxtrot: We should remove references to "method", as we only use access tokens for authentication.

languages/go/test/test_suite.go
```diff
@@ -0,0 +1,657 @@
...
+	if err != nil {
+		return false, nil, err
+	}
+
+	return true, map[string]interface{}{
+		"initial_has_changes": syncResponse.HasChanges,
+		"after_has_changes":   syncResponseWithDate.HasChanges,
+		"secret_count":        len(syncResponse.Secrets),
+	}, nil
+}
```
tangowithfoxtrot: These properties don't convey much meaningful information. Maybe something like a "sync_succeeded" would be better. The success status should be checked in one case by passing `nil` for the `lastSyncedDate` and ensuring that `hasChanges == true`. Then, pass `time.Now()` for the last sync date and ensure that `hasChanges == false`.

languages/go/test/test_suite.go
```diff
@@ -0,0 +1,657 @@
...
+	if err := s.cleanupProject(projectID); err != nil {
+		return false, nil, err
+	}
+
+	return true, map[string]interface{}{
+		"deleted":  1,
+		"id":       secretID,
+		"verified": s.testMode == "real-server",
+	}, nil
+}
```
tangowithfoxtrot, line 441: Similarly here, would it be better to have this data reflect whether the deletion succeeded, rather than that 1 secret was deleted?

languages/go/test/test_suite.go
```diff
@@ -0,0 +1,657 @@
...
+		{"test_auth", s.TestAuth, "Authentication"},
+		{"test_secret_create", s.TestSecretCreate, "Create Secret"},
+		{"test_secret_get", s.TestSecretGet, "Get Secret"},
+		{"test_secret_update", s.TestSecretUpdate, "Update Secret"},
+		{"test_secret_sync", s.TestSecretSync, "Sync Secrets"},
+		{"test_secret_delete", s.TestSecretDelete, "Delete Secrets"},
+		{"test_project_list", s.TestProjectList, "List Projects"},
+		{"test_project_update", s.TestProjectUpdate, "Update Project"},
+	}
+}
```
tangowithfoxtrot, line 712: The tests are missing the secrets generator.

languages/go/test/test_suite.go
```diff
@@ -0,0 +1,657 @@
...
+
+// GenerateReport outputs the test report
+func (s *GoSDKTestSuite) GenerateReport(totalDuration int64) {
```
tangowithfoxtrot, line 715: Just wondering how this will affect the ability to generate meaningful XUnit/JUnit style outputs for CI test results? I would assume that generating them at the language level would probably be easier to get meaningful output. Or were you going to try to do that with the C# framework?

languages/go/test/test_suite.go
```diff
@@ -0,0 +1,657 @@
...
+
+
+func main() {
+	jsonOutput := flag.Bool("json", false, "Output JSON format")
+	verbose := flag.Bool("verbose", false, "Verbose output")
+	flag.Parse()
+
+	suite := NewTestSuite(*jsonOutput, *verbose)
+	os.Exit(suite.RunAllTests())
+}
```
tangowithfoxtrot: Could you run `go fmt` on any go related files that have been changed?

tests/README.md
```diff
@@ -0,0 +1,284 @@
...
+
+That's it! The tests should now run successfully. 🎉
+
+## 📋 What Gets Tested
+
+The framework tests these SDK operations:
+- Authentication with access tokens
+- Secret CRUD operations (Create, Read, Update, Delete)
+- Project CRUD operations
+- Password generation
```
tangowithfoxtrot: super small nit:

```suggestion
- Secret generation
```

tests/README.md
tangowithfoxtrot: Thanks. My thought behind retaining configuration via environment variables is that if we're going to retain the ability to run these tests without dotnet, we wouldn't be able to just `cargo|go|python test ...` unless you had something configured for all of the languages to pull those JSON values.

We'd also need to retain some ability to kick of the tests via the shell scripts if we don't have a way to run the new test suite in CI. If we can get the testing workflows updated to work with the new test suite, then configuring via env vars is probably less important.

tests/README.md
```diff
@@ -0,0 +1,284 @@
...
+
+# Run with specific configuration
+dotnet test -- TestRunParameters.Parameter(name="verbose", value="true")
+```
+
+## 📁 Project Structure
+
+```
+tests/
+├── SdkTestFramework/           # Core framework library
```
tangowithfoxtrot:
```suggestion
├── SdkTestFramework/          # Core framework library
```

tests/README.md
```diff
@@ -0,0 +1,284 @@
...
+│   ├── Config/                # Configuration management
+│   ├── Models/                # Test result models
+│   ├── Platform/              # Platform detection
+│   ├── Services/              # Test services
+│   └── TestRunners/           # Language-specific runners
+│       ├── PythonTestRunner.cs
+│       └── GoTestRunner.cs
+│
+├── SdkTestFramework.Tests/    # Test project
```
tangowithfoxtrot:
```suggestion
├── SdkTestFramework.Tests/   # Test project
```

tests/SdkTestFramework.Tests/SdkWrappers/GoTests.cs
```diff
@@ -0,0 +1,39 @@
...
+    [Property("TestId", "SDK-GO-001")]
+    public async Task Go_SDK_Should_Pass_All_Tests()
+    {
+        // Execute SDK tests
+        var result = await ExecuteSdkTests();
+
+        // Validate results
+        ValidateTestResult(result, SdkLanguage);
+    }
+}
```
tangowithfoxtrot, line 61: If we are running all the tests for a single language at once, can we have each language provide detailed, granular output so that we can get comprehensive test summaries similar to the ones in the [test repo](https://github.com/bitwarden/test/actions/runs/26109796914)?

tests/SdkTestFramework.Tests/SdkWrappers/SdkWrappersDataSetup.cs
```diff
@@ -0,0 +1,32 @@
...
+namespace SdkTestFramework.Tests.SdkWrappers
+{
+    /// <summary>
+    /// Data setup for SDK wrapper tests
+    /// </summary>
+    [SetUpFixture]
+    public class SdkWrappersDataSetup
+    {
+        [OneTimeSetUp]
+        public async Task SetUp()
```
tangowithfoxtrot: Are these setup steps still going to be implemented? I believe they would be necessary to get the test suite runnable in CI. The `bootstrap.sh` script currently handles a lot of this.

tests/SdkTestFramework/Common/PathUtilities.cs
```diff
@@ -0,0 +1,59 @@
...
+        return FindRootPath(startPath, ".git", isDirectory: true);
+    }
+
+    /// <summary>
+    /// Finds the SDK root by looking for Cargo.toml file
+    /// </summary>
+    private static string? FindSdkRoot(string startPath)
+    {
+        return FindRootPath(startPath, "Cargo.toml", isDirectory: false);
+    }
```
tangowithfoxtrot, line 82: This won't always be true. The SDK root has a `Cargo.toml` file for the workspace, but each crate within the workspace will also have its own cargo file. Example: `./crates/*/Cargo.toml`.

tests/SdkTestFramework/Services/ProcessRequest.cs
```diff
@@ -0,0 +1,54 @@
+
```
tangowithfoxtrot: I think the C# files could benefit from being run through a formatter as well.

tests/SdkTestFramework/Services/TestResultFormatter.cs
```diff
@@ -0,0 +1,172 @@
+using System.Text;
```
tangowithfoxtrot, line 1: This is where I would hope that we could get a nice NUnit/XUnit style report on the status of the language wrapper tests. Otherwise, maybe it'd be better to generate those in the language wrapper tests instead of here in C#?

tangowithfoxtrot (changes requested):

languages/go/test/test_suite.go
ddanielyan: fixed and remove this logic

languages/go/test/test_suite.go
ddanielyan: this is not as useful for when running on fake-server
but on a real server if for some reason data gets left behind from a test it wont have conflicts + if we decide to run in parallel it will help avoid conflicts and probably less likely to happen if 2 people are running tests against the same creds

languages/go/test/test_suite.go
ddanielyan: fixed and added for fake server

languages/go/test/test_suite.go
ddanielyan: ⏺ Perfect! The verification already happens in verifySecretDeleted which:
  - For real-server: Tries to GET the secret and expects a "not found" error
  - For fake-server: Skips verification (trusts the delete response)

 1. Simplified return values:
    - Removed "deleted": 1 - redundant since we only delete one secret
    - Removed "id": secretID - not useful information
    - Removed "verified": s.testMode == "real-server" - confusing
  2. Single meaningful return value:
    - Just "deletion_succeeded": true
    - The test passing itself means deletion worked
  3. Verification strategy:
    - Real server: Actually tries to GET the deleted secret and expects "not found"
    - Fake server: Trusts the delete response 
    - If verification fails, the test fails with an error

languages/go/test/test_suite.go
ddanielyan: 1. Added proper assertions:
    - First sync (nil date) MUST return has_changes: true
    - Second sync (current time) MUST return has_changes: false
  2. Simplified return values:
    - Replaced meaningless properties with sync_succeeded: true
    - Only return initial_secrets count for debugging
  3. Handle server differences:
    - Added comments explaining the fake-server's flawed implementation
    - Both environments now expect has_changes: false for the second sync

tests/SdkTestFramework.Tests/SdkWrappers/GoTests.cs
ddanielyan: Added Detailed Summaries:
  ═══════════════════════════════════════════════════════
    Python SDK Test Operations Summary
  ═══════════════════════════════════════════════════════
    ✅ PASSED - test_auth (45ms)
    ✅ PASSED - test_secret_create (23ms)
    ❌ FAILED - test_secret_update (31ms)
           Error: Update failed: permission denied
  ─────────────────────────────────────────────────────
    Total: 8 | Passed: 7 | Failed: 1

languages/go/test/test_suite.go
ddanielyan: ## CI Test Reporting

  a hybrid approach that gives us the best of both worlds:

  **For Developers**: Run tests directly without .NET (`python test_suite.py --json`)

  **For CI/CD**: Use the Dynamic test classes which convert each operation into a separate NUnit test case. This gives us proper JUnit/XUnit XML output where each operation (auth, create, delete, etc.) shows up as an individual test in CI dashboards, rather than one
  big pass/fail for the entire language.

  Example: Instead of "Python_SDK_Tests: ✅", CI shows:
  - Python_Auth: ✅ (45ms)
  - Python_Secret_Create: ✅ (23ms)
  - Python_Secret_Delete: ✅ (19ms)

  This way we get granular test reporting without implementing XML generation in each language - NUnit handles it for us.

tests/SdkTestFramework.Tests/SdkWrappers/SdkWrappersDataSetup.cs
ddanielyan: i have removed the SdkWrapperDataSetup.cs the global.cs handles all the needed setup
i have brought back the bootstrap.sh and the test.sh so the workflows currently still pass and the tests can run without the dotnet, in another PR we can update the workflow to use dotnet

tests/SdkTestFramework/Common/PathUtilities.cs
ddanielyan: updated this to 
    -Check for [workspace] in Cargo.toml
    - Most accurate - definitively identifies workspace root
    - Reading file is not expensive
    - Future-proof - Cargo workspace structure is standard

tests/SdkTestFramework/Services/TestResultFormatter.cs
ddanielyan: This has been updated
  
  1. NUnit-style reporting - The test results are properly integrated with NUnit's test runner, showing up in the standard NUnit test output with proper test names, categories, and properties
  2. Language-specific details - The TestResultFormatter takes the JSON output from language tests and formats it nicely for display within the NUnit context

languages/go/test/test_suite.go
ddanielyan: added generator tests

tests/README.md
ddanielyan: Verified tests are hitting Real-server and Fake-server , this should be working as intended now

tests/README.md
ddanielyan: updated readme with steps on how to run tests with dotnet and without

tests/README.md
ddanielyan: should be fixed now

languages/go/test/test_suite.go
ddanielyan: fixed

languages/go/test/test_suite.go
tangowithfoxtrot: I'm not sure I understand why we wouldn't just reference the secret ID, rather than a secret name with an ID embedded into it. Using the actual secret ID would avoid conflicts on the real server too, right?

ddanielyan: fixed and removed UUID from names

tangowithfoxtrot:
Hey. Sorry it's been a while since I looked at this. How far away are we from this PR enabling the following?

- Tests can be run on built SDK artifacts pulled from GitHub Actions and built from source. 
- JUnit-style test summaries in the workflow results
- Better support for running tests on Windows

PR #1520, opened 2026-05-20
fix: workaround broken githubs macos runners

tangowithfoxtrot:


tangowithfoxtrot:
[Looks like](https://github.com/bitwarden/sdk-sm/pull/1515) the macOS runners have been fixed. I'll take it!

PR #1567, opened 2026-08-11
chore: group clap crates

tangowithfoxtrot:
## 📔 Objective

See what happens when we group clap, clap_complete, & clap_mangen crates together for Renovate updates.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -17,6 +17,12 @@
     },
     {
       "matchManagers": ["cargo"],
+      "matchPackagePrefixes": ["clap"],
+      "groupName": "clap crates",
+      "groupSlug": "clap"
+    },
+    {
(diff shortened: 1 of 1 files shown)


.github/renovate.json
```diff
@@ -15,6 +15,12 @@
       "groupName": "bitwarden crates",
       "groupSlug": "bitwarden"
     },
+    {
+      "matchManagers": ["cargo"],
+      "matchPackagePrefixes": ["clap"],
```
vincentsalucci, line 20: `matchPackagePrefixes` is removed as of v38 (we're using ~v44) so we should be using  `matchPackageNames`  with something like the following:

` "matchPackageNames": ["/^clap/"],`

OR explicitly list the package names in case you don't want any further "clap" dependencies included in this grouping.

